PR #14, opened 2019-10-18
Dispatch to efficient methods for map(func, AbDimArray) and reduce(..., dims=:)

ivirshup:
Adding a couple more methods I found that are slower than they could be. Using this setup:

```julia
using DimensionalData
using DimensionalData: X, Y
using SparseArrays
using BenchmarkTools

s = sprand(1000, 1000, .1)
sd = DimensionalArray(s, (X <| 1:1000, Y <| 1:1000))
```

On 9a6b121cc:

```julia
julia> @btime map(sin, $s);
  944.490 μs (6 allocations: 1.54 MiB)

julia> @btime map(sin, $sd);
  3.355 s (10 allocations: 3.07 MiB)

julia> @btime reduce(+, $s);
  19.714 μs (1 allocation: 48 bytes)

julia> @btime reduce(+, $sd);
  11.689 ms (8 allocations: 160 bytes)
```

In this PR:

```julia
julia> @btime map(sin, $s);
  941.790 μs (6 allocations: 1.54 MiB)

julia> @btime map(sin, $sd);
  941.316 μs (7 allocations: 1.54 MiB)

julia> @btime reduce(+, $s);
  19.799 μs (1 allocation: 48 bytes)

julia> @btime reduce(+, $sd);
  19.812 μs (1 allocation: 48 bytes)
```

This doesn't cover the case of `map(f, ::Varargs{<:AbDimArray, N})` since I wasn't sure what to do if the dimensions don't match.


rafaqz:
Thanks!

PR #18, opened 2019-11-20
Reverse

rafaqz:
Fix reverse. Reversing dimension index order and array order are different things.


rafaqz:
Also includes new `hasdim` method

PR #26, opened 2019-12-19
add more plot recipes

jw3126:
```julia
using Plots
using DimensionalData: X,Y,DimensionalArray

xs = -1:0.1:2
ys = -2:0.1:5
data = [sin(x^2 + y^2) for x in xs, y in ys]
darr = DimensionalArray(data, (X(xs), Y(ys)))
plot(
    heatmap(darr),
    contour(darr),
    surface(darr),
    wireframe(darr),
)
```
grafik

### DIFF
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -17,3 +17,25 @@ end
     title --> label(refdims(ga))
     val(dims(ga)[1]), parent(ga)
 end
+
+struct HeatMapLike end
+@recipe function f(ga::AbstractDimensionalArray{<:Any,2})
+    sertyp = get(plotattributes, :seriestype, nothing)
+    if sertyp in [:heatmap, :contour, :surface, :wireframe]
+        ga, HeatMapLike()
+    else
+        parent(ga)
+    end
+end
+
+@recipe function f(ga::AbstractDimensionalArray, ::HeatMapLike)
+    @assert ndims(ga) == 2
+    dim1, dim2 = dims(ga)
+    # Notice that dim1 corresponds to Y
+    # and dim2 corresponds to X
+    # This is because in Plots.jl
+    # The first axis of a matrix is the Y axis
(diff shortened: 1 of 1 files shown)


rafaqz:
Thanks! Plot recipes do need a bit more love.

This should be available in 0.4.2 in a few hours.

jw3126:
Cool, thanks for quick merge + release!

rafaqz:
I'm thinking we should change how this works a little. If the actual `X()` and `Y()` dims are used we should reorder them to match Plots.jl x/y order. Otherwise it will plot as currently. Does that make sense?

jw3126:
Ok we can do that. Or should we always transpose? I think there is no right  way to do this, just wrong ones.

rafaqz:
What do you mean exactly? I'm suggesting we do:

```julia
permutedims(A, (X, Y))
```

If the dims are Y, X. We should probably also put `Time` on the x axis if it exists and `X` doesn't. I'm just hoping that most of the time the plots are sensible, but you are right there is not real correct answer.

jw3126:
Really I think this is an unfortunate design decision in `Plots.jl`. There are a couple of desirable properties and we can't have them all:

1. It would be nice if `heatmap(dimarray)` and `heatmap(parent(dimarray))` would look the same (current situation) and not transposed
2. It would be nice if the `X` axis is the `x` axis in the plot and the `Y` axis is the `y` axis in the plot.

We have to violate one, but I don't care much about which one. I am fine with your suggestion.

rafaqz:
Ok that makes sense. I was figuring that the utility of arbitrary dim ordering is automatic permutation. This is important in GeoData.jl where there is one right way to show a map, but at least 4 ways of organising the data - it often has to be reversed as well as permuted. This might be less common with other types of data, I guess thatp most of the dimension and grid operations here will be overkill in other fields.

PR #32, opened 2020-01-13
export @dim macro

Datseris:
This is an important macro in the case where a dimension does not pre-exist, and as far as I can tell, the go-to way to create a dimension with a specified name instead of `Dim{:symbol}`. The readme docs also use it as unspecified.

### DIFF
--- a/src/DimensionalData.jl
+++ b/src/DimensionalData.jl
@@ -78,6 +78,8 @@ export AbstractDimensionalArray, DimensionalArray
 export data, dims, refdims, metadata, name, shortname, 
        val, label, units, order, bounds, locus, grid, <|
 
+export @dim
+
 include("interface.jl")
 include("grid.jl")
 include("dimension.jl")


rafaqz:
It's mostly for devs not users. I dont really want people making their own very often better to just use X, Y and Z for consistency.

Datseris:
Okay, I see. But even for personal use (which is indeed one of the reasons I am currently trying this package out for), I'd still find it better to be able to do `T(0:25:350)`, standing for `Temperature` than `X(0:25:350)`, and I think it is fair to say that `T` would be preferred over `dim{:T}` by anyone user or developer.

Your call.

rafaqz:
Yeah I'll think about it. GeoData.jl Defines Lat and Lon using that macro so that plot recipes work for maps. I guess I'm worried about the proliferation of lots of dims that we wont be able to use dispatch on in plot recipes. 

Hopefully a lot of things will just plot the right way and have the right labels - like time on the x axis, x, y, z axes in the right places etc. And we can't do that for user types - unless we use abstract type groups or something.

Its also just one line to import it in a script... If that gets really annoying some day I'll merge this.

Datseris:
@rafaqz I've just now noticed that `X, Y, Z` are also not exported. Is this intentional or just a missed export?

rafaqz:
That's really up for discussion. I mostly left them out because they are small and common as type parameters so they might annoy people. But maybe we can just claim them and export and people can just use `import` if they don't want them.

This is really an unclear area in the direction of the package. Are X, Y, Z, Time canonical and widely used? Do people make their own often or rarely? Do we use Dim{:foo} a lot or hardly every? We need it to exist at least for loading rare/custom spatial data dims but the syntax is awful. But we could even do A[:foo = Near(33)].

Because we're using structs the possibilities are pretty broad. So I'm really not certain where it will go. Maybe just use it a bit and see what you find inadequate in practice.

Datseris:
> Do we use Dim{:foo} a lot or hardly every? 

Please no :P that was the first thing that made me thing, "well, I'll probably won't be using AxisArrays cause of this syntax" (and this one is in fact the same).

X, Y, Z are indeed canonical I'd say. In many fields/arrays people use capital letters to represent matrices (I do it for example). `Time` is not something I would use myself, as it implies a type definition, but unfornately it has the problem that it conflicts with base.

From my end the best scenario (and what I do at the moment) is to let the user decide their names by exporting `@dim`, and let `XYZ` unexported as is. In fact I've made a dictionary "string" to `Lat/Temp/Whatever` just like you've defined in GeoData.jl, but for my personal use. 

---

For your argument 

> I guess I'm worried about the proliferation of lots of dims that we wont be able to use dispatch on in plot recipes.

Well, its fair that if one wants to use the specific plotting recipes from GeoData they should follow the convection. But if they don't no reason to force it, since as you said the recipes would work but with arbitrary axis names.

I need to come back to your argument

rafaqz:
Yes that syntax is a horror show I had the same response. I'm really trying to make it feel a bit nicer to use here.

With `Time` I use it a lot because I need to find it in modelling tools - it's usualy the axis the simulation follows. The other dims can be whatever but Time is crucial.... Dispersal.jl and GrowthMaps.jl both use it explicitly.

I think you are right about `X`, `Y`, `Z` lets just export them. Making `Time` `T` would avoid the clash with DateTime but then we clash with the most common type parameter `T`. `Ti`? `Tim`? I don't know...

Datseris:
Just to ensure that we are clear, my suggestion was to _not_ export X, Y or Z from _this_ repo. I don't think its a big bummer to export them, as thankfully there are many capital letters to use for matrices (:P) but my vote would be to not export them. As this is let's say the "core" package, more frontend oriented packages like GeoData.jl export common field-specific names that are given to dimensions anyway.

When I mentioned that they are unexported, I wasn't trying to argue that they should be, I was just pointing it out as it is common to forget exports. One should just be careful to use define them in the docs.

For `Time` the situation is harder and I also not know how to proceed, so best to leave it as is now (which I think it is _not_ exported?) and return later.

rafaqz:
Ok yeah that was my feeling. It seems you understand why things are how they are lol. It's just hard to know what the best strategy is early one.

I going to think about making Time `Ti` to avoid the common clash with `Dates.Time`, and making it easier to write.

rafaqz:
Abstract dimension types are in, so I'm reopening and exporting this.

PR #33, opened 2020-01-13
improve pretty print for DimensionalArray

Datseris:
Now from
```
using Dates, DimensionalData
using DimensionalData: Time, X
t = Time(timespan)
x = X(10:10:100)
```

one gets:
image


Datseris:
I don't know if this is a bug, unwanted behavior, or wanted behavior, but this PR made me realize that:
```
julia> B = mean(A; dims = Time)
DimensionalArray with dimensions:
  Time: DateTime[2001-01-01T00:00:00]
  X: 10:10:100
and data:
1×10 Array{Float64,2}:
 0.573145  0.576649  0.454423  0.528486  0.639443  …  0.373484  0.464725  0.47384  0.526452
```
reducing operations keep the dimensions even if reduced over them (in contrast to `xarray` from Python).

rafaqz:
That's because Julia keeps them! After a reducing operation you always have a length 1 dim. So we have to keep the dimension type too... I just change the step size and make it cover the whole original dim length.

I'm not sure what python does.

rafaqz:
But otherwise that looks great. You might want to try it with a large vector of DateTime or something as `val`  and see if it holds up - we might need to just use the start and end however julia does that with arrays.

Datseris:
> You might want to try it with a large vector of DateTime or something as val and see if it holds up - we might need to just use the start and end however julia does that with arrays.

Good point and thankfully I know how to fix it since I've already done it in DynamicaSystems.jl. I'll check it tomorrow at the office.

Datseris:
> That's because Julia keeps them! After a reducing operation you always have a length 1 dim. So we have to keep the dimension type too... I just change the step size and make it cover the whole original dim length.

Yeah, you are right (although I have to admit I never liked that Julia decision). Regardless, what I do personally is implement `drop` versions of these functions, like `dropmean` or `dropsum` that just do 
```
dropmean(A; dims = 1) = dropdims(mean(A; dims = dims); dims = dims)
```
As you can see, for standard Julia arrays this is no big deal. But for DimensionalArray, I would imagine that the processes of "dropping" the singleton dimension is much more involved for the general user. Perhaps I should contribute `dropmean/sum/etc.` here?

Python's `xarray` automatically drops singleton dimensions, e.g. if you `sum` over Lattitude, the final array will have only longitude.

rafaqz:
dropdims is actually easier with dimensions as you don't need to know the number: `dropdims(A; dims=X)

Look in the methods.jl tests to see everything that's implemented with dims. It's quite a lot.

Datseris:
running `test/prettyprint.jl` now gives:

image

Unfortunately, even though I've used the limiter function suggested by Sebastian, it works but not "precisely". It limits the vector output to a elements to fill the row, as it should, but its just goes a little overboard in Juno. Regardless, this is unrelated with this PR, and as far as Julia Base is concerned, I used all the correct tools to make it print the dimension values in 1 line. The rest is up to Juno, and I've reported it.

This is good to go from my end.

test/prettyprinting.jl
```diff
@@ -0,0 +1,6 @@
+using Dates, DimensionalData
+using DimensionalData: Time, X
+timespan = DateTime(2001):Month(1):DateTime(2001,12)
+t = Time(timespan)
+x = X(Vector(10:10:500))
+A = DimensionalArray(rand(12,length(x)), (t, x))
```
rafaqz: Can we run `show` on these? We probably don't need an actual test, but it would be good to [run the code](https://codecov.io/gh/rafaqz/DimensionalData.jl/pull/33/diff?src=pr&el=tree#diff-c3JjL3ByZXR0eXByaW50Lmps) to make sure it doesn't error.

rafaqz:
The outputs look great. Thanks for working on the print limiting I'm happy with that.

Datseris:
> The outputs look great. Thanks for working on the print limiting I'm happy with that.

Thanks, you are welcome. Can I ask why you didn't merge it if you liked the PR? Seems pretty useful to have and I'd like to work on other features as well.

rafaqz:
See the inline comment above! Maybe you missed it. Just add run `show` methods in the tests to make sure they don't error, as a minimal test.

Datseris:
Ah sorry, yes I missed it. I can add proper tests, hold on.

rafaqz:
Thanks thats even better

PR #40, opened 2020-01-21
Implement getkey/haskey

### LINKED ISSUE #38: Convenience way to check if a dimension exists in `dims(s)`
Hi there, I've finally started using GeoData, things seems good so far. I have the following:

image

(which shows I have to work more on pretty printing, but that's for another PR). So as you can see `Longitude` is one of the dimensions. Of course, if one wants to be precise, they have to understand that "Longitude" is just a name. The _actual_ dimension is `Lon`. 

What I am currently missing is a simple way to check if a "dimension" (whatever this means) is in the dimensions of a given DimensionalArray. For example:

```
julia> "Longitude" ∈ dims(s)
false

julia> Lon ∈ dims(s)
false
```

I propose that we make a simple nice function `hasdim(s, dim)` in both of the above examples would return `true`. For the string case, it is obvious, one can use `dim \in name.(dims(s))`. But for the second case, I must admit I find it really hard... This is in general confusing for me:
```
julia> typeof(Lon)
UnionAll

julia> supertype(ans)
Type{T}

julia> Lon <: Dim
false
```
what _is_ `Lon`???

### LINKED ISSUE #39: How to get the value of a dimension by _name_ from a DImensionalArray?
I have a nice dimensional array, that has dimensions
image

At the moment, if I want to get the _value_ of the i-th dimension, I can do `dims(s)[i].val`. But, how do I get the dimension's value by qualifying the dimension type?

Like, I want the values of the `Lon` dimension of `s`. 

Related with #38 .

Datseris:
Closes #38
Closes #39

Here I chose to extend the names of Base, as they simply make sense in the given setting, but should you prefer the naming `getdim/hasdim`, no prob.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DimensionalData"
 uuid = "0703355e-b756-11e9-17c0-8b28908087d0"
 authors = ["Rafael Schouten <rafaelschouten@gmail.com>"]
-version = "0.5"
+version = "0.6"
 
 [deps]
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Dims can be used for indexing and views without knowing dimension order:
 `mean(a, dims=Time)`, or permute `permutedims(a, [X, Y, Z, Time])` in julia
 `Base` and `Statistics` functions that have dims arguments.
 
+Notice that you can obtain the dimensions of an `A::AbstractDimensionalArray` by simply calling `dims(A)`, which returns them as a tuple. `haskey` and `getkey` are also implemented for `A`, where one specifies the dimension either by name (string) or by type.
 
 ## Selectors
 
@@ -46,7 +47,7 @@ _Example usage:_
 using Dates, DimensionalData
 using DimensionalData: Time, X
 timespan = DateTime(2001):Month(1):DateTime(2001,12)
-A = DimensionalArray(rand(12,10), (Time(timespan), X(10:10:100))) 
+A = DimensionalArray(rand(12,10), (Time(timespan), X(10:10:100)))
 A[X<|Near([12, 35]), Time<|At(DateTime(2001,5))]
 A[Near(DateTime(2001, 5, 4)), Between(20, 50)]
 ```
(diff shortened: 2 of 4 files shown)


Datseris:
`isnothing` is not defined, which is why the tests fail, but let's see first whether this will be merged or not before I continue to work on it.

rafaqz:
Already implemented :)

Datseris:
@rafaqz I think you closed this one too fast. The current functions don't work with strings, i.e. when specfying the name of the dimension.

rafaqz:
I'm not sure that should be a thing. The name was meant for display in plots and similar, like ("latitude" for `Lat`) not as a key. What is the reason would this be used over just using the type?

If it does make sense to have, could be retreived at compile time, as it's stored in a function. It doesn't need to have a runtime cost.

Datseris:
> I'm not sure that should be a thing. The name was meant for display in plots and similar, like ("latitude" for Lat) not as a key. What is the reason this would be used over just using the type? 

Hm I think you need to see things from the eyes of a beginner, not of a developer to understand why this is useful.

For a front-end user, the dimension is displayed in the REPL with its name, and as you said, it is also plotted with its name. Thus, it seems natural, that when a user tries to access the information they would use what is displayed to them, i.e. the name. And that's why I think this is important to have.

---

I don't know what you mean by doing this at compile time. The value of a string can't be known at compile time, so I don't know how to do it. I also don't think this operation should be performance sensitive to care that much about this point.

rafaqz:
I disagree, and excuse me I spend a lot of time thinking about what a beginner needs. Please refrain from that kind of inflammatory comment in future and stick to explaining your design goals.

Mostly what beginners need is clean simple syntax and good documentation of that syntax. So the simplest approach is to encourage people to use the types for everything, as they work in all the other methods too. Dimension types are the core, fundamental components of this package.

What this package needs is better documentation, not extra methods. The thing holding that back is my available time being spread very thin, not the will to do it.

Datseris:
> I disagree, and excuse me I spend a lot of time thinking about what a beginner needs. Please refrain from that kind of inflammatory comment in future and stick to explaining your design goals.

Sorry, you misinterpreted my wording. It wasn't judgmental, I was the beginner in this case trying to access with the strings. It wasn't my intention to provide "inflammatory" comments or any such sort.

Okay, then we close this discussion.

rafaqz:
I see your problem with the printed out suggesting the dim name instead of the type. We should to make it suggest the type first, like `Lat "Latitude" Float64[...`

Issue #46, opened 2020-01-30
Repeated use of a dimension


ivirshup:
For context, I would like to be able to have labelled arrays which represent pairwise measures. These measures can be distances, or they could be adjacency matrices for some network structure.

For these pairwise measures, I think it makes sense for the resulting array to use the same dimension twice. As an example:

```julia
using Dates, DimensionalData
using DimensionalData: Time, X
using Distances

timespan = DateTime(2001):Month(1):DateTime(2001,12)
A = DimensionalArray(rand(12,10), (Time(timespan), X(10:10:100))) 
D = DimensionalArray(  # Since the dimensions don't propagate
    pairwise(CosineDist(), A),
    (X(10:10:100), X(10:10:100))
)
```

Subsetting by dimension results in both axes being subset:

```julia
D[X<|Between(0, 50)]
# DimensionalArray with dimensions:
#  X: 10:10:50
#  X: 10:10:50
# and data: 5×5 Array{Float64,2}:
#  0.0       0.321451  0.352489  0.398895  0.444586
#  0.321451  0.0       0.160367  0.193874  0.143861
#  0.352489  0.160367  0.0       0.232279  0.111173
#  0.398895  0.193874  0.232279  0.0       0.247832
#  0.444586  0.143861  0.111173  0.247832  0.0  
```

If I try to look at the distance between 10 and 20, I get the wrong result (both axes are being indexed by the first indexer):

```julia
D[X<|At(10), X<|At(20)]
# 0.0
```

For my work, there would be a lot of value in being able to use a dimension multiple times for one array, and being able to apply different selections to those indices. However I'm not sure how this can be resolved without losing the "order of indices doesn't matter" feature here.

Can these be reconciled? If not, I don't think it'd that large of a restriction to make order matter when indices are repeated, or just in general. Having the order be important makes the performance implications of the code be more explicit.

--------------------------

Using the order of the selectors to specify axis does is a workaround when dimensions are repeated:

```julia
 D[2:3, 1:2] == D[Between(20, 30), Between(10, 20)]
```

As an aside, this took me a little bit to find out, as I expected this to work initially:

```julia
julia> D[Between(20, 30), :]
ERROR: ArgumentError: invalid index: Between{Tuple{Int64,Int64}}((20, 30)) of type Between{Tuple{Int64,Int64}}
```

rafaqz:
Yes that is currently undefined behaviour.

[...]

Issue #31, opened 2020-01-11
[Dev Question] What's the purpose/use of Grid and co types?


Datseris:
It is useful for the community to have a way to transform a loaded NetCDF variable into an array with Dimension data attached it, like what this package does. We will implement such an interface in NCDatasets.jl, cf https://github.com/Alexander-Barth/NCDatasets.jl/issues/60 , but unfortunately there are numerous AxisArray-like packages. I've had a look in all of them, and at least from the surface, this one seems to be the strongest candidate. 

But since it has been developed by a single person, and I would like to have at least some basic understanding before basing NCDatasets.jl on this one, there are some dev questions. The first one is regarding the `Grid` types/subtypes/supertypes, which at the moment is something I am not sure about.

1. What is their purpose?
2. Why is it that the dimension data type by itself is not enough? For example, a `Range` should always be a "standard grid" (AlignedGrid as its called here), while a `Vector` should always be a `BoundedGrid`?

rafaqz:
They will eventually move here:
https://github.com/JuliaGeo/DimensionalArrayTraits.jl

The design had a lot of input from multiple people in JuliaGeo, not just me.

The grid type is to deal with the many ways that dimension indices are stored. Ranges equal spaced. Vectors (youre not sure they are equal spaced but they actually can be). Matrices for all lat/lon combos. Vectors with coordinate transforms. Categorical indices. No index.

On some of these grids the `Near` Selector makes sense. But categories logically can't be nearest anything. They have no size. `At` mostly always makes sense but might need different techniques for performance.

Some datasets come with some dimension reversed. Or both. Or the data reversed and the dimension forwards. If we want `searchsortedfirst` to work and plots to come out the right way around we need to know and track which is which all the time. So you need the `Ordered` type.

Then you want a bounding box for each dimension. So `bounds` works no matter what the contents are after every subset/view that you do of the data. This is easy with a range, harder with a vector.

If you take the mean of the time dimension the bound should stay the same but the step size should now cover the whole span. It should do that for any reducing method. Then a plot of aggregated data will show the right time span.

etc etc etc.

Grids track all that stuff so that plots always plot correctly, we can always use `bounds`, and we can use the same methods on really quite different underlying data.

But just use GeoData.jl. It already works for NCDatasets.jl lol

rafaqz:
Also ranges really map to `RegularGrid`, not `AllignedGrid`. Alligned just means not rotated or warped, while Regular means regularly spaced as well.

[...]

PR #48, opened 2020-01-30
[WIP] Support for mixed positional indexing

ivirshup:
Based on some discussion in #46. Given this setup:

```julia
# Setup
using Dates, DimensionalData
using DimensionalData: Time, X, AbDim
using Distances

```

Meant to allow statements like:

- [x] `A[Between(DateTime(2001, 1), DateTime(2001, 5)), :]`
- [x] `A[:, At(30)]`
- [x] `A[Between(DateTime(2001, 1), DateTime(2001, 5))]`
- [x] `A[:, At(50)] = rand(12)`
- [x] `A[1:5]`

Any reason to believe this won't work out?


test/selector.jl
```diff
@@ -38,6 +38,24 @@ a = [1 2  3  4
         @test view(da, Between((11, 20)), Between((2u"s", 3u"s"))) == [6 7]
     end
 
+    @testset "mixed selectors and standard" begin
+        indices = [
+            (Between(9, 31), Near((3:4)u"s")),
+            (Near(22), At([3.0u"s", 4.0u"s"])),
+            (At(20), At((2:3)u"s")),
+            (Near<|13, Near<|[1.3u"s", 3.3u"s"]),
```
rafaqz: These all seem to be selectors?

Shouldn't you have `:`, `2`, `1:10` etc in here as well?

src/selector.jl
```diff
@@ -149,7 +152,7 @@ relate(::Forward, dim, val) = val
 relate(::Reverse, dim, val) = lastindex(dim) .- val .+ 1
 
 # Selector indexing without dim wrappers. Must be in the right order!
-Base.@propagate_inbounds Base.getindex(a::AbstractArray, I::Vararg{Selector}) =
+Base.@propagate_inbounds Base.getindex(a::AbDimArray, I::Vararg{Union{Selector, StandardIndices}}) =
     getindex(a, sel2indices(a, I)...)
 Base.@propagate_inbounds Base.setindex!(a::AbstractArray, x, I::Vararg{Selector}) =
     setindex!(a, x, sel2indices(a, I)...)
```
rafaqz: We should also make this change for `setindex!` and `view`. You could make a `SelectorOrStandard` constant for the type.

rafaqz (commented):
This should really work already, thanks for the PR.

test/selector.jl
ivirshup: I convert these to their equivalent indexing statements with `sel2indices` below in lines 51:53.

src/selector.jl
ivirshup: > We should also make this change for setindex! and view.

That's the plan!

Why were all these methods defined for `AbstractArray` and not `AbstractDimArray`? Modify that doesn't break any tests, but I wasn't sure if there was some untested behaviour I should be aware of here.

ivirshup:
Should these do the same thing?

```julia
data(A[1:5]) == data(A)[1:5]
data(A[Between(DateTime(2001, 1), DateTime(2001, 5))]) == data(A)[1:5]
```

src/selector.jl
rafaqz: Great!

I originally had the idea that you don't have to inherit from AbstractDimensionalArray to use a lot of this package. Just a `dims` and `rebuild` method and dimension indexing gets you a long way. 

It might not end up making sense. My goal was to make everything as widely extensible as possible, as the opposite of AxisArrays that really isn't extensible at all. I probably went too far.

test/selector.jl
rafaqz: Maybe make it more explicit and just write it out - it doesn't really communicate what is being tested here.

ivirshup: Would this be more clear?

```julia
equivalent_cases = product(sel_pair, sel2indices(da, sel_pair))
for (case1, case2) in combinations(equivalent_cases, 2)
    # Tests that the cases are equivalent
end
```

ivirshup: These are the same test cases as are used in the `"selectors without dim wrappers"` tests.

My preference is to seperate the test data from the code being tested so it's easy to have new data go through all checks, or replace the cases with a case generator.

rafaqz: I don't want to use sel2indices in the tests like that. Its too coupled and not clear to people reading them what you intend to test. Just write them out.

rafaqz:
Yes this should hold

`data(A[1:5]) == data(A)[1:5]`

Datseris:
Please try to have some basic documentation about this somewhere (probably in the readme until we have real docs), so that it is easy for other people to know about this change and to be able to take it into account for their contributions. 

This will make it possible to understand whats happening without having to go through the full PR. And in this scenario this is just a simple statement in the readme like:

> Selectors also work by position (when not specified the target dimension), i.e. if `X, Y` are the two dimensions of `A` (in order), then `A[At([1, 2]), At([4, 5])]` , `A[X <| At([1, 2]), Y <| At([4, 5])]` and `A[Y <| At([4, 5]), X <| At([1, 2])]` are all equivalent.

Of course the above is only valid _if_ I actually understood this PR correctly... Thank you :)

Datseris:
Meh, actually I've just put it in #50

Datseris:
BTW forgot to menton, @ivirshup this is _reaaaally_ nice!

ivirshup:
@rafaqz, for this case: `data(A[1:5]) == data(A)[1:5]`

If A has more than one dimension, would `A[idx::StandardIndices]` always return an unlabelled array? Is it weird that it's different for a 1D array?

ivirshup:
@Datseris thanks!

What you wrote looks right to me. What's changing here is that these are now equivalent:

```julia
# Example data
A = DimensionalArray(reshape(1:9, (3, 3)), (X(10:10:30), Y(2:2:6)))

A[At(20), At(4)] == A[2, At(4)]
A[X<|At(20)] == A[At(20), :]
```

Previously, errors would have been thrown, since you couldn't mix standard indices and selectors.

src/selector.jl
ivirshup: I'm not sure it makes sense for this method to apply to normal arrays, but it would be cool if this could dispatch on some meta-package's trait for labelled arrays. I had to change it from `AbstractArray` here since it was causing dispatch errors to be that general.

Datseris:
```
A[At(20), At(4)] == A[2, At(4)]
A[X<|At(20)] == A[At(20), :]
```
useful examples as well, I'll put them in the docs once this is merged.

Datseris:
@ivirshup I am not fully sold on the syntax of `A[1:5]`. I think allowing it for `A` having only one dimension is fine, but I would find it quite unintuitive if `A[1:5]` did the same as standard Julia arrays and accessed `A` as it was a vector.

For base Julia, it makes sense that single indicing acts on `A` as if it is a vector, because in fact the implementation of abstract array is in plinciple "always a vector", and the only diffecerence between `Matrix` and `Vector` is the indexing implementation.

For an Array with named dimensions, this is no longer the case, and you cannot consider a Matrix (with 2 dimensions) and a Vector equivalent things. At least this is how I think about it. 

Do you really gain something by having the syntax `A[1:5]` for multidimensional arrays? I would vote that we don't support something that is unintuitive.

ivirshup:
> implementation of abstract array is in principle "always a vector"

Do you mean it's always a vector as in, always a vector in memory? I'm not sure this is the case, for example: `A'[1:5]` does not return the values in the order they are stored in memory.

Otherwise, are there some docs about why indexing is like this? I'm not sure I get the value of this indexing style for multidimensional arrays.

> Do you really gain something by having the syntax `A[1:5]` for multidimensional arrays?

Being able to pass a `DimensionalArray` to a method written for `AbstractArray` and having it work all of the time would be a very nice feature. I think it would play well to julia's composability. I don't know if "all of the time" is possible, but I think it's worth trying out.

src/selector.jl
ivirshup: Just had a thought about this and the use of `sel2indices` in the tests.

Why not assign these methods to `to_indices(::DimensionalArray, ...)`?

Datseris:
> Being able to pass a DimensionalArray to a method written for AbstractArray and having it work all of the time would be a very nice feature. 

Okay this convinced me, and I assume that you'll try to make single index access zero-cost then, as in simply `getindex(A::DimAr, i::Int) = A.data[i]`.

> Do you mean it's always a vector as in, always a vector in memory?

Yeap, I think so. The difference that dimension of the array makes is, to the best of my knowledge, multi-index access and multiple dispatch (and of course other secondary aspects like e.g. printing).

rafaqz:
> @rafaqz, for this case: `data(A[1:5]) == data(A)[1:5]`
> 
> If A has more than one dimension, would `A[idx::StandardIndices]` always return an unlabelled array? Is it weird that it's different for a 1D array?

I'm not sure I understand, but an `AbstractDimensionalArray` should always return another `AbstractDimensionalArray` from all operations that we can reasonably support. in my quote both arrays have the `data(A)` method applied, which is returning the internal array without dims on both sides, just at different points in the process.

ivirshup:
> If A has more than one dimension, would `A[idx::StandardIndices]` always return an unlabelled array? 

Here's an example of the case I'm talking about:



```julia
# For an array that looks like:
julia> da
DimensionalArray with dimensions:
 X: 2:2:6
 Y: 3:3:9
and data: 3×3 Array{Int64,2}:
 1  4  7
 2  5  8
 3  6  9

julia> data(da)[3:4]
2-element Array{Int64,1}:
 3
 4
```

`da[3:4]` should have those points, but these data points don't share any indices. So what would the dimensional array look like?

Datseris:
Isn't this ambiguous behavior though...?

If `A` is one-dimensional, clearly `A[3:5]` should return a DimensionalArray, in order to be in line with the rest of this package. If `A` has many dimensions `A[3:5]` can't do that anymore... Why support something so ambiguous, i.e. accessing multidimensional A with a range?

rafaqz:
>If A has more than one dimension, would A[idx::StandardIndices] always return an unlabelled array?

Ok I understand you now. We are talking about linear indexing.

Linear indexing on a multi-dimensional array will have to return a regular `Array` as the dimensions are meaningless after indexing across them. But a `Vector` always uses linear indexing and only has one dimension before and after indexing, so we can still return an `AbstractDimensionalArray`.  

I think this is broken currently. It's also interesting to think about what indexing with one `Selector` like `At(1:10)` does in a multidimensional array - that shouldn't be allowed.

ivirshup:
Figuring out the dispatch there was a little painful, but I think I've got it right now. Let me know if this looks right.

For `view(::DimArray{<:Any, 2}, ::Int)`, I've gone with dropping labels instead of figuring out which labels to refer to.

rafaqz:
Should we do the same for `setindex!`? (bottom of the array.jl page) 

It probably should dispatch on `AbstractDimensionalArray` not just `DimensionalArray`, if you want to move it up.

ivirshup:
> Should we do the same for setindex!? (bottom of the array.jl page)

I don't think it needs to be changed, but I've added tests for it and modified the signature.

------------------

For defining a `to_indices` method, `sel2indices` would be straight forward:

```julia
Base.to_indices(A::AbstractDimensionalArray, lookup::NTuple{N, SelectorOrStandard} where N) = \
    sel2indices(dims(A), lookup)
```

But it gets a bit more complicated with `dims2indices`. I'm thinking this would be worth taking a little closer look at, so might be better in another PR. It would also be good if I had a chance to understand exactly what the `Grid`s do.

rafaqz:
So does `setindex!` work with mixed Selector/StandardIndices?

Maybe read #31 for grids. They are also for handling transformed/rotated etc indexes where you always need to specify multiple indices for a matrix lookup or a coordinate transform function.

ivirshup:
> So does `setindex!` work with mixed Selector/StandardIndices?

Yup, that's tested around here: https://github.com/rafaqz/DimensionalData.jl/blob/f69f8fb10c11b633a4aa1182cceec8cb851acf26/test/selector.jl#L56-L62

ivirshup:
@rafaqz, if I delay the `to_indices` change to a separate PR, is this good to go?

rafaqz:
Yeah let's do that separately.

rafaqz:
Good to go, just waiting on that change to the tests: write out the test data as both selectors and regular indices instead of unwrapping them with `sel2indices`. It confused me and I wrote `sel2indices` so it will confuse other devs too as to what you are actually testing.

I also used to aim for DRY tests, but now I think simplicity and communicating your intent is more important (not that mine are in any way perfect either).

ivirshup:
I can, but would the test have been confusing if it used `to_indices`? If so, would a `# TODO` comment work until it uses `to_indices`?

I like the decoupling of tests and test data since it makes it easy to eventually use a property testing approach where you can generate the test data. For code that's supposed to work with a large number of types, I've been impressed by the number of bugs I can catch this way.

rafaqz:
Yes it would be a little less confusing to outsiders using a base method. 

But the point is that it should show exactly what input you are testing without that much convolution, it just isn't necessary to save 5 lines of code.

rafaqz:
I'm just going to merge this as we need to move on

Datseris:
Awesome, I'll update the docs tomorrow (at least I hope by tomorrow) and then we will also have cool docs as well!

ivirshup:
Oh, sorry! Bad timing, was just thinking about this on the tram!

I've modified the tests in #62

rafaqz:
Oh great thanks. I was going to change them myself but it's complicated. For my sanity try keeping them a little simpler ;)

And trams are such a giveaway lol, I'm also in Melbourne. What are you doing at unimelb?

ivirshup:
Sorry, totally forgot about the last message!

I'm doing my phd in computational biology, mainly looking at single cell transcriptomics (scRNA-seq). What about you?

rafaqz:
Good to know there are other people writing Julia around! especially in biosciences.

I finished a masters last year in biophysical ecology, also at unimelb BioSciences. But currently on a break from that, working as an ecological modeller (writing Julia packages all day), mostly just down the road from you at cesar.

ivirshup:
Sadly, I'm mostly writing python packages (beats working in R). There's a group at MIG which uses a bunch of Julia.

> mostly just down the road from you at cesar.

Well, we should meet up at some point! I'd really like to hear how these kinds of data structures are used by people doing geospatial work.

rafaqz:
>Well, we should meet up at some point!

Yeah was thinking that. Maybe a beer after work/uni? I'm in sydney next week, but could do friday, or thurs or friday this week.

ivirshup:
That'd sounds good! I'll have to get back to you about timing, but I'll send you my contact info on the julia discourse.

PR #54, opened 2020-02-03
multiply methods

### LINKED ISSUE #47: Incorrect dimensions after matrix multiplication
My assumption here is that `x::DimensionalArray -> length.(dims(x)) == size(x)` should be true for enumerated dimensions. Here's an example:

```julia
using Dates, DimensionalData
using DimensionalData: Time, X

@assert length.(dims(A)) == size(A)
@assert length.(dims(A')) == size(A')

length.(dims(A * A')) == size(A * A')
# false
```

This seems right:

```julia
size(A * A')
# (12, 12)
```

But the first dimension is `X`, when it should be `Time`:

```julia
dims(A * A')
# (X: 10:10:100, Time: 2001-01-01T00:00:00:1 month:2001-12-01T00:00:00)
```

However, if both dimensions were `Time`, that would start leading to problems with #46.

rafaqz:
Closes #47

This is mostly working, although it runs into problems with `Adjoint`/`Transpose`, as multiplying them can return a scalar instead of a length 1 array.

We'll need to match the internal `Adjoint`/`Transpose` arrays with a second method. Which is annoying, and I don't have time for now.

### DIFF
--- a/src/dimension.jl
+++ b/src/dimension.jl
@@ -111,6 +111,13 @@ shortname(::Type{<:Dim{X}}) where X = "$X"
 basetypeof(::Type{<:Dim{X}}) where {X} = Dim{X}
 
 
+struct EmptyDim <: AbstractDimension{Int,NoGrid,Nothing} end
+
+val(::EmptyDim) = 1:1
+grid(::EmptyDim) = NoGrid()
+metadata(::EmptyDim) = nothing
+name(::EmptyDim) = "Empty"
+
 """
     @dim typ [name=string(typ)] [shortname=string(typ)]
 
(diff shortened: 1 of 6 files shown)


ivirshup:
@rafaqz, do you have an example of the failure for Adjoint/ Transpose? I think I may be onto a fix in #78

rafaqz:
This is the main problem with adjoint:

```julia
julia> permutedims(ones(10)) * ones(10)
1-element Array{Float64,1}:
 10.0

julia> ones(10)' * ones(10)
10.0
```

Multiplying an adjoint doesn't always return a vector. So we'll need to special case it somehow.

PR #55, opened 2020-02-03
Further improve pretty printing

### LINKED ISSUE #53: Further improve pretty printing of DimensionalArray
Although the PR #33 improved the situation, much is left to be desired:

image

The printing of the dimensions is not optimal, too many elements are shown and they are taking multiple lines, which messes the alignment. The array data take up several pages when printed, which is also not so helpful.

No matter how much I hacked at it I couldn't make it work cleanly. I don't know why it worked cleanly for DynamicalSystems.jl.

Anyway, here is what I propose to make the situation better, which I can easily put in a PR:
1. For any dimension by default show at most 4 elements: first, second, end-1 and end, and put `...` in between.
2. For array data with dimension 3 or larger, only show the first 2D frame, i.e. `A[:, :, 1, 1, 1, 1,...]`.

This will make the printing both useful as well as readable.

Datseris:
Closes #53 

Alright, I think this leaves noone unsatisfied! We now do
```julia
using DimensionalData, Test
using DimensionalData: Time, Z, @dim
using Dates: DateTime, Month

# define dims with both long name and Type name
@dim Lon "Longitude" "lon"
@dim Lat "Latitude" "lat"

t = Time(timespan)
x = Lon(Vector(0.5:1.0:359.5))
y = Lat(Vector{Union{Float32, Missing}}(-89.5:1.0:89.5))
z = Z('a':'z')
d = (x, y, z, t)

```
and get:
image

image

By the way, if you want to have same "only first frame printing" for all your Julia multidim arrays you can put this in your startup.jl:

```julia
function Base.show_nd(io::IO, A::AbstractArray, print_matrix::Function, lab::Bool)
    o = ones(Int, length(size(A))-2)
    frame = A[:, :, o...]
    onestring = join(o, ", ")
    println(io, "[:, :, $(onestring)]")
    print_matrix(IOContext(io, :compact => true, :limit=>true), frame)
    print(io, "\n[and ", prod(size(A,d) for d=3:ndims(A)) - 1," more slices...]")
end
```

I did it already :D 

---

updating the tests now, the this can be reviewed/merged.

### DIFF
--- a/src/prettyprint.jl
+++ b/src/prettyprint.jl
@@ -61,10 +61,10 @@ end
 # printing for DimensionalArray
 function Base.show(io::IO, A::AbDimArray)
     l = nameof(typeof(A))
-    printstyled(io, nameof(typeof(A)); color=:magenta)
+    printstyled(io, nameof(typeof(A)); color=:blue)
     if label(A) != ""
         print(io, " (labelled ")
-        printstyled(io, label(A); color=:magenta)
+        printstyled(io, label(A); color=:blue)
         print(io, ")")
     end
 
--- a/src/prettyprint.jl
+++ b/src/prettyprint.jl
@@ -1,14 +1,73 @@
-# Credit to Sebastian Pfitzner for `printlimited`
-function printlimited(io, x; Δx = 0)
-    sz = displaysize(io)
-    ctx = IOContext(io, :limit => true, :compact => true,
-    :displaysize => (1, sz[2]-Δx))
-    Base.print(ctx, x)
+function printlimited(io, v::AbstractVector)
+    s = string(eltype(v))*"["
+    if length(v) > 5
+        svals = "$(v[1]), $(v[2]), …, $(v[end-1]), $(v[end])"
+    else
+        svals = join((string(va) for va in v), ", ")
+    end
+    print(io, s*svals*"]")
 end
 
+# Thanks to Michael Abbott for the following function
+function custom_show_nd(io::IO, A::AbstractArray{T, N}) where {T,N}
+    o = ones(Int, length(size(A))-2)
+    frame = A[:, :, o...]
+    onestring = join(o, ", ")
+    println(io, "[:, :, $(onestring)]")
+    Base.print_matrix(IOContext(io, :compact => true, :limit=>true), frame)
+    print(io, "\n[and ", prod(size(A,d) for d=3:N) - 1," more slices...]")
+end
+
+custom_show_nd(io::IO, A::AbstractArray{T, 2}) where {T} =
+Base.print_matrix(IOContext(io, :compact => true, :limit=>true), A)
+
+custom_show_nd(io::IO, A::AbstractArray{T, 1}) where {T} =
+Base.show(IOContext(io, :compact => true, :limit=>true), A)
+
+# Full printing version for dimensions
+function Base.show(io::IO, ::MIME"text/plain", dim::AbDim)
+    print(io, "dimension ")
+    printstyled(io, name(dim); color=:red)
+    if name(dim) ≠ string(nameof(typeof(dim)))
+        print(io, " (type ")
+        printstyled(io, nameof(typeof(dim)); color=:red)
+        print(io, ")")
+    end
+    print(io, ": ")
+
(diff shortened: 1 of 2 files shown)


Datseris:
@rafaqz I want to test this with an `A` with label. What is the easiest way to give a label when constructing a `DimensionalArray` , so that `label(A)` returns it?

rafaqz:
`@dim A "A"`.

`label` also includes units if the `units` method returns anything. It's for plotting etc, but feels a little unfinished at this stage. `name` just returns the name.

Datseris:
Sorry, I wasn't clear. I wanted to label an array, not the dimension of an array. 

Does `DimensionalArray` not have "metadata"? Because i thought the dimensions themselves have metadata. I was thinking for something simple like
```
DimensionalArray(A, dims; label = "Test label")
```

rafaqz:
`DimensionalArray` doens't have a `name` or `metadata` field. I guess they're kind of optional and DA was meant to be light weight. `GeoArray` in GeoData has both. The intention is that `metadata` should hold wrapped metadata objects that we can write `convert` methods for to transform to different metadata to save in another format, probably not including the name. And `name` should just be some string that might go on the colorbar in a heatmap, or similar. DimensionalArray could have a `name` field if you want, I'm not sure it needs a `metadata` field.

rafaqz:
Want to try and clean up those few code coverage misses? It would be good to aim for 100% sometime soon.

Datseris:
For `metadata` I am also sold that it shouldn't have that field. A dedicated `name` field though I think is useful. Not only for printing purposes. But also because it is typically the case that one would have several fields with the same dimensions, so mentioning a name when showing the fields is not only "pretty" but also useful.

The source implementation of `DimensionalArray` is quite straightforward, I could add this field to this PR if you also agree.

--- 

> Want to try and clean up those few code coverage misses? It would be good to aim for 100% sometime soon.

Unfortuantely I really have no clue how to read the coverage services. If you can tell me which functions should be tested I will add it in the PR.

rafaqz:
Yeah I can't actually get into the codecov for this either for some reason so don't worry about it.

Datseris:
Wait, do you want the `name` in the end?

rafaqz:
Oh I thought you would add `name` in a different PR. 

Also maybe test refdims printing a little more, this just broke GeoData.

Datseris:
ok I'll add it in a new PR.

What broke? Post the error and I'll see what I can do. I've used it in my own code that did `mean(...; dims)` and other reducing operations that port dims to refdims and it worked there.

rafaqz:
This is broken:

```julia
Lon(Between(-125.0, -66.96)), Lat(Between(25.0, 49.5))
```

Atom error:

```julia
Error displaying Tuple{Lon{Between{Tuple{Float64,Float64}},UnknownGrid,Nothing},Lat{Between{Tuple{Float64,Float64}},UnknownGrid,Nothing}}: MethodError: no method matching printlimited(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Between{Tuple{Float64,Float64}})
Closest candidates are:
  printlimited(::Any, !Matched::AbstractArray{T,1} where T) at /home/raf/julia/DimensionalData/src/prettyprint.jl:2
in render at Atom/lBERI/src/display/display.jl:19 
in Copyable at Juno/oLB1d/src/types.jl:40 
in Copyable at Juno/oLB1d/src/types.jl:39 
in render at Atom/lBERI/src/display/display.jl:16 
in sprint at base/strings/io.jl:101 
in #sprint#339 at base/strings/io.jl:105
in  at Atom/lBERI/src/display/display.jl:17
in show at base/multimedia.jl:47 
in show at base/show.jl:715 
in show_delim_array at base/show.jl:682 
in show_delim_array at base/show.jl:697
in show at julia/DimensionalData/src/prettyprint.jl:58
```

`show` needs to handle anything that dims can hold - which will be `Int` and `Selector` as well as `AbstractArray`. But actually they can hold anything as the type isn't limited, and who knows how people want to use it. So it needs to fail gracefully with anything in the val field.

rafaqz:
Ok easy fix. I'll just push it to master.

PR #61, opened 2020-02-06
Add `name` to DimensionalArray

### LINKED ISSUE #58: Design of label field for DimArray
Cf. #53 . I was about to do the label field (I don't know what is better: label or name?), but I"ve realized that we should discuss first whether this field should propagate.

Initially my plan was to simply print the DimArray as `DimensionalArray, labelled $(label) with dim....`. But my concern is, should the `label` field be propagated after operations? At first I though "sure". And my plan was to simply edit `rebuild` to propagate the field.

The issue of course is that one can just add two DimArrays with different labels. Then there it doesn't make sense to propagate either of the labels...

So, should we just keep it simple and not propagate the name at any operation? Or are there some specific operations that it makes sense to propagate the field?

Datseris:
CLoses #58

Adding the name and putting it into the print was straight forward.

Buuut, the name does not propagate through get index, e.g. `A[Time(1:4)]` does not propagate the name. Reducing operations like `sum` propagate the name.

I want to do this as simple as possible and extend as little behavior as possible. I couldn't yet find where exactly is the lowest-lowest-truly-lowest method of `getindex`. Can you please help here and point me there?


rafaqz:
This is the lowest level getindex:
https://github.com/rafaqz/DimensionalData.jl/blob/master/src/array.jl#L18-L24

But you probably want to change `rebuildsliced`
https://github.com/rafaqz/DimensionalData.jl/blob/master/src/dimension.jl#L72

I will do some reorganising, I realise that it's a bit confusing where things are currently.

Datseris:
Cool thanks! It's a bit late here now, soI'll try to do this by tomorrow night.

Datseris:
OKay, now name propagates after indexing, but unfortunately it also propagates after binary operations, e.g. `C = A + B` gets the name of `A`. Could you also point to where the basic math happens?

Sorry to ask many basic questions but I've checked primitives.jl and methods.jl and I couldn't find it. `@which` also doesn't help, since 
```
julia> @which A + B
+(A::AbstractArray, B::AbstractArray) in Base at arraymath.jl:38
```

rafaqz:
Yeah that stuff is pretty much magic.

+` on `AbstractArray` calls `similar` on the first array. It would be good to override these to make sure the dimensions actually match before adding.

But for now changing rebuild in similar would do the trick. https://github.com/rafaqz/DimensionalData.jl/blob/master/src/array.jl#L47-L59

Datseris:
Alright @rafaqz this is good to go for review. 2 things from my side for you to check:

1. Is the way I print the name the best? "named $name" ?
2. Are there any performance regressions? A lot of functions were changed to keyword versions, and I am not sure whether this affects performance...

rafaqz:
1. Maybe just `DimensionalArray "name"`. But whatever looks best.
2. That could be a problem, I'm not sure how much that compiles away at this stage. Check the performance at the end of the test against master (don't check against travis though, its wrong due to `--no-inline` for codecov accuracy)

Datseris:
Master:
```
julia> @btime $A .+ $A;
  59.476 ms (11 allocations: 154.25 MiB)

julia> @btime $A .+ $A;
  64.484 ms (11 allocations: 154.25 MiB)

julia> @btime $A .+ $A;
  72.964 ms (11 allocations: 154.25 MiB)

julia> @btime $A .+ $A;
  65.184 ms (11 allocations: 154.25 MiB)

julia> @btime $A[Time(1:4)];
  17.549 ms (19 allocations: 51.42 MiB)

julia> @btime $A[Time(1:4)];
  17.594 ms (18 allocations: 51.42 MiB)
```

This PR:
```
julia> @btime $A[Time(1:4)];
  17.534 ms (22 allocations: 51.42 MiB)

julia> @btime $A[Time(1:4)];
  17.550 ms (22 allocations: 51.42 MiB)

julia> @btime $A .+ $A;
  47.932 ms (14 allocations: 154.25 MiB)

julia> @btime $A .+ $A;
  47.997 ms (14 allocations: 154.25 MiB)
```

`@btime` reports consistently "more" allocations in this PR, but the actual amount of memory, as well as spent time is identical. (it seems that addition is faster here but I blame this on my laptop)

rafaqz:
You probably need to time much smaller atray ops to see the difference.

Datseris:
In the meantime, I am gettingn test errors in this part:
```julia
@testset "dimension mirroring methods" begin

    a = rand(5, 4)
    da = DimensionalArray(a, (Y((10, 20)), X(1:4)))

    cvda = cov(da; dims=X)
```
which gives
```

ERROR: ArgumentError: length of X (4) does not match size of array dimension (Base.OneTo(5))
```
I have trouble understanding how can this be related wit the current PR...

Since a week now I have been getting warnings like
```
WARNING: Method definition permutedims(Tuple{#s16, Vararg{DimensionalData.AbstractDimension{T, G, M} where M where G where T, N}} where #s16<:(DimensionalData.AbstractDimension{T, G, M} where M where G where T) where N, Tuple{#s16, Vararg{DimensionalData.AbstractDimension{T, G, M} where M where G where T, N}} where #s16<:(DimensionalData.AbstractDimension{T, G, M} where M where G where T) where N) in module DimensionalData at C:\Users\m300808\.julia\dev\DimensionalData\src\primitives.jl:25 overwritten at C:\Users\m300808\.julia\dev\DimensionalData\src\primitives.jl:28.
  ** incremental compilation may be fatally broken for this module **
```
which may be related?

Datseris:
OKay, this is the benchmark file:
```julia
using DimensionalData, Test
using DimensionalData: Time, Z, @dim
using Dates: DateTime, Month
using BenchmarkTools

# define dims with both long name and Type name
@dim Lon "Longitude" "lon"
@dim Lat "Latitude" "lat"

dt = 12
dx = 20.0

timespan = DateTime(2001):Month(dt):DateTime(2011,12)
t = Time(timespan)
x = Lon(Vector(0.5:dx:359.5))
d = (x, t)

println("Time to access via Tim(1:3)")
@btime $(A)[$(Time(1:3))];
println("Time to access via A.data[:, 1:3]")
@btime $(A.data)[1:3];

println("Time to do A+A")
@btime $(A) .+ $(A);

println("Time to do A.data + A.data")
@btime $(A.data) .+ $(A.data);
```

Master:
```
Time to access via Tim(1:3)
  672.308 ns (3 allocations: 624 bytes)
Time to access via A.data[:, 1:3]
  28.039 ns (1 allocation: 112 bytes)
Time to do A+A
  3.575 μs (7 allocations: 2.14 KiB)
Time to do A.data + A.data
  145.745 ns (1 allocation: 1.77 KiB)
```

This PR:
```
Time to access via Tim(1:3)
  907.107 ns (6 allocations: 720 bytes)
Time to access via A.data[:, 1:3]
  28.241 ns (1 allocation: 112 bytes)
Time to do A+A
  3.862 μs (9 allocations: 2.19 KiB)
Time to do A.data + A.data
  153.800 ns (1 allocation: 1.77 KiB)
```

Conclusion: **the keyword arguments made it slower**.

---

We will leave this open for a while, as I don't have the time/willpower to (a) understand and (b) solve this regression. I'll work on the doc PR instead so that we have it running.

rafaqz:
Mostly you're timing DateTime and vector allocation. Use simpler ranges in benchmarks (Unitful compiles away):

```julia
using DimensionalData, Test
using DimensionalData: Time, Z, @dim
using Dates: DateTime, Month
using BenchmarkTools

# define dims with both long name and Type name
@dim Lon "Longitude" "lon"
@dim Lat "Latitude" "lat"

dt = 12
dx = 20.0

timespan = 1u"d":1u"d":10u"d"
t = Time(timespan)
x = Lon(0.5:dx:359.5)
d = (x, t)
```

Now you can see that access is fine on master, but adding the arrays is really bad. Looks like the generic broadcast op isn't very efficient.

```
julia> println("Time to access via Tim(1:3)")
Time to access via Tim(1:3)

julia> @btime $(A)[$(Time(1:3))];
  127.993 ns (2 allocations: 656 bytes)

julia> println("Time to access via A.data[:, 1:3]")
Time to access via A.data[:, 1:3]

julia> @btime $(A.data)[1:3];
  37.681 ns (1 allocation: 112 bytes)

julia> println("Time to do A+A")
Time to do A+A

julia> @btime $(A) .+ $(A);
  7.825 μs (2 allocations: 1.70 KiB)

julia> println("Time to do A.data + A.data")
Time to do A.data + A.data

julia> @btime $(A.data) .+ $(A.data);
  237.567 ns (1 allocation: 1.59 KiB)
```

See how it compares to your PR now. Obviously the broadcast is bad already and I'll have to fix it, but the access we should try to keep it fast.

rafaqz:
I'm going to copy in a broadcast implementation from another similar package. I just never implemented it and the default does a lot of slow checks.

Datseris:
Yes of course, you have a correct point that I completely forgot: the dimensions are also allocated (which makes sense). With ranges that is not the case and indeed the indexing performance is identical. With the updated file we now get:

### Master
```
Time to access via X(1:3)
  69.774 ns (2 allocations: 432 bytes)
Time to access if X(1:3) is not interpolated constant
  66.393 ns (2 allocations: 432 bytes)
Time to just access A.data
  65.884 ns (2 allocations: 432 bytes)
Time to do A+A
  1.460 μs (2 allocations: 1.13 KiB)
Time to do A.data + A.data
  117.467 ns (1 allocation: 1.06 KiB)
```

### This PR
```
Time to access via X(1:3)
  108.244 ns (5 allocations: 608 bytes)
Time to access if X(1:3) is not interpolated constant
  107.521 ns (5 allocations: 608 bytes)
Time to just access A.data
  104.873 ns (5 allocations: 608 bytes)
Time to do A+A
  1.580 μs (6 allocations: 1.31 KiB)
Time to do A.data + A.data
  115.756 ns (1 allocation: 1.06 KiB)
```

which probably come from having keyword arguments. I remember quite fondly that keyword arguments bring performance loss if they are put into nanosecond functions.

Let's leave this open, until me or someone else finds the time to work on this (I actually don't know how to make this better, besides just remaking all keyword argument versions a 4 or 5 positional argument version instead).

rafaqz:
Maybe kwargs, otherwise its the addition of allocating the name? 

The default name should probably be `nothing` so it compiles away, you'll just have to handle it on the printing end.

I've got a fix for broadcast too :)

Datseris:
> Maybe kwargs, otherwise its the addition of allocating the name?

Oh snap, that is a good point that may be true. It would still take a bit of effort to test this, as I haven't parameterized the `name` field, and probably some changes here and there will be necessary to make it work.

rafaqz:
It would be great to get this in a release in the next few days. Let me know if you think that's possible otherwise I'll release now and do this it separately later.

Datseris:
Please don't let this hold you for a release, I can't guarantee something fast at the moment... 

Besides, doing releases with Julia is now super fast: you just increment the version in Project.toml and call JuliaRegistrator. Releasing new versions is fast enough that one shouldn't have to wait for a pending PR :)

rafaqz:
When you have a tower of dependent packages and you have to bump them all it still gets annoying. But yeah I'll bump the minor version :)

Datseris:
This makes the `name` a positional argument and restores performance to the previous value:
```
Original version:
N=10. Time to access via X(1:3)
  110.991 ns (5 allocations: 576 bytes)
N=100. Time to access via X(1:3)
  511.050 ns (5 allocations: 2.73 KiB)
N=1000. Time to access via X(1:3)
  3.843 μs (6 allocations: 23.75 KiB)

New version:
N=10. Time to access via X(1:3)
  65.542 ns (2 allocations: 416 bytes)
N=100. Time to access via X(1:3)
  368.778 ns (2 allocations: 2.58 KiB)
N=1000. Time to access via X(1:3)
  3.300 μs (3 allocations: 23.59 KiB)
```

Datseris:
@rafaqz this is good to go, the current performance is the same as the one before the PR.

Datseris:
@rafaqz sorry to ask this, but can you be the one that does the conflict resolution? You have done a lot of PRs while this was open, that I don't think anyone else had the chance to see, so it will be massively more easy and fast for you to resolve the conflicts.

BTW, since now there are other people contributing here, you can consider asking for PR reviews before merging a change. They [generally help increase the quality of the code](https://discourse.julialang.org/t/the-art-of-giving-and-receiving-code-reviews/25002).

rafaqz:
Ok conflict resolved. Looks like the error is in prettyprinting.jl. 

You can probably put the prettyprinting tests before benchmarks too, or even just remove the benchmarks from the tests as they are slow and wrong on travis anyway.

Datseris:
I've updated the tests and removed the benchmarks from travis.

The tests _fail_ at the moment here:
```
B = DimensionalArray(rand(length.(d)...), d, "test2")
C = A .+ B
```
because `C` retains the name `"test"` from `A`. This was not the case before merging with master branch, so I assume that it is because of the changes you did to increase the performance of broadcasting. Can you point me to the method that does this broadcasting? Fixing this is as simple as explicitly giving the `rebuild` result of this method the empty string as the third argument.

rafaqz:
Yep that's probably the problem:
https://github.com/rafaqz/DimensionalData.jl/blob/master/src/broadcast.jl#L45-L56

Datseris:
@rafaqz this is really, really finally good to go. During testing this I've realized there is something "wrong" with the setindex broadcasting version. If you look at the pretty printing test file, specifically the last lines where we do
```
D = DimensionalArray(ones(length.(d)...), d, "olo")
    @. D = A + B
```
the return value of `@. D = A + B` is in fact an array without the name `"olo"`. Notice that `D` itself, if printed after the operation, correctly retains the name `olo` and it also has the correct values. But something makes me think that there is an intermediate array created in the call `@. D = A + B`. However that shouldn't happen performance wise, as the broadcasting "should" unroll this operation to: 
```
for i in eachindex(D)
D[i] = A[i] + B[i]
```

anyway, this is unrelated with this PR, and I hope we can merge this PR and tag a new version asap :P

rafaqz:
That is what NamedDims does. I think it makes sense in some ways. The return value is the same as the regular operation, but the in-place value is what you would expect.

I haven't checked, but if the return value is not used it may be compiled away in some cases. But if it has a large performance cost we can change it.

PR #75, opened 2020-02-14
tweak rebuild

rafaqz:
@Datseris a few tweaks to your rebuild PR ( probably should have done during the PR...)

Name is also good to have on `AbstractDimensionalArray` instead of just `DimensionalArray` - that way it will all work in GeoData too, which already has a name field. So name is opt-out in `rebuild`: implementations can just ignore the name argument if they don't use it.
I also removed the kwarg version of `rebuild` - `ConstructionBase.setproperties` just does that automatically for all fields of any `AbstractDimensionalArray`, as magical as that sounds.

https://github.com/rafaqz/DimensionalData.jl/blob/master/src/interface.jl#L51

Lastly I cleaned up all the unnecessary parameters I had been passing to `rebuild` - we can let the optional args take care of that.

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -8,9 +8,11 @@ const StandardIndices = Union{AbstractArray,Colon,Integer}
 # Interface methods ############################################################
 
 dims(A::AbDimArray) = A.dims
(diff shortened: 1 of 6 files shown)


Datseris:
Sounds great, thanks!

PR #81, opened 2020-02-17
Better plots

rafaqz:
This PR gives correct plots for most of the major Plots.jl and StatsPlots.jl plot types.
Plots have named axes and labels from the dimension axis where it makes sense, and are permuted to have `IndependentDim` on the x axis, and similar.

Permutedims is also fixed for bugs with plotting, and more abstract types added to dimensions.


src/dimension.jl
```diff
@@ -5,26 +5,35 @@ It can also contain spatial coordinates and their metadata. For simplicity,
...
+# TODO think about the meaning of X and Y. It is slightly
+# different in different contexts: 
+# 2d plot: independent/dependent variables 
+# Some 3d plots: a spactial axis like Zdim
+# We may need to use traits to deal with this dual meaning.
+abstract type XDim{T,G,M} <: IndependentDim{T,G,M} end
+abstract type YDim{T,G,M} <: DependentDim{T,G,M} end
 abstract type ZDim{T,G,M} <: AbstractDimension{T,G,M} end
-abstract type CategoricalDim{T,G,M} <: AbstractDimension{T,G,M} end
+abstract type TimeDim{T,G,M} <: IndependentDim{T,G,M} end
```
Datseris: Nice, I'll put this to the docs PR as well.

ivirshup:
I'm curious, why define independent and dependent dimensions in the active module's scope instead of passing the dimensions when calling the plot function?

rafaqz:
Because I want it to just work most of the time without having to do that, or think about plot keyword args. `X` is an independednt dim, `Y` is dependent... its pretty standard. Can you think of any time that will be annoying or incorrect?

There will be a method for swapping the dimensions so you can do
`swapdim(A, (X, Y))` |> plot if you need to do something different.

If you just want to plot the raw data, do `collect(A) |> plot`.

I think that will cover most needs, with "smart" plotting being the default.

ivirshup:
I remember being really frustrated with how holoviews handled this. There were lots of issues of the form:

* PlotType1 needs the dim on the y-axis to be independent
* PlotType2 needs the dim on the y-axis to be depenedent

Now it's become incredibly difficult to have these plots in the same figure. They seem to have backed off making the user have to deal with this. There's also that since the dimensions are essentially defined globally, it's nice for them to not be restrictive.

I think traits like Ordinal, Categorical, Continuous are good to have (i.e. vega-lite's encodings), but I want these concepts enforced globally, instead of on a plot by plot basis.

rafaqz:
> I want these concepts enforced globally, instead of on a plot by plot basis.

What do you mean exactly? What are we doing on a plot-by-plot basis here?

>  PlotType1 needs the dim on the y-axis to be independent
    PlotType2 needs the dim on the y-axis to be depenedent

How I've written this leaves everything optional. If there is an independent and dependent dim, they go on the X and Y axes and the data might get permuted so that happens. If there isn't we just plot it as-is and label the correct axes with whatever we have. Half of it is just reducing the work of knowing which plot types need which label where.

ivirshup:
> What do you mean exactly? What are we doing on a plot-by-plot basis here?

Haha whoops, I deleted a very important word. I DON'T want these concepts enforced globally.

> How I've written this leaves everything optional. If there is an independent and dependent dim, they go on the X and Y axes and the data might get permuted so that happens. If there isn't we just plot it as-is and label the correct axes with whatever we have. Half of it is just reducing the work of knowing which plot types need which label where.

I think optional is good, but I'd like to be able to override these as well. As an example usecase, let's say I've got some time course data. Here, time is probably the independent variable. Now if I want to train a model which predicts timepoints from input data. Now, time can be a function of input data/ parameters and is dependent. Can it be easy to include predicted and imputed data on the same plot?

Datseris:
Good point @ivirshup . I have worked on timeseries prediction and I can imagine in that scenario "time" being the independent variable as well. 

However, to the best of my understanding this is solved quite easily by defining your own time dimension and making it subtype `DependentDIm` instead. This what I would do. That's the good thing with this system: it provides some basics, but all of them are completely optional. From my perspective, this is system is really good for using in multiple dispatch.

ivirshup:
If I was using the same time variable, I could just do something like: `plot(concat(real, pred), Time, Value)`, but I'm not sure what this would look like if they were different dimensions.

rafaqz:
We can probably have both without much trouble. I was also thinking a `plot(A, (X, Y))` method would help setting the plot axis order if the defaults don't work for you.

Issue #83, opened 2020-02-18
Documentation of invariants


ivirshup:
I've got some ideas about what the invariant properties of DimensionalArrays should be, but then run into cases where they aren't true. I'm then not sure whether my mental model is wrong, or if I've found a bug. It'd be good to get some consensus on this, and document these. Potentially these could also be enforced at runtime, potentially in constructors. Here's a few of the cases I can recall:

### Size of dimensions

Should this always be true?

```julia
length.(dims(da)) == size(da) == size(data(da))
```

### Broadcasting and dimension equality

Should dimensions which broadcasting is happening over have to be equal? I.e.:

```julia
a = DimensionalArray(1:3, X(1:3))
b = DimensionalArray(1:3, X(3:-1:1))

a .+ reverse(b)  # I think this should work
a .+ b  # But shouldn't this error?
```

### Equality and dimension equality

```julia
# If this is true
dims(a::DimensionalArray) != dims(b::DimensionalArray)
# Shouldn't this be true?
a != b
```

rafaqz:
> `length.(dims(da)) == size(da) == size(data(da))`

Mostly, but it isn't when the dim is just `X()`. It currently contains `Colon()` and doesn't know it's length. Indices are just passed to the array unchanged. But we could fix that if that should really always hold. `formatdims` could replace the colon with something length aware.

> Should dimensions which broadcasting is happening over have to be equal? I.e.:

Yes probably, but with some caveats. I've taken an incremental approach to implementing this. Currently we just check the dims are the same according to `basetypeof(x)`. But we should also check that the order is the same and other grid related comparisons. These comparisons are cheap and compile away. Checking the whole index matches exactly can be more expensive if you are doing something in a loop and the index is an array - I would like to be able to turn it off if we did implement something like that. We could have a `Strictness` trait somewhere.

Equality of `DimensionalArray` is just using the `AbstracAtrray` default. So a dimensional array will equal a normal array if the values match. Basically because I haven't written anything else,  and I'm not sure what the behaviour should be.

ivirshup:
> Mostly, but it isn't when the dim is just X(). It currently contains Colon() and doesn't know it's length. Indices are just passed to the array unchanged. But we could fix it if that should really always hold. formatdims could replace the colon with something length aware.

This is one of the cases that made me wonder if my thinking was right. I think it would make sense for the value to have a length. It's a nice assumption to have hold, and it makes it easy to catch bugs where dimensions don't match the data. 

[...]

PR #85, opened 2020-02-19
Invariant sizes

ivirshup:
Trying to enforce `length.(dims(da)) == size(da) == size(data(da))` following up from  #83.

This is pretty messy at the moment, but it kept bugging me throughout the day. There are a few cases that need thinking about. 

* How "unevaluated" dimensions are used. That is, does it make sense to have an instance of `X()`?
* I'm not sure how `TransformedGrids` are supposed to work. Why are the dims of an array on this grid the transformer, and not the original dimensions with the transform applied?
* `formatdims` could be cleaner
* Whyare `cov` and `cor` defined the way they are?

### DIFF
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -1,6 +1,3 @@
-# Since dims is a common argument name
-dims_alias = dims 
-
 # Array info
 for (mod, fname) in ((:Base, :size), (:Base, :axes), (:Base, :firstindex), (:Base, :lastindex))
     @eval begin
@@ -105,8 +102,8 @@ for fname in (:cor, :cov)
     @eval Statistics.$fname(A::AbDimArray{T,2}; dims=1, kwargs...) where T = begin
         newdata = Statistics.$fname(data(A); dims=dimnum(A, dims), kwargs...)
         removed_idx = dimnum(A, dims)
-        newrefdims = dims_alias(A)[removed_idx]
-        newdims = dims_alias(A)[abs(3 - removed_idx)]
+        newrefdims = $dims(A)[removed_idx]
+        newdims = $dims(A)[3 - removed_idx]
         rebuild(A, newdata, (newdims, newdims), (newrefdims,))
     end
 end
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -1,3 +1,6 @@
+# Since dims is a common argument name
+dims_alias = dims 
+
 # Array info
 for (mod, fname) in ((:Base, :size), (:Base, :axes), (:Base, :firstindex), (:Base, :lastindex))
     @eval begin
@@ -101,9 +104,10 @@ end
 for fname in (:cor, :cov)
     @eval Statistics.$fname(A::AbDimArray{T,2}; dims=1, kwargs...) where T = begin
         newdata = Statistics.$fname(data(A); dims=dimnum(A, dims), kwargs...)
-        I = dims2indices(A, dims, 1)
-        newdims, newrefdims = slicedims(A, I)
-        rebuild(A, newdata, (newdims[1], newdims[1]), newrefdims)
+        removed_idx = dimnum(A, dims)
+        newrefdims = dims_alias(A)[removed_idx]
+        newdims = dims_alias(A)[abs(3 - removed_idx)]
+        rebuild(A, newdata, (newdims, newdims), (newrefdims,))
     end
 end
 
--- a/src/methods.jl
+++ b/src/methods.jl
@@ -65,9 +65,9 @@ end
 # Dimension dropping
 
 Base._dropdims(A::AbstractArray, dim::DimOrDimType) =
-    rebuildsliced(A, Base._dropdims(A, dimnum(A, dim)), dims2indices(A, basetypeof(dim)(1)))
+    rebuildsliced(A, Base._dropdims(data(A), dimnum(A, dim)), dims2indices(A, basetypeof(dim)(1)))
 Base._dropdims(A::AbstractArray, dims::AbDimTuple) =
-    rebuildsliced(A, Base._dropdims(A, dimnum(A, dims)),
+    rebuildsliced(A, Base._dropdims(data(A), dimnum(A, dims)),
                   dims2indices(A, Tuple((basetypeof(d)(1) for d in dims))))
 
 
--- a/test/array.jl
+++ b/test/array.jl
@@ -124,12 +124,13 @@ end
     @test dims(da_float) == dims(da)
     @test refdims(da_float) == refdims(da)
 
-    da_size_float = similar(da, Float64, (10, 10))
-    @test eltype(da_size_float) == Float64
-    @test size(da_size_float) == (10, 10)
+    # TODO: Yeah, what should this be? Some dimensions (i.e. where values are not explicitly enumerated) could be resizable?
+    # da_size_float = similar(da, Float64, (10, 10))
+    # @test eltype(da_size_float) == Float64
+    # @test size(da_size_float) == (10, 10)
     # TODO what should this actually be?
     # @test dims(da_float) == dims(da)
-    @test refdims(da_float) == refdims(da)
+    # @test refdims(da_float) == refdims(da)
 end
 
 @testset "broadcast" begin
--- a/test/array.jl
+++ b/test/array.jl
@@ -87,14 +87,13 @@ b = [4 4 4 4
      4 4 4 4 
      4 4 4 4]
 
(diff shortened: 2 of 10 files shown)


rafaqz:
Sorry I wont be able to review this for a couple days.

src/methods.jl
```diff
@@ -101,9 +104,10 @@ end
 for fname in (:cor, :cov)
     @eval Statistics.$fname(A::AbDimArray{T,2}; dims=1, kwargs...) where T = begin
         newdata = Statistics.$fname(data(A); dims=dimnum(A, dims), kwargs...)
-        I = dims2indices(A, dims, 1)
-        newdims, newrefdims = slicedims(A, I)
-        rebuild(A, newdata, (newdims[1], newdims[1]), newrefdims)
+        removed_idx = dimnum(A, dims)
+        newrefdims = dims_alias(A)[removed_idx]
+        newdims = dims_alias(A)[abs(3 - removed_idx)]
+        rebuild(A, newdata, (newdims, newdims), (newrefdims,))
```
ivirshup: This is ugly, but does the right thing. Any suggestions for a cleaner implementation?

rafaqz: Maybe `getdims` instead of `dims_alias`. 

And  can you explain what `abs(3 - remove_idx)` is doing? Why do you need `abs` ?

ivirshup: Oh, since I was thinking about some other code when I wrote it. I'll change that.

ivirshup: Fixed. Also realized I could just escape `dims`.

ivirshup:
Not sure if you've started looking at this yet, but do you have a preference on dealing with conflicts? E.g. should I rebase or merge?

rafaqz:
rebase when its easy but I dont mind really.

ivirshup:
No worries, it was a pretty easy rebase.

ivirshup:
Oh, and my point about the test on the TransformedGrid. My understanding of `TransformedGrid` is that it's supposed to allow regridding operations like [xESMF](https://xesmf.readthedocs.io/en/latest/).

It doesn't make sense to me that the value of the dimension would be the transformation. It seems more like a property of the dimension's grid. Right now the test fails since the value of the dimension doesn't have `length`.

rafaqz:
Sure thtransormation can be a in grid field. The choice was pretty arbitrary.

rafaqz:
> My understanding of TransformedGrid is that it's supposed to allow regridding operations like xESMF.

Maybe eventually. Currently it's just to be able to index points using a transformed coordinate system, and to be able to actually represent gdal etc files than use a coordinate transform.

It's also not at all my area of expertise, I don't actually use these grids. But we need to be able to represent them somehow and I wanted to be sure that could happen with the existing architecture - and it can. But it's not a complete implementation.

rafaqz:
Also realised I never answered these (too busy...):

> How "unevaluated" dimensions are used. That is, does it make sense to have an instance of X()?

The instance of X() is for indexing the x dim with `Colon()`. Or for passing to `mean` etc as a dimension. 


> I'm not sure how TransformedGrids are supposed to work. Why are the dims of an array on this grid the transformer, and not the original dimensions with the transform applied?

Because then you have to compute every transform for every dim combination, and its at minimum 2d, so that also means allocating a matrix.

> formatdims could be cleaner

Absolutely. I have a rework of it in progress.

> Why are cov and cor defined the way they are?

I have no idea, but also not sure exactly what you mean

ivirshup:
Does anyone currently use the grids? I'm not sure I'm the person to redefine them, since I definitely don't use them. Would it be alright to leave the test disabled and open an issue for fixing this?

About what the dim values for transformed grid are, I was thinking something like:

DimArray `A` with dims `θ, r` is on a transformed grid that maps (X, Y) -> (θ, r). So if you access `A` with `θ(Between(τ//4, τ//2)), r(Near(3.))`, no transformation occurs. But if indexed with: `X(Near(5)), Y(Near(10))` the transformation would be applied to the coordinates being searched for before the lookup happens.

--------------------

> The instance of X() is for indexing the x dim with Colon(). Or for passing to mean etc as a dimension.

I think what I mean by this, is should `X()` and `X` mean different things? I've gone ahead and made them equivalent in the cases I've come across, since that fits the tests, but I think it's worth considering.

> format dims

Oh, I ended up doing a bit of cleanup here while fixing a bug...

> Why are cov and cor defined the way they are?

Basically, if you did `cov(DimArray(ones(5, 4), (X, Y)), dims=X)` you get an array of size (4, 4) but with two X dimensions of length five. I've changed this, but there's a comment in the tests (`tests/methods.jl`) that says:

```
    # Need to think about dims for these, currently (Y, Y) etc.
    # But you can't index (Y, Y) with dims as you get the
    # first Y both times. It will plot correctly at least.
```

And I wasn't quite sure what this meant.

rafaqz:
No-one uses the TransformedGrid as far as I know, but I use the others every day in GeoData.jl - they are what lets it abstract the array access over different base data types, with different grid orders, projections and range/array indices etc. It's the basis of a lot of my other modelling packages and the use case I wrote these DD for.

I think the transform is from some coordinates to the actual index integers, so you would only use `Between` with the untransformed values `A[X(Between(35.9, 47.4)), Y(Near(145))]`. Otherwise you can just use a unit range `A[θ(1:3), r(7)]`.

We need both `X` and `X()` because `X()` is more reliably type stable than `X`. But you should be able to use them interchangeably everywhere. I think the `X` syntax feels cleaner.

`cov`/`cor` just sounds like an error from rushing the implementation. That comment is from before I fixed #46, and can be deleted now.

ivirshup:
> No-one uses the TransformedGrid as far as I know, but I use the others every day in GeoData.jl 

Oh, I definitely just meant TransformedGrid. So, do you mind if I leave the tests for `TransformedGrid` commented out with a TODO?

> We need both X and X() because X() is more reliably type stable than X. 

Yeah, the `typeof((X, Y)) === Tuple{UnionAll, UnionAll}` had me frustrated for a bit.

> I think the X syntax feels cleaner.

Me too, that's what I've been going for. I've got another commit that makes `mean(da, dims=(X, Y))` work (for a few other reducers as well). Could I add that here, or should it be another PR?

-------------------------------

Apart from deciding on the multidimensional reduction stuff, does this need any changes?

rafaqz:
I guess we can leave TransformedGrid tests out, it feels a bit dirty but at least we'll know if anyone else is using it...

Maybe I'll merge this and you can make a new PR for the reduce methods, we should try to keep these focussed.

Also are there actual tests for invariant size?

test/array.jl
```diff
@@ -123,11 +123,11 @@ end
     @test dims(da_float) == dims(da)
     @test refdims(da_float) == refdims(da)
 
-    # TODO: Yeah, what should this be? Some dimensions (i.e. where values are not explicitly enumerated) could be resizable?
-    # da_size_float = similar(da, Float64, (10, 10))
-    # @test eltype(da_size_float) == Float64
-    # @test size(da_size_float) == (10, 10)
+    da_size_float = similar(da, Float64, (10, 10))
+    @test eltype(da_size_float) == Float64
+    @test size(da_size_float) == (10, 10)
```
ivirshup: @rafaqz This is failing at the moment since it's just trying to use the previous dims, but the new array is a different shape.

Does it make sense to just drop some info? I'm not sure I can think of an example of when I wouldn't want this to error.

rafaqz: Hmm yeah. It's good that is caught, don't think it should error... We are going to have to think about what to do in these cases. Normally `similar` provides fallbacks instead? 

That could mean anything from `Array` to a `DimensionalArray` with `EmptyDim` dims to the same dim types without an index is a possibility.

rafaqz: I would lean towards the latter. Then the only thing that breaks are `Selector`s

ivirshup: Both `OffsetArrays` and `AxisArrays` fallback to the underlying array.

I think this seems like a good option, since it might help increase compatibility. I don't think it's likely a person would intentionally call this, but it might help with compatibility for a method being called.

Also, I don't think `EmptyDim` could do this quite yet.

rafaqz: Ok lets go with `Array` for now and rethink it later. At least that will avoid the error.

ivirshup:
Travis is just failing due to build errors for the plotting tests. The relevant tests are passing.

rafaqz:
Ok this is a Plots/FFMPEG_jll problem on windows builds. We might need to disable windows testing temporarily to get this passing.

PR #86, opened 2020-02-21
Documentation

rafaqz:
Some edits and reorganisation of the docs.

### DIFF
--- a/src/primitives.jl
+++ b/src/primitives.jl
@@ -107,7 +107,8 @@ the new struct but are useful to give context to plots.
 Called at the array level the returned tuple will also include the
 previous reference dims attached to the array.
 """
-# Results are split as (dims, refdims)
+function slicedims(A, I) end
+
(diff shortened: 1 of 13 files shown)


Datseris:
I don't see the updated docs here: https://rafaqz.github.io/DimensionalData.jl/dev/

rafaqz:
Yeah its not building for some reason.

src/primitives.jl
```diff
@@ -107,7 +107,8 @@ the new struct but are useful to give context to plots.
 Called at the array level the returned tuple will also include the
 previous reference dims attached to the array.
 """
-# Results are split as (dims, refdims)
+function slicedims(A, I) end
```
Datseris: Because of this definition, if calling `sliceddims` with in correct arguments, it will not throw a method error as it should but instead return `nothing`. If you want a function definition only for the documentation string, it is better practice to do 
```julia
function slicedims end
```

instead.

PR #95, opened 2020-02-23
Update README.md

ElectronicTeaCup:
Including missing line in readme (from docs) to run example.


rafaqz:
This is no longer needed on master. I'll register an update soon.

Issue #97, opened 2020-02-24
Dimensions should define `eachindex` methods


Datseris:
```
julia> lon
dimension Longitude (type Lon):
val: Float32[0.5, 1.5, …, 358.5, 359.5]
grid: DimensionalData.AlignedGrid{DimensionalData.Ordered{DimensionalData.Forward,DimensionalData.Forward,DimensionalData.Forward},DimensionalData.Start,DimensionalData.UnknownSampling}(DimensionalData.Ordered{DimensionalData.Forward,DimensionalData.Forward,DimensionalData.Forward}(DimensionalData.Forward(), DimensionalData.Forward(), DimensionalData.Forward()), DimensionalData.Start(), DimensionalData.UnknownSampling())
metadata: nothing
type: Lon{Array{Float32,1},DimensionalData.AlignedGrid{DimensionalData.Ordered{DimensionalData.Forward,DimensionalData.Forward,DimensionalData.Forward},DimensionalData.Start,DimensionalData.UnknownSampling},Nothing}

julia> eachindex(lon)
ERROR: MethodError: no method matching keys(::Lon{Array{Float32,1},DimensionalData.AlignedGrid{DimensionalData.Ordered{DimensionalData.Forward,DimensionalData.Forward,DimensionalData.Forward},DimensionalData.Start,DimensionalData.UnknownSampling},Nothing})
Closest candidates are:
  keys(::Core.SimpleVector) at essentials.jl:602
  keys(::Cmd) at process.jl:639
  keys(::NCDataset) at C:\Users\m300808\.julia\packages\NCDatasets\wcixm\src\dataset.jl:282
  ...
Stacktrace:
 [1] eachindex(::Lon{Array{Float32,1},DimensionalData.AlignedGrid{DimensionalData.Ordered{DimensionalData.Forward,DimensionalData.Forward,DimensionalData.Forward},DimensionalData.Start,DimensionalData.UnknownSampling},Nothing}) at .\abstractarray.jl:209
 [2] top-level scope at none:0
```

Useful in general for iteration.

rafaqz:
If you want to add that and anything similar that's missing here:
https://github.com/rafaqz/DimensionalData.jl/blob/master/src/dimension.jl#L94-L111

I've been putting methods on `AbDim{<:AbstractArray}` to separate out the uses of dimensions 
with some other non array/range value in them, where things like `eachindex` don't make sense.

rafaqz:
Fixed

Issue #84, opened 2020-02-18
[Doc] Non uniform grids?


Datseris:
Hi,

I remember in some issue discussion that in general dimensions could be subtype of `AbstractArray` instead of `AbstractVector` to support non-uniform grids. 

I am trying to get my head around how to use this here or even more so how to use `Near` for such grids.

I now have an unstructured grid (equal-area Gaussian) and my lons and lats are actually just a very long vector while my "data" is also a very long vector. I.e. at every point at lon lat `lon[i], lat[i]` I have the value `A[i]`.

If using such datasets is supported here, w should somehow document this somewhere with some example of how to best use them and how to best define the dimensional array with this format. Is the best option for me to make my `A` have two dimensions: `LonLat, Time`, where the `LonLat` dimension is actually a Nx2 matrix? Then I guess using `Near` could be exlained easily?

rafaqz:
It's not really supported yet, I actually commented it out:
https://github.com/rafaqz/DimensionalData.jl/blob/master/src/grid.jl#L338-L356

But it needs to be supported for this to be widely used for geospatial data. And I haven't figured out `Near` etc either...

You _can_ use an affine transform: https://github.com/rafaqz/DimensionalData.jl/blob/master/src/grid.jl#L327

I have written it a little differently to what you describe - to keep the one axis one dim symmetry of the package (things will break otherwise). The idea is Lat and Lon hold the same matrix (one could actually be empty but it seems odd) and we just combine `dims2indices` for them - you always have to use both.

Nearly all the code to do this is written - affine transforms need a very similar setup, and they are both `<: DependenGrid`. It's just a matter of finishing it and testing it. Some of the recursive primitive methods are pretty wild...

https://github.com/rafaqz/DimensionalData.jl/blob/master/src/primitives.jl#L79-L97

Datseris:
Hm, I didn't understand most of the things you said. :(   Especially this part:

> I have written it a little differently to what you describe - to keep the one axis one dim symmetry of the package (things will break otherwise). The idea is Lat and Lon hold the same matrix (one could actually be empty but it seems odd) and we just combine dims2indices for them - you always have to use both.

Here is how I see the functionality I would like to have and I could actually put it in a PR:

One can define a dimension `Coordinates`. This dimension takes as value a `Vector{SVector}`.

[...]

PR #98, opened 2020-02-24
[WIP] Coordinates dimension

Datseris:
I am opening the PR already, as this will take a lot of discussion. I also don't know much about how the selection code works, so discussion will happen for this as well.

But first, let's see what kind of interface we want for this thing. As discussed in #84 , what I want is to be able to define a dimension (called `Coord`, long name Coordinates) that stores a vector of vectors as entry, each inner vector is the location of the point in space. With this dimension I want to be able to use the following selectors:

```julia
A[Coord(Near([ SVector(1, 2), Svector(3,5) ... ])
A[Coord(Between(SVector(-Inf, -Inf), SVector(Inf, Inf))]
```

@rafaqz requested bundling the dimensions of the coordinates in there. I like this idea, and what I would like to do is (assume that `Coord` has dimension tuple `(X, Y)`):
```julia
A[Coord(X(Between(0, 10)))]
```
to e.g. select all coordinates whose first dimension satisfies this relation. But how should this function call look like? and what If I want to select all X between something and all Y between something as well? What should be the syntax then?

---

What else about the API? @ivirshup

### DIFF
--- a/src/dimension.jl
+++ b/src/dimension.jl
@@ -5,6 +5,9 @@ It can also contain spatial coordinates and their metadata. For simplicity,
 the same types are used both for storing dimension information and for indexing.
 """
 abstract type AbstractDimension{T,G,M} end
+# T is the type of underlying values
+# G is the type of the "grid"
+# M is the type of the metadata
 
 """
 Abstract supertype for independent dimensions. Will plot on the X axis.
@@ -15,7 +18,7 @@ Abstract supertype for Dependent dimensions. Will plot on the Y axis.
 """
 abstract type DependentDim{T,G,M} <: AbstractDimension{T,G,M} end
 """
-Abstract supertype for categorical dimensions. 
+Abstract supertype for categorical dimensions.
 """
 abstract type CategoricalDim{T,G,M} <: AbstractDimension{T,G,M} end
 
@@ -108,6 +111,7 @@ Base.last(dim::AbDim{<:AbstractArray}) = last(val(dim))
 Base.firstindex(dim::AbDim{<:AbstractArray}) = firstindex(val(dim))
 Base.lastindex(dim::AbDim{<:AbstractArray}) = lastindex(val(dim))
 Base.step(dim::AbDim) = step(grid(dim))
+Base.eachindex(dim::AbDim{<:AbstractArray}) = eachindex(dim.val)
 Base.Array(dim::AbDim{<:AbstractArray}) = Array(val(dim))
 Base.:(==)(dim1::AbDim, dim2::AbDim) =
     typeof(dim1) == typeof(dim2) &&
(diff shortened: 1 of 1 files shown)


src/dimension.jl
```diff
@@ -212,9 +216,27 @@ dimmacro(typ, supertype, name=string(typ), shortname=string(typ)) =
...
 """ Ti
 
 const Time = Ti # For some backwards compat
+
+#########################################################################
+# coordinate dimensions
+#########################################################################
+struct Coord{T<:Vector{<:AbstractVector}, G, M, D} <: AbstractDimension{T,G,M}
+    val::T
+    dims::Tuple{D}
```
rafaqz: Instead of `Tuple{D}`, just use `D`, and put `D <: Tuple` in the definition above.

rafaqz:
> to e.g. select all coordinates whose first dimension satisfies this relation. But how should this function call look like? and what If I want to select all X between something and all Y between something as well? What should be the syntax then?

How about this?
```julia
A[Coord(X(Between(0, 10), Y(Between(30, 40)))]
A[Coord(Between(0, 10), Between(30, 40))]
A[Coord(At(5.5), At(9.7))]
```

We can even be tricky and make it work without using the `Coord` at all:
```julia
A[Between(0, 10), Between(30, 40)]
```

We just need a dims2indices that dispatches on a `Coord` dim with non `Coord` lookup/s.

And tor a matrix coords and a time dim:
```julia
A[Ti(1:5), Coord(At(5.5), At(9.7))]
```

Datseris:
Enabling all of these syntaxes seems like a huge amount of work, that I definitely cannot output by myself... I'll start with `A[Coord(Between(0, 10), Between(30, 40))]`, seems the most easy to do first. `A[X(Between(0, 10)), Y(Between(30, 40))]` would be great to have as well but seems too hard for me to do given the lack of background on all the details of `sel2indices`.

A comment: 
```
A[Coord(At(5.5), At(9.7))]
```
doesn't seem a good thing for me, because it could mean coordinates between 0,10 and time between 30 and 40, but also coordinate1 between 0,10 and coordinate 2 between 30,40. Coordinates can hold arbitrary amount of dimensions, and in some sense this means 1 as well.

Notice that (at least for me personally) having a DimensionalArray with _only_ coordinates is almost never the case, and one always have to put these syntaxes into perspective with the other dimensions of the array.

rafaqz:
It's about 15 lines of code. Get a basic example working and I'll make the syntax work.

> A[Coord(At(5.5), At(9.7))]

Did you quote the right one? these are both inside the `Coord` dim, so any other dims will be `Colon()`

Datseris:
Ops sorry, yes I meant to quote `A[Between(0, 10), Between(30, 40)]` ...

> That is interesting, what are often your other dimensions? 

I always have a "time" dimension, but I would imagine that in this scenario that we are discussing here people could have a "height" dimension just as well. Extremely useful are categorical dimensions as well.

> Would they ever have the same type as the `Coord` dims?

No, I can't imagine a scenario that this would make sense in the physical systems I treat.

---

Of course, I now see that even if I only had one dimension (the coordinate dimension), if we succeed in implementing all the syntaxes that are discussed here, then it would still be beneficial to bundle this dimension to a DimensionalArray with `Coords`, so that I can use `A[Y(Between(0, 90))]`. To give you some back ground, I am working on understanding the apparent hemispheric symmetries in the earth energy balance, so it is typical that I divide my datasets to north and south hemisphere. Hence the syntax `A[Y(Between(0, 90))]` is extremely useful for both grid ("standard")  as well as the Coord data.

For the categorical dimensions: I am thinking of including "surface" and "atmosphere" contributions as a categorical dimension after we are through this PR. At the moment these two contributions are in fact two different DImArrays, but I could make them one.

rafaqz:
```julia
A[Between(0, 10), Between(30, 40)]
```

Oh I can see how that is confusing. We should make it so you always have to use the `Coord` dim wrappers:
```julia
A[X(Between(0, 10)), Y(Between(30, 40))]
```

Good to hear that `A[Y(Between(0, 90))]` seems useful for both types of data! I'm liking this idea a lot now.

Datseris:
Yes, but then another thing is that we should throw an error when someone tries to make a dim array with dimensions `Coord(X,Y)` _and_ `X` used as a regular dimension. Otherwise it is impossible to use `A[X(Between(0, 10)), Y(Between(30, 40))]` due to the ambiguity of whether `X` means the component of `Coords` or the seperate dimension `X`

src/dimension.jl
```diff
@@ -108,6 +111,7 @@ Base.last(dim::AbDim{<:AbstractArray}) = last(val(dim))
 Base.firstindex(dim::AbDim{<:AbstractArray}) = firstindex(val(dim))
 Base.lastindex(dim::AbDim{<:AbstractArray}) = lastindex(val(dim))
 Base.step(dim::AbDim) = step(grid(dim))
+Base.eachindex(dim::AbDim{<:AbstractArray}) = eachindex(dim.val)
```
rafaqz: Can we do this in a different pr that just closes #97?

rafaqz:
Closed in favour of #254

PR #100, opened 2020-02-25
Make benchmarks easier to use

ivirshup:
Some low hanging fruit to make the benchmarks a bit easier to use.

Basically:

* Use the benchmarking tooling to save the results
* Use the repo status to name results files

This should make it easier to compare performance differences between commits.

### DIFF
--- a/benchmarks/benchmarks.jl
+++ b/benchmarks/benchmarks.jl
@@ -1,4 +1,4 @@
-using DimensionalData, BenchmarkTools
+using DimensionalData, BenchmarkTools, Unitful, SparseArrays, Dates, Statistics
 
 #= Benchmarks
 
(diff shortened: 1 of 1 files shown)


rafaqz:
Not sure why this is broken an master isn't...

rafaqz:
Just looking at this it would be good to regression test the benchmarks in travis somehow. Any ideas?

ivirshup:
I don't think this can run on travis. You need a consistent machine with nothing else going on to get comparable benchmarks.

There's some more information on this from:

* [BenchmarkTools](https://github.com/JuliaCI/BenchmarkTools.jl/blob/master/doc/linuxtips.md) 
* [`asv` – benchmarking tool for python](https://asv.readthedocs.io/en/stable/tuning.html)

For getting benchmarks running on another project of mine, we've gone with just adding a benchmarking machine to our grant.

PR #101, opened 2020-02-26
Type stability tests

ivirshup:
These are a bit much at the moment, but it's useful for figuring out what fails. They should probably start getting cut down pretty soon, since just running these takes a couple gigs of memory (presumably from all the compilation).

General design has been to generate the test cases, so some tooling around reporting would be useful.

Current test summary (beware):

```
Test Summary:          |  Pass  Fail  Total
indexing               | 11062   526  11588
  positional indexing  |  4169    41   4210
  dimensional indexing |  6893   485   7378
```

### DIFF
--- a/test/type-stability.jl
+++ b/test/type-stability.jl
@@ -1,0 +1,50 @@
+using Test, DimensionalData, SparseArrays, Combinatorics
+using Base: product
+
+"""Basic indexers which are valid for the provided axis"""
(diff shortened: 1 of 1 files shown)


ivirshup:
`setindex!` on the sparse array sometimes segfaults on my machine. I'm having trouble reproducing outside of the context of the test suite though, so I'm not sure who's fault it is.

Issue #111, opened 2020-04-02
error importing Dimension


chriswaudby:
Hi,

Newbie here so apologies if this is something obvious or stupid, but I can't figure out what's going on here. I've got a fresh julia installation (1.4, OS X) and have added DimensionalData, but after `using DimensionalData`, I find that `Dimension` is undefined. Everything else seems to work just fine:

```
julia> using DimensionalData

julia> Dimension
ERROR: UndefVarError: Dimension not defined

julia> X
X

julia> using DimensionalData: Dimension
ERROR: UndefVarError: Dimension not defined

julia> import DimensionalData.Dimension
WARNING: could not import DimensionalData.Dimension into Main

julia> import DimensionalData.Dim

julia> DimensionalArray()
ERROR: MethodError: no method matching DimensionalArray()
Closest candidates are:
  DimensionalArray(::A, ::D, ::R, ::String) where {T, N, D<:Tuple, R<:Tuple, A<:AbstractArray{T,N}} at /Users/chris/.julia/packages/DimensionalData/ECqye/src/array.jl:90
```

My installation status is:
```
(@v1.4) pkg> status
Status `~/.julia/environments/v1.4/Project.toml`
  [0703355e] DimensionalData v0.8.0
  [50d2b5c4] Lazy v0.15.0
  [c2839fe3] NMRTools v0.1.0 [`~/.julia/dev/NMRTools`]
  [189a3867] Reexport v0.2.0
```

Weirdly, it's a problem that seems to have arisen in the last 12 hours or so. A package that I'd been starting to develop, https://github.com/chriswaudby/NMRTools.jl, was using this fine, but when I pushed to github travis started flagging this as an error. I've reset my own julia installation and now I am also having this issue - unfortunately I didn't preserve my old state first!

Any ideas what's going on? Apologies again if this is something stupid :)

Chris

rafaqz:
I don't think `Dimension` existed in 0.8, it was `AbstractDimension`. That just changed today on master and on 0.9 with the huge PR I just merge. A lot of types have changed to make the package less verbose. Hopefully that will be the last major change to types for a while. 

0.9 isn't quite registered yet, it should be in a few hours. Until then you can use the master branch directly.

What are you trying to do with `Dimension` by the way?

rafaqz:
I realised the docs for `Dimension` are terrible. Will fix that. `Dimension` is the abstract supertype of all the dimensions. You would only use it in method dispatch.

[...]

PR #114, opened 2020-04-02
Metadata

rafaqz:
Add a metadata field to `DimensionalArray`, and methods to get the arrays metadata, or metadata for specific dims.


rafaqz:
discussed in #111

PR #123, opened 2020-04-14
bugfix plotting with dimension metadata

chriswaudby:
Attempting to plot data where axes contain metadata caused an error due to the `metadata` call being wrapped in `val`.

```julia
using DimensionalData
using Plots
x1 = 0:.1:1;
y1 = cos.(x1);
md = Dict{Any,Any}(:test=>"hello data");
md1 = Dict{Any,Any}(:test=>"hello axis 1");
d = DimensionalArray(y1, (X(x1,metadata=md1), ), metadata=md);
plot(d)
```

PS. first pull request so hope I'm doing this correctly :)


rafaqz:
Thanks

PR #147, opened 2020-08-18
fix #146 to broadcast with CUDA

### LINKED ISSUE #146: Fix broadcast over wrapped arrays like CuArray
A `DimArray` should broadcast using CUDA kernels if it wraps a `CuArray`

rafaqz:


rafaqz:
merge in later PR

PR #155, opened 2020-08-25
Update course.md

musm:


musm:
Also I get
```
ERROR: syntax: invalid named tuple field name ":time" around REPL[11]:1
``` 

when testing 

```
julia> A = DimArray(rand(10, 12), (:time=t, :distance=x))
```

rafaqz:
Nice catch! thanks

PR #157, opened 2020-08-27
Set

rafaqz:
Adds a `set` method to set most fields in the object or dimension and rebuild the whole immutable object around that. 

A fun outcome of DimensionalData being so heavily typed is you don't have to specify which field to set, you just pass in the tuple to set, the dims the are on and they will be set.

```julia
A = set(A, (Intervals(), Points()))
# or 
A = set(A, (X=>Intervals(), Y=>Points()))
# or
A = set(A, (X(Intervals()), Y(Points()))
# or
A = set(A, X=Intervals(), Y=Points())
```

Will set the the sampling mode to `Intervals` on `X`, `Points` on `Y`.


rafaqz:
Closed by #173

PR #161, opened 2020-08-29
Update examples of `otherdims` & `commondims`

singularitti:


singularitti:
BTW, is `otherdims` added only since 12.0? But it is not mentioned in [the release notes](https://github.com/rafaqz/DimensionalData.jl/releases/tag/v0.12.0).

rafaqz:
Otherdims was added with `dimwise` as it was directly needed for that, it's just also generally useful.

Now that people are starting to use this more I'll try to have things more separate in the release notes, but this is one of many packages for me and its not always efficient to do that

PR #162, opened 2020-08-29
add a tables interface

rafaqz:
@juliohm this is a basic tables interface implementation. Still need to work out what is involved to rebuild a `DimArray` from another `table`, and how much of the domain is lost in the conversions.

Also you probably need multiple layers in the table? We might need a `DimStack` object in dimensional data like there is in GeoData, then we can have multiple value columns in the table.

There is one issue with this. Often spatial datsets like netcdf etc have layers that some shared dimensions but may have more or less dimensions by layer - some have a time/vertical dimension, others don't. That could also be put into tabular form by looping the smaller layers over the time/vert dimension.


juliohm:
Nice @rafaqz , this new `DimTable` type implements the table interface and wraps the dimensional array types? Is that correct?

In order to plug it in GeoStats.jl the only missing component now is that other wrapper we discussed to get the underlying domain of the table values. I will try to dive in the code tomorrow, thanks for the quick PR.

rafaqz:
Yes. It wraps  `DimArray`, and loops the dimensions over the array stride in `DimColumn <: AbstractVector`.

I just pushed an update so you can use `dims` and all the other methods on `DimTable`, if we can get all the domain information in `dims` then it's already there.

test/tables.jl
```diff
@@ -0,0 +1,47 @@
...
+@test dimtypeof(:X) == X
+@test dimtypeof(:Ti) == Ti
+@dim Tst
+@test dimtypeof(:Tst) == Tst
+
+@test dimkey(X()) == :X
+@test dimkey(Tst()) == :Tst
+@test dimkey(Dim{:test}()) == :test 
+
+da = DimArray(ones(3, 3), (X([:a, :b, :c]), Dim{:test}(1.0:1.0:3.0)))
```
juliohm, line 15: @rafaqz do I understand correctly that the array has an unnamed property with all ones, and two coordinates X and "test" in this case?

In order to be able to digest the dimensional arrays as spatial data, we need to have (1) named properties which form a table (in this case a table with a single column with all ones), and (2) a domain like a regular grid of coordinates which is derived from the size of the array, origin and spacing. Makes sense?

Do you store the name of the layers of the arrays somewhere in DimensionalData.jl or this is a feature of the GeoData.jl package?

rafaqz: 1. There is a `name` field in all `AbstractDimArray` for the data, but it is an empty string by default. 2. The `X` and `:test` `columns are you domain, lazily presented in table form (looping with the correct appropriate stride), but also available with `dims(columns)`.

GeoData.jl stacks have to have names for all the layers, as it uses `NamedTuple` to manage them. But the stack in GeoData.jl don't necessarily share all the same dims. It might be good to introduce a stricter form that does share the same dims/domain.

juliohm: Maybe we got confused somewhere in our discussion. I had understood that we would be able to get the named layers as table columns and another wrapper type would implement the GeoDomain trait, is that your understanding as well? At the end of the day, we would like to plug a GeoData.jl array in a GeoStats.jl script and let the magic happen, right?

rafaqz: 1. No that's the plan. But these additions are general, not just for GeoStats.jl. This is a lazy table of everything in the `DimArray` worth having in a table, and you can take what you need from it and ignore what you don't need. Does that cause any issues?

2. My assumption is the `GeoDomain` trait will be generated directly and only from the `dims` `Tuple`? Is that correct?

3. If so that is a separate concern from this PR. It can't happen here as DimensionalData.jl does not depend on GeoStats.jl. You have to do that in GeoStats.jl, or we can maybe do it in GeoData.jl. Where do you think?

4. The other concern is getting multiple layers in one table, that is a requirement right? We will need something else for that, maybe `GeoStack` or a stricter version of it.

PR #176, opened 2020-09-22
Fix a `DimArray` constructor

### LINKED ISSUE #175: What is `A` here? A bug?
Should it be `data`?

https://github.com/rafaqz/DimensionalData.jl/blob/8ab9cc23fa5cd1d1d9d38cca20296ad203aa17ed/src/array.jl#L246

singularitti:
Closes #175.


rafaqz:
Oops. We should probably have a test for that too.

PR #183, opened 2020-10-07
Add `diff` method for `DimArray`s

singularitti:
This one is tricky: Julia does not allow dispatching on keyword args, so if I define `diff(A::AbstractDimArray; dims::DimOrDimType)`, it overwrites the default `diff(A::AbstractArray; dims::Integer)`. So I have to copy its definition [from base](https://github.com/JuliaLang/julia/blob/539f3ce943f59dec8aff3f2238b083f1b27f41e5/base/multidimensional.jl#L841-L850). Not the most ideal solution but it works.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -80,8 +80,8 @@ function _diff(a::AbstractArray{T,N}; dims::Integer) where {T,N}
 
     return view(a, r1...) .- view(a, r0...)
 end
-Base.diff(A::AbstractDimArray; dims) = _diff(A; dims = dimnum(A, dims))
-Base.diff(A::AbstractDimVector) = diff(A; dims = 1)
+Base.diff(A::AbstractDimArray; dims) = _diff(A; dims=dimnum(A, dims))
+Base.diff(A::AbstractDimVector) = diff(A; dims=1)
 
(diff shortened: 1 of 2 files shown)


rafaqz:
Nice. `eachslice` had the same problem. Eventually someone needs to go through base and fix all of the dim methods so they are consistent and make sense.

One minor minor thing, can we use no spaces in keyword arguments? I like to roughly follow https://github.com/invenia/BlueStyle

singularitti:
What about now?

rafaqz:
Yeah thats perfect. 

I'm also thinking we should do the test with real values instead of zeros, seeing that we have actually duplicated the code completely. You can just compare to the result of the base method on `parent(A)`.

Otherwise I really like how clean the implementation is. The dims are pretty much what you expect them to be - maybe a little weird if they are `Unordered` but otherwise good.

singularitti:
It's kind of lengthy to write the test for a 3D array, so I make it only 2D here.

singularitti:
Is this ready to be merged?

rafaqz:
Also thanks for writing that! It's been on the list since literally issue #1

PR #191, opened 2020-10-26
reverse reverses array and index, not relation

### LINKED ISSUE #190: Should `reverse` also reverse the dimension?
Hi there,

I have some new climate data, which have input latitude reverse, i.e. it starts from 90 and goes to -90 in descending order. This sometimes makes plotting weird, but anyways this doesn't matter.

I can `reverse` the array, but this does not reverse the actual dimension, so that even though the data of A are reverse along the latitude, the latitude dimension stays as is. 

Should there be a method that reverses both the array and the underlying dimension that is chosen? Perhaps a function with different name than `reverse`?

rafaqz:
closes #190

This is a minor breaking change. Indexing etc will still work the same way - but the index will be reversed whenever the array is reversed, instead of just having the `Relation` trait flipped.

`reverse(ArrayOrder, A, dims)` is available for the previous behaviour.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -5,7 +5,9 @@
 
 Reverse the array order, and update the dim to match.
 """
-Base.reverse(A::AbstractDimArray; dims=1) = reverse(ArrayOrder, A, dims)
+Base.reverse(A::AbstractDimArray; dims=1) =
+    reverse(IndexOrder, reverse(ArrayOrder, A, dims), dims)
+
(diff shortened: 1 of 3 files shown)


Datseris:
@rafaqz can you please tag a release with this in? I'd like to use it in a real project!

rafaqz:
I only use this in fake projects myself.

But this is slightly breaking and I'd prefer to do one breaking release with `At` and other things in it, needing to do release notes etc for minor releases now. N

Try to use `#master` for now, I should be able to do this in the next week some time.

PR #200, opened 2020-11-11
Fix typo in `@dim` example

ali-ramadhan:
The old example seemed a bit wrong but maybe I didn't fully understand the `@dim` docstring?


rafaqz:
Thanks. Thats the old syntax

PR #247, opened 2021-03-24
Simplify metadata

### LINKED ISSUE #245: `reverse` does not work with dimension having metadata
MWE
```julia
using DimensionalData

t = 0:0.1:1
x = 0:0.1:1
d = (Ti(t), X(x; metadata = Dict("test" => "test")))

reverse(A; dims = X)
```

rafaqz:
Metadata is too complicated. so this PR simplifies it somewhat.

Now you can just use `Metadata(a=1, b=2)` for any metadata field. But you can also do
`Metadata{:GDAL}(kw...)` or whatever to label it for dispatch, if that's required.

@Datseris this now forces the `metadata` field to be `<: AbstractMetadata`, so that error can't happen. It needs to go one way or the other, and this way is more restrictive, but less ambiguous.

Closes #245

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -62,13 +62,8 @@ NoName
 ## Metadata
 
 ```@docs
+AbstractMetadata
 Metadata
-AbstractDimMetadata
-DimMetadata
-AbstractArrayMetadata
-ArrayMetadata
-AbstractStackMetadata
-StackMetadata
 NoMetadata
 ```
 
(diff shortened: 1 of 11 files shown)


rafaqz:
@Datseris this version now allows `Dict` as metadata. `NamedTuple` I'll consider but `set` might need a small redesign first.

Datseris:
> NamedTuple I'll consider but set might need a small redesign first.

Not sure if this is worth the dev effort to be honest... Only if it helps your GPU applications to have NT instead of Dict, but from the user end, there isn't any difference.

rafaqz:
Ok that's good to hear. `Dict` is more user friendly anyway being mutable.

PR #254, opened 2021-04-19
WIP  Coord dim

rafaqz:
This is currently working using a tuple of selectors  and `Colon` to index.

Next steps 
- [x] add dimension sorting inside the `Coord` wrapper
- [ ] allow indexing without the `Coord` wrapper?

See #98 for previous discussion

### DIFF
--- a/test/coord.jl
+++ b/test/coord.jl
@@ -1,0 +1,10 @@
+using DimensionalData, Test
+
+dim = Coord([(1.0,1.0,1.0), (1.0,2.0,2.0), (3.0,4.0,4.0), (1.0,3.0,4.0)], (X, Y, Z))
+da = DimArray(0.1:0.1:0.4, dim)
+
+@test da[Coord()] === da
+@test da[Coord(:, :, :)] == [0.1, 0.2, 0.3, 0.4]
+@test da[Coord(Between(1, 5), :, At(4.0))] == [0.3, 0.4]
+@test da[Coord(:, Between(1, 3), :)] == [0.1, 0.2]
+@test index(da[Coord(:, Between(1, 3), :)], Coord) == [(1.0,1.0,1.0), (1.0,2.0,2.0)]
--- a/test/coord.jl
+++ b/test/coord.jl
@@ -1,22 +1,28 @@
-using DimensionalData, Test
+using DimensionalData, Test, Unitful
+using Unitful: s
 
 dim = Coord([(1.0,1.0,1.0), (1.0,2.0,2.0), (3.0,4.0,4.0), (1.0,3.0,4.0)], (X(), Y(), Z()))
 da = DimArray(0.1:0.1:0.4, dim)
+da2 = DimArray((0.1:0.1:0.4) * (1:1:3)', (dim, Ti(1s:1s:3s)))
 
 @testset "regular indexing" begin
     @test da[Coord()] === da[Coord(:)] === da
     @test da[Coord([1, 2])] == [0.1, 0.2]
     @test da[Coord(4)] == 0.4
+    @test da2[Coord(4), Ti(3)] ≈ 1.2
 end
 
-@testset "selector indexing" begin
+@testset "coord selector indexing" begin
     @test da[Coord(:, :, :)] == [0.1, 0.2, 0.3, 0.4]
     @test da[Coord(Between(1, 5), :, At(4.0))] == [0.3, 0.4]
     @test da[Coord(:, Between(1, 3), :)] == [0.1, 0.2]
+    @test da2[Ti(1s), Coord(:, Between(1, 3), :)] ≈ [0.1, 0.2]
 end
 
-@testset "dimension indexing" begin
+@testset "coord dimension indexing" begin
     @test da[Coord(Z(At(1.0)), Y(Between(1, 3)))] == [0.1]
 end
 
 @test index(da[Coord(:, Between(1, 3), :)], Coord) == [(1.0,1.0,1.0), (1.0,2.0,2.0)]
+
+@test bounds(da) == (((1.0, 3.0), (1.0, 4.0), (1.0, 4.0)),)
--- a/test/coord.jl
+++ b/test/coord.jl
@@ -3,9 +3,20 @@ using DimensionalData, Test
 dim = Coord([(1.0,1.0,1.0), (1.0,2.0,2.0), (3.0,4.0,4.0), (1.0,3.0,4.0)], (X(), Y(), Z()))
 da = DimArray(0.1:0.1:0.4, dim)
 
-@test da[Coord()] === da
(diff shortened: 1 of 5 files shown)


Datseris:
Does it work just as well if the values are `Vector{SVectror}` isntead of `Vector{Tuple}`?

rafaqz:
Yeah, I imagine any iterable will work.

test/coord.jl
```diff
@@ -0,0 +1,10 @@
+using DimensionalData, Test
+
+dim = Coord([(1.0,1.0,1.0), (1.0,2.0,2.0), (3.0,4.0,4.0), (1.0,3.0,4.0)], (X, Y, Z))
+da = DimArray(0.1:0.1:0.4, dim)
+
+@test da[Coord()] === da
```
Datseris, line 6:
```suggestion
@test da[Coord()] === da
@test da[Coord([1, 2])] == [(1.0,1.0,1.0), (1.0,2.0,2.0)]
```
Just add a test for safety of indexing via `Coord` without any of the selectors

rafaqz: Good idea, but maybe you mean:

```julia
da[Coord([1, 2])] == [0.1, 0.2]
```

Datseris:
This looks good to me. If you are not in a rush I can test it with my code as well by the end of the weekend.

rafaqz:
Yeah no rush, let me know how it goes. 

I just add dimension indexing, so you can e.g. leave out coords you want all of, or not know the coord order. 

It also works with a time dimension as well as the Coord, in a matrix.

Datseris:
Just tested this with my code works great no problemo. Would be happy to get a merge here and do the quick update at ClimateBase.jl. Regarding the point: 

> allow indexing without the Coord wrapper?

Yes, that'd be great to be able to do `X[Lon(Between(...))]` instead of wrapping it to `Coord`. This would disallow having an additional dimension that is also the mode of the `Coord`, but I think it would be nonsense to do it anyways. Why have both a `Coord` dim with `(Lon, Lat)` mode, but also an additional `Lon` dimension? Doesn't make sense, right?

Datseris:
Ah, another suggestion. `CoordMode` is printed just as `CoordMode`, but I think it would be helpful if it also displayed the wrapped dimensions.

Datseris:
Yet another suggestion: would be nice if `Coord` has a long name `Coordinates` just like `Lon` has a long name `Longitude`.

rafaqz:
Great, glad it works. It should totally have a custom `show` method too. And having two lat dims would actually still work (it would be ordered), but yes it would be weird to do that.

I might merge this and those things can be done later, as I'm not sure I'll get back to it for a week or so.

Datseris:
sure, sounds fair, just be sure to tag this in

Datseris:
@rafaqz not sure if this is of any use for you, but I typically do a performance optimization that **relies that the coordinates are sorted**. In my case they are sorted by the _second_ dimension, which is latitude. That is because it is simply more frequent that a user will do searches by latitude than by longitude or average towards the latitude dimension. In any case I have defined:


```julia
"""
    uniquelats(A::AbDimArray) → idxs, lats
    uniquelats(c::Vector{<:AbstractVector}) → idxs, lats
Find the unique latitudes of `A`. Return the indices (vector of ranges) that each latitude
in `lats` covers, as well as the latitudes themselves.
"""
uniquelats(A::AbDimArray) = uniquelats(dims(A, Coord).val)
function uniquelats(c)
    @assert issorted(c; by = x -> x[2])
    idxs = Vector{UnitRange{Int}}()
    lats = eltype(eltype(c))[]
    iprev = 1
    for i in 2:length(c)
        if c[i][2] != c[i-1][2]
            push!(lats, c[i-1][2])
            push!(idxs, iprev:(i-1))
            iprev = i
        end
    end
    push!(lats, c[end][2])
    push!(idxs, iprev:length(c))
    return idxs, lats
end

function coord_latitudes_between(c, l1, l2)
    idxs, lats = uniquelats(c)
    i1 = searchsortedfirst(lats, l1)
    i2 = searchsortedlast(lats, l2)
    i1 > i2 && ((i1, i2) = (i2, i1)) # in case bounds are given in reverse order
    return idxs[i1][1]:idxs[i2][end]
end

function DimensionalData.sel2indices(c::Coord, sel::Tuple{<:Lat{ <: Between}})
    l1, l2 = sel[1].val.val
    return coord_latitudes_between(c, l1, l2) # this is Vector{Int}
end
```

Just as an idea. On my machine I do get considerable speedup when using this method:
```
julia> @time I[Coord(Lat(Between(0, 30)))];
  0.000779 seconds (18 allocations: 1.854 MiB)

julia> @time I[Coord(Lon(Between(0, 30)))];
  0.003520 seconds (24.63 k allocations: 1.928 MiB)
```
for a 8200×242 Matrix{Float32} array (8200 is the length of the `Coord` dimension)

Because this relies on the assumption that one dimension is sorted, might not be useful for the generic package DimData.jl, but perhaps useful in your personal code.

rafaqz:
I'm all about using `searchsortedfirst` - I just made all the other indexing here (i.e. with `At`) use it where the order is known, which I should have done ages ago. And I think this is a key optimization for any use of `Coord`.

The issue is it might be slower in some contexts (e.g. indexing a short axis is measured in nanoseconds anyway). We could also do this sorting up-front, and keep the lat order after use after `Between`? If I understand how this works? Done by marking with a trait which dimension is sorted.

We already have a lot of the machinery for this - we just need to set an `IndexOrder` somewhere (`ForwardIndex`/`ReverseIndex` etc) as with regular dimensions. It could just be set in the mode of the internal dimensions inside `Coord`? maybe with a helper kw so its easier to do. Although probably that will be set by packages most of the time anyway.

So in your case the first dim would be `Unordered` while the second would be `ForwardIndex`. Or something.

Datseris:
> The issue is it might be slower in some contexts (e.g. indexing a short axis is measured in nanoseconds anyway). We could also do this sorting up-front, and keep the lat order after Between? If I understand how this works? Done by marking with a singleton which dimension is sorted.

Yeap that's how I do it, upon load arrays are sorted by the second dimension of `coord` (which is latitude)
```
A # some `ClimArray` with unstructured grid type.
coords = dims(A, Coord).val
si = sortperm(coords, by = reverse)
A = A[Coord(si)]
```

Having some kind of index for it for the generic case sounds good as well. 

> We already have a lot of the machinery for this - we just need to set an IndexOrder somewhere (ForwardIndex/ReverseIndex etc) as with regular dimensions. It could just be set in the mode of the internal dimensions inside Coord? maybe with a helper kw so its easier to do. Although probably that will be set by packages most of the time anyway.

> So in your case the first dim would be Unordered while the second would be ForwardIndex. Or something.

Yes that sounds like exactly what's going on. Good to hear that it might be quite straightforward to adjust.

PR #271, opened 2021-05-20
Pretty printing for tuples of dimensions

ali-ramadhan:
Resolves #261 
Replaces PR #262

PR #272, opened 2021-05-20
Bump v0.17.5

ali-ramadhan:
Sorry forgot to do this in PR #271


rafaqz:
No it's better not to do it in the PR, it might sit around for a while... And I have the tendency to not remember where a bump was registered so having a bump commit with a registration comment on it makes things really clear and easy to find.

rafaqz:
Also, doing it this way ensures the version bump actually occurs on the tag commit.

ali-ramadhan:
Haha good points, it was a happy mistake then.

PR #316, opened 2021-09-25
Fixed issues with Julia 1.7

mauro3:
Sorry, there are a lot of whitespace differences too! (you could use  `git diff --ignore-all-space`)

Also, I added a few other Julia versions to the tests as some of them failed locally.  Maybe remove them again before merging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
       matrix:
         version:
           - '1.6'
+          - '^1.7.0-rc1'
         os:
           - ubuntu-latest
           - macOS-latest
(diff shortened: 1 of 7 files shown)


rafaqz:
Thanks for the fix. Those spaces need deleting anyway.

The reason for only testing 1.6 was doing that much testing for everything can really slow your workflow. But its probably sensible to at least test 1.7.

mauro3:
Do you know what those test failures are about: "The operation was cancelled" https://github.com/rafaqz/DimensionalData.jl/pull/316/checks?check_run_id=3709600819.  Note this also happened on my fork where CI ran as well.  I will push with fewer test Julia versions, hopefully it will run then.

Note, I also adjusted the Julia version to fix #317.

mauro3:
Ok, now the test seem to run https://github.com/mauro3/DimensionalData.jl/runs/3711843233?check_suite_focus=true

rafaqz:
Thanks, looks good

PR #319, opened 2021-09-27
IndexMode as Lookup vectors, Dims as simple wrappers, Reduced Interface

### LINKED ISSUE #264: Dimensions after reduction have the wrong coordinates
It seems that dimensions want to retain their types so after taking a `mean` I expected the new dim to be `X: Float64[2.5]` since `mean(1:4) == 2.5` but instead it is `X: Int64[3]`.

Is this intentional for type stability or is it a bug?

```julia
julia> using Statistics, DimensionalData

julia> da = DimArray(randn(4), X(1:4))
4-element DimArray{Float64,1} with dimensions: 
  X: 1:4 (Sampled - Ordered Regular Points)
 -1.22385
 -1.35615
  1.09573
  1.43142

julia> mean(da, dims=X)
1-element DimArray{Float64,1} with dimensions: 
  X: Int64[3] (Sampled - Ordered Regular Points)
 -0.0132152
```

### LINKED ISSUE #282: getindex/view on a Dimension should return a dimension?

### LINKED ISSUE #321: Simplify dimension order
This is long over due. Mostly the complexity is not necessary, or can be worked around.

rafaqz:
A large refactor of DimensionalData.jl internals

- `Dimension`s are real wrapper types now. They only have one field.
- `IndexMode` is now called `LookupArray`, and it wraps the lookup index, type metadata and the metadata field. `LookupArray <: AbstractArray` so these can be treated as regular vectors.
- `LookupArray` is decoupled from `Dimension`. Someone could use them in another package, or we could put them in their own package just for fast lookup indexes that can represent a lot of index properties.
- indexing a `Dimension` returns a `Dimension`, or the value for an `Int`
- indexing a `LookupArray` returns a `LookupArray`, or the value for an `Int`
- makes some minor changes to how reduction works on the new Lookup compared to the old dimension index.
- remove `Val` index. It may return wrapped as a vector, but cant work with the current design.
- simplify order. now there is just `ForwardOrdered`, `ReverseOrdered`, `Unordered` for the index. 
- reduce the exported interface. This may be VERY BREAKING.
- Use submodules for `Dimensions`  and `LookupArrays` - importing these with `using` returns most removed methods to scope.

It seems simpler and more logical to me. Its a major breaking change for low level details, but doesn't change regular syntax very much, if at all. @mcabbott this is largely following your recent gripe, thanks for the feedback.

Closes #282 
Closes #264 
Closes #321


mcabbott:
Have not read closely, but this sounds good! Thanks for the mention... was pretty grumpy that day, sorry, but am very glad to hear if good things sometimes come from gripes.

Datseris:
So, the syntax of e.g. `values = dims(A, Lon).val` does not work anymore after this PR I assume? The correct way to do this would be `values = DimensionalData.value(dims(A, Lon))` ?

Datseris:
@rafaqz can you please answer here? I just want to be sure that I'm not missing something, because at the moment the only way I see to obtain a dimension's values (I guess they are called look-ups now) is by doing:
```
x = dims(X, Lon).val.data
```
right? Or is there a simple, single function that can be used instead?

rafaqz:
You can use `parent(dim)` to get the lookup. Or `parent(parent(dim))` to get the inner range/vector. 

There is also `DD.lookup` and `LookupArrays.index` that will get the inner lookup or range for all dims. Not exported as DD exported too much and `parent` is in Base and is mostly enough.

Datseris:
@rafaqz thanks, a final question then: now, the proper way to create a dimension with metadata is the following ?
```
Dim(LookupArray(values; metadata = metadata))
```
?

Before I did just `Dim(values; metadata = metadata)`

Datseris:
Ah, no, I have another question. Seems like now `sel2indices` is gone. Is there still a way to do the following?
```julia
# This modifies what happens on A[Coord(Lat(Between(x,y)))]
function DimensionalData.sel2indices(c::Coord, sel::Tuple{<:Lat{ <: Between}})
    l1, l2 = sel[1].val.val
    return coord_latitudes_between(c.val, l1, l2) # this is Vector{Int}
end
```

rafaqz:
Your old solution should still work, there us a kwargs method for dimensions that passes them to the lookup constructor when lookup properties are autodetected in `format(dims, A)` that runs on array construction. But for a package you can do it manually using e.g. `Sampled(values; order, sampling, span, metadata)`, which will also be more type stable by specifying the order explicitly.

rafaqz:
It `selectindices` now. LookupArrays may end up its own package and its a better name to export.

PR #330, opened 2021-10-18
Constructor w/ AxisKeys/AxisArrays syntax

ParadaCarleton:
This PR adds a constructor that implements the same syntax as AxisKeys and AxisArrays, which should make switching over from those easier. By AxisKeys/AxisArrays syntax I mean:
`DimArray(data; dim_name = dim)`


rafaqz:
Closing as there is no viable way this syntax can work without breaking extensibility. Hi hope `NamedTuple` is enough @ParadaCarleton

PR #351, opened 2021-12-31
Use IntervalSets

rafaqz:
This PR begins the move from `Between` to `Interval` from IntervalSets.jl

This means: 
- We can use elipses `a..b` instead of `Between(a, b)` which saves a bunch of code
- We can use `OpenInterval(a, b)` or `Interval{:closed,:open}` etc from IntervalSets
- Between can be depreciated, so it stops clashing with DataFrames.jl


Datseris:
I like a lot the usage of interval sets. I also like a lot that now the default interval is closed on both ends, always made my work a bit inconvenient to have the ending open instead of closed :D

src/LookupArrays/selector.jl
```diff
@@ -325,6 +327,13 @@ _searchfunc(::End, ::ReverseOrdered) = searchsortedlast
 
     Between(a, b)
 
+
+Depreciated: use `a..b` instead of `Between(a, b)`. Other `Interval`
```
Datseris, line 331: notice that the correct word is "Deprecated"

PR #358, opened 2022-01-05
Fix #357

### LINKED ISSUE #357: Issue with `_matches` with `Coord` dimension
I'm getting 
```
ERROR: MethodError: no method matching _matches(::IntervalSets.ClosedInterval{Int64}, ::Float32)
Closest candidates are:
  _matches(::DimensionalData.Dimensions.Dimension, ::Any) at C:\Users\m300808\.julia\packages\DimensionalData\Ma3Hp\src\Dimensions\coord.jl:73
```
when doing ` C[Coord(Lat(-86..(-82)))]`, which is a behavior I myself extended by extending
```julia
function DimensionalData.selectindices(c::Coord,
    sel::Tuple{<:Lat{ <: DimensionalData.LookupArrays.IntervalSets.AbstractInterval}})
    l1 = sel[1].val.left; l2 = sel.val.right
    return coord_latitudes_between(gnv(c), l1, l2) # this is Vector{Int}
end
```

This is a nice method I'd like to have, because I do some really nice optimizations in this `coord_latitudes_between` function. The problem I think is here:
```julia
_matches(sel::Dimension, x) = _matches(val(sel), x)
_matches(sel::Between, x) = (x >= first(sel)) & (x < last(sel))
_matches(sel::At, x) = x == val(sel)
_matches(sel::Colon, x) = true
_matches(sel::Nothing, x) = true
```
which does not contain a definition for interval sets. The fix is simple I believe.

Datseris:
Fixes #357

TODO: tests. EDIT: Not sure how to test this without extending this fancy indexing.

### DIFF
--- a/src/Dimensions/coord.jl
+++ b/src/Dimensions/coord.jl
@@ -72,7 +72,7 @@ dims(d::Coord) = dims(val(d))
 
 _matches(sel::Dimension, x) = _matches(val(sel), x)
 _matches(sel::Between, x) = (x >= first(sel)) & (x < last(sel))
-_matches(sel::AbstractInterval, x) = x ∈ sel
+_matches(sel::LookupArrays.AbstractInterval, x) = x ∈ sel
 _matches(sel::At, x) = x == val(sel)
 _matches(sel::Colon, x) = true
 _matches(sel::Nothing, x) = true
--- a/src/Dimensions/coord.jl
+++ b/src/Dimensions/coord.jl
@@ -72,7 +72,7 @@ dims(d::Coord) = dims(val(d))
 
 _matches(sel::Dimension, x) = _matches(val(sel), x)
 _matches(sel::Between, x) = (x >= first(sel)) & (x < last(sel))
-_matches(sel::LookupArrays.AbstractInterval, x) = x ∈ sel
+_matches(sel::LookupArrays.AbstractInterval, x) = x in sel
 _matches(sel::At, x) = x == val(sel)
 _matches(sel::Colon, x) = true
 _matches(sel::Nothing, x) = true
--- a/src/Dimensions/coord.jl
+++ b/src/Dimensions/coord.jl
@@ -72,6 +72,7 @@ dims(d::Coord) = dims(val(d))
 
 _matches(sel::Dimension, x) = _matches(val(sel), x)
 _matches(sel::Between, x) = (x >= first(sel)) & (x < last(sel))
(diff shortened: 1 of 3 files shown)


src/Dimensions/coord.jl
```diff
@@ -72,6 +72,7 @@ dims(d::Coord) = dims(val(d))
 
 _matches(sel::Dimension, x) = _matches(val(sel), x)
 _matches(sel::Between, x) = (x >= first(sel)) & (x < last(sel))
+_matches(sel::LookupArrays.AbstractInterval, x) = x ∈ sel
```
rafaqz, line 75: Can you use `in` here?

Datseris, line 75:
```suggestion
_matches(sel::LookupArrays.AbstractInterval, x) = x in sel
```

rafaqz:
To test I think you can just define an array with a coordinate dim and use an interval to select the coords?

Datseris:
Unfortunately I get `ERROR: ArgumentError: Cannot use an interval or `Between` with Unordered`. I've pushed the test here for reference.

rafaqz:
Your Interval syntax is odd... I don't think there is code to handle intervals between tuples?

Why not just swap this test from between to intervals, or have both:

```julia
@test da[Coord(Z(At(1.0)), Y(Between(1, 3)))] == [0.1]
@test da[Coord(Z(At(1.0)), Y(1..3))] == [0.1]
```
or like
```julia
@test da[Coord(1..5, 10..20, 8..9))] == ...
```

Datseris:
Is it really an odd syntax? I think it is perfectly valid:
```julia

julia> int = (0.1, 0.1)..(0.2, 0.2)
(0.1, 0.1)..(0.2, 0.2)

julia> (0.15, 0.15) ∈ int
true

julia> (0.25, 0.15) ∈ int
false
```
It simply expands each dimension.

But okay, for dimensional data it makes more sense to think in terms of individual dimensions, so I added the test you requested, which indeed passes.

rafaqz:
Sure, but we would still need to handle that here as an individual selector for multiple dimensions... DD doesn't know that exists.

Datseris:
I would appreciate this PR merged and tagged, extra features can go in another PR if possible. I'm currently satisfied with this change anyways :)

Datseris:
@rafaqz thanks for merging this, can you please also tag a new patch release that contains this?

PR #366, opened 2022-03-21
Add `All` selector

### LINKED ISSUE #342: Is a Union selector possible?
Hi,
In my work I very often do the operation "Select latitude -90 to -30 and 30 to 90", but no matter the format of my data, I always end up needing to write a custom function for such a selection because there is no combination of `At` or `Between`  etc. that just works.

I am actually hoping that I am missing something and I can actually write `A[UnionSelector(...)]` so that I keep using pure DimensionalData.jl syntax instead. But I am not sure. 

So I was wondering whether it is of any worth to consider using some kind of `Union` selector?

rafaqz:
Closes #342 

@Datseris see if this works for you as a union selector.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,7 +1,7 @@
 using Pkg
 Pkg.activate(@__DIR__)
 
-using Documenter, DimensionalData, CoordinateTransformations, Dates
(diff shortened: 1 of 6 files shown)


Datseris:
It does!

PR #389, opened 2022-08-03
Show lookup labels with AbstractDimArray

### LINKED ISSUE #318: Display Axis Labels Alongside Columns/Rows?
Currently, DimensionalData.jl displays axis labels separately from the data itself. This makes it hard to tell what each column means when a DimArray is displayed; it would be nice to have column labels displayed clearly above each column.

rafaqz:
Closes #318

@Datseris @ParadaCarleton seeing you both expressed interest in #318, do you want to try this out and see how you like the style? It uses the `:light_black` color for labels which I think is minimally intrusive when you just want to look at the array, besides the slight loss of screen space.

It has a few problems - the labels use `Base.alignment` which means they try and align with the other array values. Which is a bit of a hack, they should just sit in whatever space the array values fill - but things get much trickier to do that.

The other downside is that array types with specific `Base.print_matrix` methods don't use those methods - a `SparseArray` isprinted with regular zeros instead of a dot for empty values.

The github output loses colors so looks a lot worse than in the REPL, but to demonstrate:

```julia
julia> rand( X(200.0:1:400), Y('a':'z'), Z(5))
201×26×5 DimArray{Float64,3} with dimensions:
  X Sampled 200.0:1.0:400.0 ForwardOrdered Regular Points,
  Y Categorical 'a':1:'z' ForwardOrdered,
  Z
[:, :, 1]
         'a'       'b'        'c'         'd'      …   'w'         'x'       'y'        'z'
 200.0  0.673935  0.141439   0.955314    0.717369     0.740125    0.955894  0.826375   0.44628
 201.0  0.712164  0.481134   0.542074    0.67757      0.617952    0.282931  0.540955   0.450101
 202.0  0.648451  0.650928   0.408938    0.150393     0.342722    0.775194  0.414076   0.860255
 203.0  0.936581  0.469827   0.441755    0.243855     0.856957    0.583006  0.741963   0.839532
 204.0  0.353114  0.0873615  0.446055    0.678356  …  0.592797    0.305778  0.881453   0.129537
 205.0  0.914772  0.399673   0.00789701  0.22394      0.161127    0.849938  0.0727497  0.90776
   ⋮                                               ⋱                        ⋮
 395.0  0.993582  0.987368   0.711423    0.949472     0.00607777  0.699784  0.0920469  0.802377
 396.0  0.855686  0.887951   0.683934    0.720047  …  0.639821    0.671521  0.200358   0.371284
 397.0  0.27446   0.404087   0.783541    0.198848     0.275439    0.084968  0.886439   0.248281
 398.0  0.542105  0.206849   0.622234    0.425789     0.451899    0.96857   0.71085    0.430552
 399.0  0.662324  0.336501   0.608798    0.320062     0.0179621   0.814293  0.91817    0.667375
 400.0  0.456497  0.634181   0.687159    0.717509     0.670711    0.365766  0.313009   0.856011
[and 4 more slices...]
```

### DIFF
--- a/src/Dimensions/dimension.jl
+++ b/src/Dimensions/dimension.jl
@@ -38,9 +38,10 @@ A = DimArray(zeros(3, 5, 12), (y, x, ti))
   X Sampled 2:2:10 ForwardOrdered Regular Points,
   Ti Sampled DateTime("2021-01-01T00:00:00"):Month(1):DateTime("2021-12-01T00:00:00") ForwardOrdered Regular Points
 [:, :, 1]
- 0.0  0.0  0.0  0.0  0.0
- 0.0  0.0  0.0  0.0  0.0
- 0.0  0.0  0.0  0.0  0.0
+       2    4    6    8    10
+  'a'  0.0  0.0  0.0  0.0   0.0
+  'b'  0.0  0.0  0.0  0.0   0.0
+  'c'  0.0  0.0  0.0  0.0   0.0
 [and 11 more slices...]
 ```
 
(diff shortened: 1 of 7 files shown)


Datseris:
For me this looks cool, but I am wondering, couldn't you use the Tables.jl interface, to utilize their printing, that basically already does all this, and save a lot of code writing here? So step 1 is make a table with the first slice, stept 2 print the table, step 3 add the info "and x more slices.."

Datseris:
I don't think the second downside is really that much of a bummer by the way, depending on whetehr the sparse type is somehow shown. E.g., does `DimArray{Float64}` change if the internal parent is sparse?

rafaqz:
Yeah Tables.jl is another option. But most of the work here is actually still in Base `print_matrix`. The reason to use it instead of Tables.jl is the output looks the most similar to showing a regular `Array`

rafaqz:
> I don't think the second downside is really that much of a bummer by the way, depending on whetehr the sparse type is somehow shown. E.g., does `DimArray{Float64}` change if the internal parent is sparse?

It actually doesn't change... It might be nice to show what the parent array is though, that's a good idea.

PR #407, opened 2022-09-04
Allow String metadata keys

### LINKED ISSUE #406: Metadata with String as key?
It seems that an error (see below) occurs when the key is a `String` for `Metadata{Rasters.NCDfile}`.  When a `Metadata` is created, its type is either `{Symbol, String}` or `{String,String}` depending on how it's created.  But, when it's `{String,String}`, an additional key-value pair with a `String` key is refused.

Because the netCDF standard

https://docs.unidata.ucar.edu/netcdf-c/current/programming_notes.html#object_name

allows for some special characters which aren't allowed for Julia identifiers, `String`s are a bit more convenient than `Symbol`s as the name of an attribute.

```julia
using DimensionalData: Metadata
using Downloads
using Rasters

const NC_meta = Metadata{Rasters.NCDfile}

const url = "https://www.unidata.ucar.edu/software/netcdf/examples/sresa1b_ncar_ccsm3-example.nc"
infile = Downloads.download(url, "tmp.nc")
#infile = "tmp.nc"

function sym_sym(ua)
  v = rebuild(similar(ua[:,:,1,1]); name=:v, metadata=NC_meta(:history => "sample"))
  display(metadata(v))
  metadata(v)[:units] = "someunits"
  v
end

function sym_str(ua)
  v = rebuild(similar(ua[:,:,1,1]); name=:v, metadata=NC_meta(:history => "sample"))
  display(metadata(v))
  metadata(v)["units"] = "someunits"
  v
end

function str_sym(ua)
  v = rebuild(similar(ua[1,:,:,1]); name=:v, metadata=NC_meta("history" => "hello"))
  display(metadata(v))
  metadata(v)[:units] = "otherunits"
  v
end

function str_str(ua)
  v = rebuild(similar(ua[1,:,:,1]); name=:v, metadata=NC_meta("history" => "hello"))
  display(metadata(v))
  metadata(v)["units"] = "otherunits"
  v
end

ua = Raster(infile; name = "ua")

# ERROR: LoadError: MethodError:
#    Cannot `convert` an object of type Symbol to an object of type String
#v = sym_sym(ua) # fine. Dict{Symbol, String}
#v = sym_str(ua) # fine. Dict{Symbol, String}
#v = str_sym(ua) # error. Dict{String, String}
v = str_str(ua) # error. Dict{String, String}
write("tmp2.nc", v)
```

rafaqz:
Closes #406

### DIFF
--- a/src/LookupArrays/metadata.jl
+++ b/src/LookupArrays/metadata.jl
@@ -17,9 +17,9 @@ const _MetadataContents =Union{AbstractDict,NamedTuple}
 const AllMetadata = Union{AbstractMetadata,AbstractDict}
 
 Base.get(m::AbstractMetadata, args...) = get(val(m), args...)
-Base.getindex(m::AbstractMetadata, key) = getindex(val(m), Symbol(key))
-Base.setindex!(m::AbstractMetadata, x, key) = setindex!(val(m), x, Symbol(key))
-Base.haskey(m::AbstractMetadata, key) = haskey(val(m), Symbol(key))
+Base.getindex(m::AbstractMetadata, key) = getindex(val(m), key)
+Base.setindex!(m::AbstractMetadata, x, key) = setindex!(val(m), x, key)
+Base.haskey(m::AbstractMetadata, key) = haskey(val(m), key)
 Base.keys(m::AbstractMetadata) = keys(val(m))
 Base.iterate(m::AbstractMetadata, args...) = iterate(val(m), args...)
 Base.IteratorSize(m::AbstractMetadata) = Base.IteratorSize(val(m))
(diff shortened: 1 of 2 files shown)


Datseris:
I'm trying to figure out why this is a breaking release, but I can't! Any help please?

rafaqz:
It's tiny, but previously `String` keys would be converted to `Symbol`, now they won't be because the keys might actually be `String`. It breaks Rasters.jl at least.

Datseris:
Ah no worries, version bumps don't worry me :P I know the come with improvement to the software so its a pleasant pain :D

Issue #423, opened 2022-11-18
keyword indexing fails for 0-dimensional DimArray


sethaxen:
```julia
julia> using DimensionalData

julia> x = DimArray(fill(3), ());

julia> y = DimArray(randn(2, 3), (Dim{:a}(1:2), Dim{:b}(1:3)));

julia> y[c=1]  # for a >0-dim array, indexing a missing dim just raises a warning
┌ Warning: (Dim{:c},) dims were not found in object
└ @ DimensionalData.Dimensions ~/.julia/packages/DimensionalData/K9D4P/src/Dimensions/primitives.jl:645
2×3 DimArray{Float64,2} with dimensions: 
  Dim{:a} Sampled{Int64} 1:2 ForwardOrdered Regular Points,
  Dim{:b} Sampled{Int64} 1:3 ForwardOrdered Regular Points
     1          2         3
 1   0.295756   1.10136   0.855181
 2  -0.963257  -0.4447   -0.95428

julia> x[c=1] # but for 0-dim we StackOverflow
ERROR: StackOverflowError:
Stacktrace:
 [1] dims2indices(x::Tuple{}, I::Tuple{Dim{:c, Int64}}) (repeats 79984 times)
   @ DimensionalData.Dimensions ~/.julia/packages/DimensionalData/K9D4P/src/Dimensions/indexing.jl:34
```

sethaxen:
I thought what was missing was the following method
```julia
julia> @inline Dimensions.dims2indices(::Tuple{}, I) = ()
```

But this has an unintended result. Indexing by a missing keyword then behaves like `x[]`:
```julia
julia> x[c=1]
3

julia> ds = DimStack((; x, y));

julia> ds[c=1]
┌ Warning: (Dim{:c},) dims were not found in object
└ @ DimensionalData.Dimensions ~/.julia/packages/DimensionalData/K9D4P/src/Dimensions/primitives.jl:645
(x = 3, y = [-1.5090420829521751 -0.629637266507813 -1.1542533262114874; 1.78624724297272 -1.341032743849202 0.2629446089245767])
```

I would expect `x[c=1]` to return `x` as a 0-dimensional array with a warning, and similarly I would expect indexing with `ds` to return the dataset unchanged.

rafaqz:
It is pretty weird I agree.  But I would also have defined your missing method, expecting the result you got.

[...]

PR #426, opened 2022-11-19
Fixes for viewing all dims of array or stack

sethaxen:
This PR fixes the issues with `view` raised in #423

There are some errors in the `single-arity views` testset that I don't understand yet.

### DIFF
--- a/test/selector.jl
+++ b/test/selector.jl
@@ -1024,13 +1024,16 @@ end
             4:-2:1,
         ]
         for idx in indices
-            idx = 3
             from2d = view(da, idx)
             @test from2d == view(parent(da), idx)
-            @test from2d isa SubArray
+            if idx isa Integer
+                @test from2d isa DimArray
+            else
+                @test from2d isa SubArray
+            end
             from1d = view(da[Y(At(10))], idx)
             @test from1d == view(parent(da)[1, :], idx)
-            @test from1d isa SubArray
+            @test from1d isa DimArray
         end
     end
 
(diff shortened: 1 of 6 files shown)


sethaxen:
The failing tests look broken. `idx` is set for each loop iteration but then always set to an `Int` internally:
https://github.com/rafaqz/DimensionalData.jl/blob/70f2aefc6c4248a0ab957f51cf8b415f9799a39e/test/selector.jl#L821-L837

sethaxen:
The only remaining error seems to be related to uploading the coverage report to Codecov on OS X, e.g. https://github.com/rafaqz/DimensionalData.jl/actions/runs/3560347385/jobs/5980312809#step:7:47

rafaqz:
Yeah that's happening everywhere now, but not sure what the issue is.

rafaqz:
You may need to rebase this against main for the codecov fix

test/selector.jl
```diff
@@ -1016,24 +1016,6 @@ end
         end
     end
 
-    @testset "single-arity views" begin
```
rafaqz, line 1019: Are we losing code coverage from deleting these?

rafaqz (commented):
LGTM, just wondering if deleting those tests is a good ides...

test/selector.jl
sethaxen: It doesn't seem so (diff coverage is 100% and total coverage is increased), but I've reverted the removal and instead fixed the tests.

rafaqz (approved):
Thanks for the PR!

PR #434, opened 2022-12-03
Stack getproperty

rafaqz:
I've meant to do this for ages, but the idea is to make stacks more like `NamedTuple` so that `getproperty` gets layers by name like `stack.layer`. ConstructionBase.jl also works on the fields, so you can do `Setfield.@set stack.layer = somedimarray`

This may effect existing code, any uses of `stack.dims` or similar should use `dims(stack)` and any custom fields  `getfield(stack, :field)` in an accessor method.

Objects that extend `AbstractDimStack` may need a few changes. Now that `ConstructionBase.setproperties` has been modified, all `<: AbstractDimStack` constructors will need a method that accepts `NamedTuple` of `AbstractDimArray`. We also need to define `rebuild` with keyword arguments now as ConstructionBase wont give us this version for free anymore:

```julia
function rebuild(s::MyDimStack; data=data(s), dims=dims(s), refdims=refdims(s),
    layerdims=layerdims(s), metadata=metadata(s), layermetadata=layermetadata(s)
    other_custom_kw=other_custom_kw(s),
)
    basetypeof(s)(data, dims, refdims, layerdims, metadata, layermetadata)
end
```

@sethaxen, @Datseris  and @sefffal this may effect your packages if / where you extend `AbstractDimStack`. Comments or reviews appreciated, and otherwise just a heads up that there will be a minor bump with this change arriving soon.


sethaxen:
Awesome! I already had modified my `AbstractDimStack` subtype to have this behavior. I'll review this weekend.

src/stack/stack.jl
```diff
@@ -44,6 +46,7 @@ Base.last(s::AbstractDimStack) = s[last(keys(s))]
 # Only compare data and dim - metadata and refdims can be different
 Base.:(==)(s1::AbstractDimStack, s2::AbstractDimStack) =
     data(s1) == data(s2) && dims(s1) == dims(s2) && layerdims(s1) == layerdims(s2)
+Base.getproperty(s::AbstractDimStack, x::Symbol) = s[x]
```
sethaxen, line 49: I think you also need a `propertynames` overload.

sethaxen (commented):
LGTM, I only noticed one missing overload.

Issue #438, opened 2023-01-13
A 0-dimensional view of a `DimArray` has non-zero dimensions


sethaxen:
```julia
julia> using DimensionalData

julia> da = DimArray(randn(10, 4, 2), (:draw, :chain, :var));

julia> view(da, 1)
0-dimensional DimArray{Float64,0} with dimensions: Dim{:chain}, Dim{:var}
and reference dimensions: Dim{:draw}

1.26981
```
I would expect `:chain` and `:var` to also be reference dimensions.

rafaqz:
Ok something is broken in `reducedims I guess, we need to generated all the trailing 1s

Issue #437, opened 2023-01-13
Constructing a similar AbstractDimArray for specific dimensions


sethaxen:
I'm writing a generic method that takes a 3d `x::AbstractArray{<:Any,3}` and returns a `y::AbstractVector{T}`, where `axes(x, 3) === axes(y, 1)`. It's effectively a reduction but can't be computed with the existing reduction machinery.

I'd like to allocate the output in a way that preserves any named dimensions but without depending on DimensionalData. Currently the best I can do is
```julia
julia> x = DimArray(randn(10, 4, 2), (Dim{:draw}(1:10), Dim{:chain}(1:4), Dim{:var}([:a, :b])));

julia> T = Bool;

julia> y = similar(@view(x[begin, begin, :]), T)
2-element DimArray{Bool,1} with dimensions: 
  Dim{:var} Categorical{Symbol} Symbol[a, b] ForwardOrdered
and reference dimensions: 
  Dim{:draw} Sampled{Int64} 1:1 ForwardOrdered Regular Points,
  Dim{:chain} Sampled{Int64} 1:1 ForwardOrdered Regular Points
 :a  0
 :b  0
```

@devmotion suggested using the `similar(A, axes)` syntax documented in [Base](https://docs.julialang.org/en/v1/devdocs/offset-arrays/), but this is what we get:

```julia
julia> similar(x, (axes(x, 3),))
2-element Vector{Float64}:
 6.9364632819056e-310
 6.9364632819064e-310
```
It's not surprising this is a `Vector` since `axes` returns a tuple of `Base.OneTo`, and we don't even know which dimension these indices should correspond to. We could slice the first two axes, but this doesn't even work in Base:
```julia
julia> similar(parent(x), Bool, (axes(x, 1)[begin:begin], axes(x, 2)[begin:begin], axes(x, 3)))
ERROR: MethodError: no method matching similar(::Array{Float64, 3}, ::Type{Bool}, ::Tuple{UnitRange{Int64}, UnitRange{Int64}, Base.OneTo{Int64}})

Closest candidates are:
  similar(::AbstractDimStack, ::Any...)
   @ DimensionalData ~/.julia/packages/DimensionalData/KC0aG/src/stack/stack.jl:58
  similar(::AbstractArray, ::Type{T}, ::Union{Integer, AbstractUnitRange}...) where T
   @ Base abstractarray.jl:838
  similar(::Array, ::Type, ::Tuple{Vararg{Int64, N}}) where N
[...]

PR #439, opened 2023-01-14
Add DimUnitRange

sethaxen:
As suggested in #437, this PR adds a `DimUnitRange` type. Now all calls to `axes` return an object of this type. This allows `similar` to take axes of a dimensional array and still produce a dimensional array type. It also allows `DimArray`s to be constructed from other array types.

Here's an example:
```julia
julia> using DimensionalData

julia> da = DimArray(randn(10, 4, 2), (:draw, :chain, :var));

julia> axes(da)  # this could be prettier
(DimensionalData.Dimensions.DimUnitRange(range=Base.OneTo(10), dim=Dim{:draw, DimensionalData.Dimensions.LookupArrays.NoLookup{Base.OneTo{Int64}}}([1, 2, 3, 4, 5, 6, 7, 8, 9, 10])), DimensionalData.Dimensions.DimUnitRange(range=Base.OneTo(4), dim=Dim{:chain, DimensionalData.Dimensions.LookupArrays.NoLookup{Base.OneTo{Int64}}}([1, 2, 3, 4])), DimensionalData.Dimensions.DimUnitRange(range=Base.OneTo(2), dim=Dim{:var, DimensionalData.Dimensions.LookupArrays.NoLookup{Base.OneTo{Int64}}}([1, 2])))

julia> similar(da, Int, axes(da))
10×4×2 DimArray{Int64,3} with dimensions: Dim{:draw}, Dim{:chain}, Dim{:var}
[:, :, 1]
               3  140601401134352               10  140601401134992
               3  140601401134416               11  140601401135056
 140601401134096                7  140601401134736               15
 140601401134160                8  140601401134800               17
               4  140601401134480               12  140601401135120
               5  140601401134544               12  140601401135184
 140601401134224                9  140601401134864               18
 140601401134288                9  140601401134928               26
               6  140601401134608               13  140601401135248
               6  140601401134672               14  140601401135312
[and 1 more slices...]

julia> similar(trues(2, 2), axes(da))
10×4×2 DimArray{Bool,3} with dimensions: Dim{:draw}, Dim{:chain}, Dim{:var}
[:, :, 1]
 1  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0
[and 1 more slices...]

julia> similar(BitArray, (axes(da, 1),))
10-element DimArray{Bool,1} with dimensions: Dim{:draw}
  1  0
  2  0
  3  0
  4  0
  5  0
  6  0
  7  0
  8  0
  9  0
 10  0

julia> ax = axes(da, 1);

julia> similar(ax)
10-element DimArray{Int64,1} with dimensions: Dim{:draw}
  1  1
  2  2
  3  1
  4  1
  5  1
  6  1
  7  1
  8  1
  9  0
 10  1

julia> map(sin, ax)
10-element DimArray{Float64,1} with dimensions: Dim{:draw}
  1   0.841471
  2   0.909297
  3   0.14112
  4  -0.756802
  5  -0.958924
  6  -0.279415
  7   0.656987
  8   0.989358
  9   0.412118
 10  -0.544021
```

There are still a handful of methods I need to write tests for.

Currently the test suite fails due to #438


sethaxen:
All methods except `print` now have tests. Perhaps you have suggestions on how to have a prettier print method? The default for `AbstractUnitRange` would look like `1:1:10`, but it would be nice to 1) show as the range the underlying range object (e.g. `Base.OneTo`) and 2) also show the dim, though it seems unnecessarily verbose to show the dim indices.

rafaqz:
If you are showing it with keyword args we need a keyword constructor so it's runnable code

sethaxen:
> If you are showing it with keyword args we need a keyword constructor so it's runnable code

Good point. A keyword construct would be unnecessary, so I've removed keywords from the show method.

src/Dimensions/dimunitrange.jl
```diff
@@ -0,0 +1,48 @@
...
+
+# the below are adapted from OffsetArrays
+# https://github.com/JuliaArrays/OffsetArrays.jl/blob/master/src/axes.jl
+
+for f in [:length, :isempty, :first, :last]
+    @eval @inline Base.$f(r::DimUnitRange) = Base.$f(parent(r))
+end
+@inline Base.axes(r::DimUnitRange) = (r,)
+@inline Base.axes1(r::DimUnitRange) = r
+if VERSION < v"1.8.2"
```
rafaqz, line 29: Why the version check here? maybe add a comment

src/array/array.jl
```diff
@@ -127,6 +127,23 @@ Base.similar(A::AbstractDimArray, ::Type{T}, i::Integer, I::Vararg{<:Integer}) w
...
+# when the axes are DimUnitRanges we can return an `AbstractDimArray`
+function Base.similar(A::AbstractDimArray, T::Type, shape::Tuple{Dimensions.DimUnitRange,Vararg{Dimensions.DimUnitRange}})
+    data = similar(parent(A), T, map(parent, shape))
+    return rebuild(A; data=data, dims=deepcopy(dims(shape)), refdims=(), metadata=NoMetadata())
+end
+function Base.similar(
+    A::AbstractArray, T::Type, shape::Tuple{Dimensions.DimUnitRange,Vararg{Dimensions.DimUnitRange}}
+)
+    data = similar(parent(A), T, map(parent, shape))
+    return DimArray(data, deepcopy(dims(shape)))
```
rafaqz, line 139:
```suggestion
    C = dimconstructor(dims(shape))
    return C(data, deepcopy(dims(shape)))
```

We also have a trick for getting back other kinds of AbstractDimArray, as used in `rand` etc. It uses the maybe poorly named `dimconstructor` method:

https://github.com/rafaqz/DimensionalData.jl/blob/main/src/array/array.jl#L477-L481

It will return `DimArray` by default but allows packages to override that for `Dimension` or `LookupArray` that they define. So Rasters.jl will make a `Raster` here if any `LookupArray` is `Rasters.AbstractProjected`

https://github.com/rafaqz/Rasters.jl/blob/main/src/array.jl#L314

rafaqz (changes requested):
Very nice PR, thanks for the through test suite.

src/array/array.jl
```diff
@@ -127,6 +127,23 @@ Base.similar(A::AbstractDimArray, ::Type{T}, i::Integer, I::Vararg{<:Integer}) w
...
+end
+function Base.similar(
+    ::Type{T}, shape::Tuple{Dimensions.DimUnitRange,Vararg{Dimensions.DimUnitRange}}
+) where {T<:AbstractArray}
+    data = similar(T, map(parent, shape))
+    return DimArray(data, deepcopy(dims(shape)))
```
sethaxen, line 145:
```suggestion
    C = dimconstructor(dims(shape))
    return C(data, deepcopy(dims(shape)))
```

src/Dimensions/dimunitrange.jl
sethaxen, line 29:
```suggestion
if VERSION < v"1.8.2"
    # On recent Julia versions, these don't need to be defined, and defining them may
    # increase validations, see https://github.com/JuliaArrays/OffsetArrays.jl/pull/311
    Base.axes1(r::DimUnitRange) = r
```

sethaxen: It was due to it being unnecessary on recent Julia versions and increasing invalidations. I added a comment.

rafaqz (approved):

sethaxen:
@rafaqz do you mind if I tag a patch release after merging this PR?

rafaqz:
Go for it. I sent you a collaborator invite.

PR #440, opened 2023-01-14
fix dims for `view` with linear indexing

### LINKED ISSUE #438: A 0-dimensional view of a `DimArray` has non-zero dimensions

rafaqz:
Fixes #438 

@sethaxen this should fix your issue. A review would be good to make sure this covers everything this time.

### DIFF
--- a/src/array/indexing.jl
+++ b/src/array/indexing.jl
@@ -14,9 +14,14 @@ const SelectorOrStandard = Union{SelectorOrInterval,StandardIndices}
 for f in (:getindex, :view, :dotview)
     @eval begin
         if Base.$f === Base.view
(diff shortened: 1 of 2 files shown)


sethaxen (approved):
LGTM! With these changes, the issue in #438 is fixed and the tests in #439 pass

PR #441, opened 2023-01-14
Increment patch number

sethaxen:


rafaqz:
You can just merge these directly without PR I you like. Although I do like to put the version in the commit message!

sethaxen:
> You can just merge these directly without PR if you like. Although I do like to put the version in the commit message!

When I tried it said that main is a protected branch and can't be committed to directly.

rafaqz:
Ahh right. It would be good to be able to whitelist files from that check

rafaqz (approved):

PR #443, opened 2023-01-25
check dim length in inner constructor

rafaqz:
We need a check besides `format` for dim length that catches errors in `rebuild`. This should be a compile-time check.

Slightly annoying is that an inner constuctor like this needs to be added for all types that extend `AbstractDimArray`. It would be best if we could embed this in the abstract type so `N` has to match the length of the dims tuple... but that's only possible with `NTuple`, which isn't what we have.

### DIFF
--- a/src/array/array.jl
+++ b/src/array/array.jl
@@ -347,11 +347,12 @@ function DimArray(f::Function, dim::Dimension; name=Symbol(nameof(f), "(", name(
      DimArray(f.(val(dim)), (dim,); name)
 end
 
-checkdims(A::AbstractArray{<:Any,N}, dims::Tuple) where N =
-    length(dims) == N || _dimlengtherror(N, length(dims))
+checkdims(A::AbstractArray{<:Any,N}, dims::Tuple) where N = checkdims(N, dims)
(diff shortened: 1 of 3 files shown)


src/array/array.jl
```diff
@@ -283,6 +289,12 @@ function DimArray(f::Function, dim::Dimension; name=Symbol(nameof(f), "(", name(
      DimArray(f.(val(dim)), (dim,); name)
 end
 
+checkdims(A::AbstractArray{<:Any,N}, dims::Tuple) where N =
```
sethaxen, line 292: Should this have a unit test?

sethaxen (commented):
_Could_ the check be placed in a `rebuild` method itself so that it would work for all `AbstractDimArray`?

src/array/array.jl
rafaqz: Yeah probably, its just a test of `rebuild` currently.

rafaqz:
That's a good idea. The problem is `rebuild` methods are all defined for each type as well, so there is nowhere to put the check where you can guarantee it happens.

PR #446, opened 2023-02-12
dont deepcopy on similar

rafaqz:
I had this for safety in case lookup vectors are written to, but that's a pretty rare thing in practice.

And `deepcopy` is an expensive operations to couple to every use of `similar`.


sethaxen (approved):
To me this makes sense. I would expect `similar` to be used to allocate a suitable container with the same dimensions, almost irrespective of whether the array has named dimensions.  It seems wrong to then modify the dimension names; a user should construct new dimensions then.

PR #458, opened 2023-02-18
Update to ArrayInterface v7

ChrisRackauckas:
It's now setup with v1.9's package extensions features, so core is discontinued and everything works off of just ArrayInterface.


rafaqz:
implemented elsewhere

PR #463, opened 2023-02-19
Build docs on latest Julia release

sethaxen:
Currently the docs are being built on the LTS. This PR changes to build the docs for the latest Julia release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
         with:
-          version: '1.6'
+          version: '1'
       - run: |
           julia --project=docs -e '
             using Pkg
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,11 +36,15 @@ jobs:
   docs:
     name: Documentation
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        version: ['1.6', '1']
     steps:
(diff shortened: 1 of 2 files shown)


rafaqz:
Should we do both?

sethaxen:
Is that possible? I've never seen it. I don't think it's necessary though. It would be better to add admonitions indicating when a method is only available for certain Julia versions or when behavior changes between versions.

rafaqz:
You can just add another line with `- 1` instead of replacing `- 1.6`, and it will run both.

Its good to test 1.6 for use cases where people are stuck with LTS. They do exist. And we should check for LTS compatible syntax. When `NamedTuple` destructing came I broke the previous LTS in some of my packages by mistake, because the new syntax is nice and you want to used it.

sethaxen:
Ah, I thought you were asking whether we should _deploy_ docs on both LTS and latest release. Sure, we can build docs on both. I'll just need to add a way to disable deployment for v1.6, because otherwise one will overwrite the other.

rafaqz:
Ahh yeah that makes sense, I also didn't totally think through what I was saying with deployment of both docs. Actually having two versions will break the docstests because the `show` output changes.

But building on 1.6 and leaving it there was a conscious choice to force the docs to work on LTS - which means they also work on 1.9. The other way around means they will work on 1.9 but may not work on 1.6.

I chose the conservative option where the docs are less likely to be wrong for any users. But the tradeoff to use the latest syntax may also be valid.

rafaqz:
Ok I'm sold after trying to build docs in the other PR...

rafaqz (approved):

PR #468, opened 2023-02-24
Improve performance of At for Regular Ordered lookup array

metab0t:
Improve performance of At for Regular Ordered lookup array by using step information directly.

```julia
julia> using BenchmarkTools, DimensionalData

julia> A = DimArray([1,2,3,4,5], X(3:7))
5-element DimArray{Int64,1} with dimensions:
  X Sampled{Int64} 3:7 ForwardOrdered Regular Points
 3  1
 4  2
 5  3
 6  4
 7  5

# master
julia> @btime $A[X=At(7)]
  4.500 ns (0 allocations: 0 bytes)
5

# this PR
julia> @btime $A[X=At(7)]          
  3.100 ns (0 allocations: 0 bytes)
5

# Reference
julia> @btime $A[X=5]
  2.600 ns (0 allocations: 0 bytes)
5                                  
```

### DIFF
--- a/src/LookupArrays/selector.jl
+++ b/src/LookupArrays/selector.jl
@@ -126,7 +126,7 @@ function at(
     x = unwrap(selval)
     i = searchsortedlast(lookup, x)
     # Try the current index
-    if i == firstindex(lookup)
+    if i == firstindex(lookup) - 1
         i1 = i + 1
         if checkbounds(Bool, lookup, i1) && _is_at(x, lookup[i1], atol)
             return i1
--- a/src/LookupArrays/selector.jl
+++ b/src/LookupArrays/selector.jl
@@ -112,7 +112,7 @@ function at(
     x = unwrap(selval)
     Δ = step(span)
     i, remainder = divrem(x - first(lookup), Δ)
-    i += 1
+    i += firstindex(lookup)
     if remainder == 0 && checkbounds(Bool, lookup, i)
         return i
     else
@@ -126,7 +126,7 @@ function at(
     x = unwrap(selval)
     i = searchsortedlast(lookup, x)
     # Try the current index
-    if i == 0
+    if i == firstindex(lookup)
         i1 = i + 1
         if checkbounds(Bool, lookup, i1) && _is_at(x, lookup[i1], atol)
             return i1
--- a/src/LookupArrays/selector.jl
+++ b/src/LookupArrays/selector.jl
@@ -106,7 +106,7 @@ function at(lookup::LookupArray, sel::At; kw...)
     at(order(lookup), span(lookup), lookup, val(sel), atol(sel), rtol(sel); kw...)
 end
 function at(
-    ::Ordered, span::Regular, lookup::LookupArray{<:Union{Integer,Dates.TimeType}}, selval, atol::Nothing, rtol::Nothing;
+    ::Ordered, span::Regular, lookup::LookupArray{<:Integer}, selval, atol::Nothing, rtol::Nothing;
     err=_True()
 )
     x = unwrap(selval)
--- a/src/LookupArrays/selector.jl
+++ b/src/LookupArrays/selector.jl
@@ -103,16 +103,30 @@ function at(lookup::NoLookup, sel::At; kw...)
     return r
 end
 function at(lookup::LookupArray, sel::At; kw...)
-    at(order(lookup), lookup, val(sel), atol(sel), rtol(sel); kw...)
+    at(order(lookup), span(lookup), lookup, val(sel), atol(sel), rtol(sel); kw...)
 end
 function at(
-    ::Ordered, lookup::LookupArray{<:Union{Number,Dates.TimeType}}, selval, atol, rtol::Nothing;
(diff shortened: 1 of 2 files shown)


rafaqz:
Yeah I think `rtol` is not even used currently. But it probably should be?

But this makes me wonder _why_ this is slower. I assumed this just compiled away. What is actually taking that time by having one more `nothing` field.

metab0t:
I think the acceleration comes from new specialization of `at` for regular ordered lookup array.

rafaqz:
Oh right. So this PR is sort of mixing concerns...

I did mean to use `rtol` at some stage, why not implement it rather than remove it?

rafaqz:
This PR wont get merged with two ideas mixed like this. 

Can you separate them out into two PRs?

metab0t:
I will open a new PR without discarding rtol next week when I have more bandwidth.

metab0t:
@rafaqz Ready for review

src/LookupArrays/selector.jl
```diff
@@ -103,16 +103,30 @@ function at(lookup::NoLookup, sel::At; kw...)
...
-    ::Ordered, lookup::LookupArray{<:Union{Number,Dates.TimeType}}, selval, atol, rtol::Nothing;
+    ::Ordered, span::Regular, lookup::LookupArray{<:Integer}, selval, atol::Nothing, rtol::Nothing;
+    err=_True()
+)
+    x = unwrap(selval)
+    Δ = step(span)
+    i, remainder = divrem(x - first(lookup), Δ)
+    i += 1
+    if remainder == 0 && checkbounds(Bool, lookup, i)
+        return i
```
rafaqz, line 117: Is this still correct for an `OffsetArray`? It would be nice to test for this in the ecosystem.jl tests.

I imagine you will need `i += firstindex(lookup)` rather than `i += 1`, which seems to imbed a `fistindex(lookup) == 1` assumption, unless I misunderstand something.

rafaqz (changes requested):
We will just need to test this for OffsetArrays compatibility

metab0t:
> We will just need to test this for OffsetArrays compatibility

It fixes a bug in existing code 😆 because OffsetArrays axis are regarded as irregular. Why?

```
julia> oda
3×4 DimArray{Int64,2} with dimensions:
  X Sampled{Int64} Int64[100, 200, 300] ForwardOrdered Irregular Points,
  Y Categorical{Symbol} Symbol[a, b, c, d] ForwardOrdered
       :a   :b   :c   :d
 100  1    2    3    4
 200  3    4    5    6
 300  4    5    6    7
```

src/LookupArrays/selector.jl
```diff
@@ -103,16 +103,30 @@ function at(lookup::NoLookup, sel::At; kw...)
...
+end
+function at(
+    ::Ordered, ::Span, lookup::LookupArray{<:Union{Number,Dates.TimeType}}, selval, atol, rtol::Nothing;
     err=_True()
 )
     x = unwrap(selval)
     i = searchsortedlast(lookup, x)
     # Try the current index
-    if i === 0
+    if i == firstindex(lookup)
```
metab0t, line 129: It fixes OffsetArray bug but causes other errors. Not sure about the reason.

rafaqz: Thats not the change I meant

src/LookupArrays/selector.jl
```diff
@@ -103,16 +103,30 @@ function at(lookup::NoLookup, sel::At; kw...)
...
         i1 = i + 1
```
rafaqz, line 130:
```suggestion
        i1 = i + firstindex(lookup)
```

This is probably what you need to do

src/LookupArrays/selector.jl
metab0t: I have fixed it. It should be `if i == firstindex(lookup) - 1`

rafaqz:
> because OffsetArrays axis are regarded as irregular. Why?

It's not OffsetArrays, but any `AbstractArray` that is not an `AbstractRange`. Because we would have to calculate the step size. In some cases we don't know how to do that correctly (like with `DateTime`), so we just generally make the user set it if they have to with `span=Regular(step)`. Ranges we _know_ are regular just from the type, and we have the step in `step(range)` so we just go ahead and set it.

rafaqz:
We need a block like this to test `Regular` sampling OffsetArray with your changes. It wasn't actually tested before so no fault of yours. But I'm more onto spotting this now after all the OffsetArray dramas that happened in the ecosystem lately.

```julia
@testset "And when sampling is Regular" begin
    odimz_r = (X(OffsetArray(100:100:300, -1:1); span=Regular(100)), Y(OffsetArray([1.0, 2.0, 3.0, 4.0], 5:8); span=Regular(1)))
    oda_r = DimArray(oa, odimz_r)
    @test oda_r[At(100), At(1.0)] == oa[-1, 5] == 1
    @test oda_r[At(200), At(3.0)] == oa[0, 7] == 5
    @test oda_r[Between(100, 250), At(1)] == oa[-1:0, 5] == [1, 3]
end
```

And this at the top of the file 
```julia
using DimensionalData.LookupArrays
```

rafaqz (approved):
Thanks!

PR #469, opened 2023-02-28
Remove deprecated Vararg usage

sethaxen:
On main, if we load the package using `julia --depwarn=yes`, we get 50 deprecation warnings reading:

```julia
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
```

Running with `--depwarn=error`, we can pinpoint that the problem seems to be the syntax `Vararg{<:T}`. It seems instead one should use `Vararg{T}`. Similarly, `Tuple{<:T,Vararg{<:T}}` is apparently unnecessary. e.g.

```julia
julia> (1, 2) isa Tuple{Integer,Vararg{Integer}}
true

julia> (Int32(1), false) isa Tuple{Integer,Vararg{Integer}}
true
```

This PR removes all problematic lines, so that no more of these deprecation warnings are made.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DimensionalData"
 uuid = "0703355e-b756-11e9-17c0-8b28908087d0"
 authors = ["Rafael Schouten <rafaelschouten@gmail.com>"]
-version = "0.24.4"
+version = "0.24.5"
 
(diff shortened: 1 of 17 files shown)


rafaqz:
Thanks

PR #471, opened 2023-03-04
reduce compile time of stack construction and map

rafaqz:
This PR reduces compile time of `DimStack` and functions relying on `map` by avoiding intermediate `NamedTuple`s of `DimArray`. For whatever reason these are much slower to compile than `Tuple`s of `DimArray`.

There is also some work to use `Vector` rather than `Tuple` for some methods, so that compilation only has to happen once and `map` is cheaper to compile. 

Also some `@nospecialise` to reduce compilation (it helps even when the method is not called directly, oddly enough)

@sethaxen this has a lot less recompilation for stacks with different numbers of layers - for `map` and `DimStack`. It would be good to know if it helps with your specific workloads. 

Now that precompiling "sticks" a bit, we could add some precompile statements to remove even the first compilation for some methods.


rafaqz:
This ended up being quite rambling. Compile time of `show` was a big factor in REPL performance, so I fixed that too to some extent.

There's a proper SnoopPrecompile.jl block now that seems to help quite a bit with show.

More work could be done with SnoopCompile getting SnoopPrecompile to precompile more common use cases.

Other than that, simplification of the `DimStack` type could be done 
- move keys to a K parameter 
- use Tuple for data, layermetadata and layerdims

`combinedims` and `rebuild_from_arrays` could be further optimised for compilation time, they're currently the worst offenders with most operations on a stack that return another stack passing through them.

sethaxen:
@rafaqz I'm traveling right now but will try to review this weekend.

rafaqz:
No rush

rafaqz:
The idea with the `show` optimisation is that in practice it was making everything else feel slower -often show was taking longer than the methods that led to it. So:
1. precompile the show methods up front for some basic types (its half base methods that are not precompiled like `printstyled`)
2. reduce compile time for the inevitable custom dim and array type combinations to something like 0.2-0.3 seconds.

Getting it down to 0.1 seconds would be best, so you basically don't notice any lag, but its hard to get there.

sethaxen:
Indeed, I'm seeing `show` be much snappier. In particular the first `display(::DimArray)` for typical arrays used to take 0.52s but now takes 0.17s.  `display(::AbstractDimStack)` went from 1.27s to 0.96s; do you think this could be sped up further? 

With these changes we don't see much of a reduction in TTFX for I/O and common analyses. After performing the analysis once on one `AbstractDimStack`, the compile time for a different `AbstractDimStack` is substantially reduced, so precompilation could help us a lot here. But I wonder if the implementations are doing things that substantially increase compile time. How have you been going about identifying the compile time bottlenecks in DimensionalData?

rafaqz:
I've been using `@time` and SnoopCompile `@snoopi_deep` on the furst run of various methods and visualising the output with ProfileView.jl. Also Cthulhu `@descend` to understand what is being compiled. Its a little clunky as a workflow as you need to keep restarting julia, but ut works.

Youre right, we could precompile more things to make the first runs better too. Feel free to PR more, but we can keep working on this over time too. Its also a good excercise to understand the tradeoffs to guide future design decisions.

rafaqz:
I can post my ttfx benchmark script here when I get back to copenhagen, if you want something to get started.

rafaqz:
```julia
using SnoopCompile
using ProfileView
@time using DimensionalData
# trees = invalidation_trees(invalidations)
# using Cthulhu

# Define dimarrays
A = [1.0 2.0 3.0; 4.0 5.0 6.0]
x, y, z = X([:a, :b]), Y(10.0:10.0:30.0; metadata=Dict()), Z()
dimz = x, y
da1 = permutedims(DimArray(A, (x, y); name=:one));
da2 = DimArray(Float16.(2A), (x, y); name=:two)
da3 = DimArray(Int.(3A), (x, y); name=:three);
da4 = DimArray(cat(4A, 5A, 6A, 7A; dims=3), (x, y, z); name=:exteradim);

# Snoop display
tinf = @snoopi_deep @time display(da1)
tinf = @snoopi_deep @time display(da2)
tinf = @snoopi_deep @time display(da3)
tinf = @snoopi_deep @time display(da4)

# View flamegraph of some snoop
fg = flamegraph(tinf)
ProfileView.view(fg)

# Some mixed type DimArrays to generate random stacks from
das = (da1, da2, da3, da4)
# Define random stacks
many_das1 = Tuple(rand(das, 14)); @time s1 = DimStack(many_das1) 
many_das2 = Tuple(rand(das, 14)); @time s2 = DimStack(many_das2) 

# Merging stacks
tinf = @snoopi_deep @time merge(s1, s2)

s = s1
tinf = @snoopi_deep @time display(s)

# Some functions over stacks
tinf = @time @snoopi_deep map(sum, s)
tinf = @snoopi_deep @time map(A -> A[X=1], s)
tinf = @snoopi_deep @time prod(s)

fg = flamegraph(tinf)
ProfileView.view(fg)

```

PR #476, opened 2023-04-02
Make PermutedDimsArray type-inferrable

sethaxen:
`PermutedDimsArray` is currently not type-inferrable, likely because `PermutedDimsArray(parent(da), perm)` ends up not being inlined, and since its output type contains `perm` and `invperm(perm)`, this type is not inferrable.

This PR reimplements `PermutedDimsArray` to directly construct the type of the the array, so that when the types of the dimensions are known, then so is the output.

### DIFF
--- a/src/array/methods.jl
+++ b/src/array/methods.jl
@@ -201,12 +201,15 @@ for (pkg, fname) in [(:Base, :permutedims), (:Base, :adjoint),
             rebuild(A, $pkg.$fname(parent(A)), (AnonDim(Base.OneTo(1)), dims(A)...))
     end
 end
-for fname in [:permutedims, :PermutedDimsArray]
-    @eval begin
-        @inline function Base.$fname(A::AbstractDimArray, perm)
-            rebuild(A, $fname(parent(A), dimnum(A, Tuple(perm))), sortdims(dims(A), Tuple(perm)))
-        end
-    end
+@inline function Base.permutedims(A::AbstractDimArray, perm)
+    rebuild(A, permutedims(parent(A), dimnum(A, Tuple(perm))), sortdims(dims(A), Tuple(perm)))
+end
+@inline function Base.PermutedDimsArray(A::AbstractDimArray{T,N}, perm) where {T,N}
+    perm_inds = dimnum(A, Tuple(perm))
+    iperm_inds = invperm(perm_inds)
+    data = parent(A)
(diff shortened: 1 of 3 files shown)


src/array/methods.jl
```diff
@@ -201,12 +201,15 @@ for (pkg, fname) in [(:Base, :permutedims), (:Base, :adjoint),
...
-    end
+@inline function Base.permutedims(A::AbstractDimArray, perm)
+    rebuild(A, permutedims(parent(A), dimnum(A, Tuple(perm))), sortdims(dims(A), Tuple(perm)))
+end
+@inline function Base.PermutedDimsArray(A::AbstractDimArray{T,N}, perm) where {T,N}
+    perm_inds = dimnum(A, Tuple(perm))
+    iperm_inds = invperm(perm_inds)
+    data = parent(A)
+    data_perm = PermutedDimsArray{T,N,perm_inds,iperm_inds,typeof(data)}(data)
+    rebuild(A, data_perm, sortdims(dims(A), Tuple(perm)))
```
sethaxen, line 212: Alternatively, we can just use the `@inline` syntax introduced in v1.8 to make this type-stable on more recent Julia versions:

```suggestion
for fname in [:permutedims, :PermutedDimsArray]
    @eval begin
        @inline function Base.$fname(A::AbstractDimArray, perm)
            @static if VERSION ≥ v"1.8"
                @inline rebuild(A, $fname(parent(A), dimnum(A, Tuple(perm))), sortdims(dims(A), Tuple(perm)))
            else
                rebuild(A, $fname(parent(A), dimnum(A, Tuple(perm))), sortdims(dims(A), perm))
            end
        end
    end
```

rafaqz: Your furst implementation  is fine with me but whatever you prefer.

rafaqz:
Thanks, I dont think I ever checked stability for tgese methods. It might be good to do this systematically one day.

src/array/methods.jl
sethaxen: I prefer the one currently in the PR, since it works for all Julia versions.

sethaxen:
> Thanks, I dont think I ever checked stability for these methods. It might be good to do this systematically one day.

Agreed! At the moment I'm just using a lot of the primitives in implementations, and whenever a primitive causes type-instability I open an issue.

sethaxen:
@rafaqz this will need an approval from you before merging.

PR #487, opened 2023-05-24
add JuMP extension, update package

longemen3000:
based on https://github.com/jump-dev/JuMP.jl/issues/3214#issuecomment-1431813447. that functionality will only be available from 1.9 onwards. also:
- updates `ArrayInterfaceCore` -> `ArrayInterface` v7 (#458 )
- moves from `SnoopPrecompile` to `PrecompileTools`

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,16 +5,16 @@ version = "0.24.12"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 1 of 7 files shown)


odow:
This should probably instead go as an extension of JuMP, not an extension of DimensionalData.

odow:
Can be closed in favor of https://github.com/jump-dev/JuMP.jl/pull/3413

PR #502, opened 2023-07-27
fix cat for vectors with AnonDim

### LINKED ISSUE #433: `cat` on Regular Points creates incorrectly typed indices
The type for the new indices is wrong; they're typed as regular points, despite no longer being regular:
```
julia> da
3×4 DimArray{Int64,2} with dimensions: 
  Y Sampled{Int64} 10:10:30 ForwardOrdered Regular Points,
  Ti Sampled{Int64} 1:4 ForwardOrdered Regular Points
     1  2  3  4
 10  1  2  3  4
 20  3  4  5  6
 30  5  6  7  8

julia> cat(da, da; dims=Y)
6×4 DimArray{Int64,2} with dimensions: 
  Y Sampled{Int64} Int64[10, 20, …, 20, 30] ForwardOrdered Regular Points,
  Ti Sampled{Int64} 1:4 ForwardOrdered Regular Points
     1  2  3  4
 10  1  2  3  4
 20  3  4  5  6
 30  5  6  7  8
 10  1  2  3  4
 20  3  4  5  6
 30  5  6  7  8
```
Two possibilities here, assuming we want to keep type stability:
1. Always assume `cat` on regular points generates irregular points, or
2. Throw an error when the regular points don't line up -- I think this makes more sense.

rafaqz:
`hcat` and `vcat` were not tested with DimVector and were not actually working. This PR fixes that. It also standardises the output of `cat` on missing dimensions to use `AnonDim` so we always return a `DimArray`.

We also now rigorously check that the shared dimensions actually match all values as well as dimension type, and that span types are not mixed or broken (closes #433)

@sethaxen you may want a look at this, I think this was partly your code

Note: also fixed problems in docs in primatives as I used the keywords here

### DIFF
--- a/src/array/show.jl
+++ b/src/array/show.jl
@@ -73,7 +73,7 @@ function print_name(io::IO, name)
     end
 end
 
-# Base.print_matrix(io::IO, A::AbstractDimArray) = _print_matrix(io, parent(A), lookup(A))
+Base.print_matrix(io::IO, A::AbstractDimArray) = _print_matrix(io, parent(A), lookup(A))
 # Labelled matrix printing is modified from AxisKeys.jl, thanks @mcabbot
 function _print_matrix(io::IO, A::AbstractArray{<:Any,1}, lookups::Tuple)
     h, w = displaysize(io)
--- a/src/array/show.jl
+++ b/src/array/show.jl
@@ -73,7 +73,7 @@ function print_name(io::IO, name)
     end
 end
 
-Base.print_matrix(io::IO, A::AbstractDimArray) = _print_matrix(io, parent(A), lookup(A))
+# Base.print_matrix(io::IO, A::AbstractDimArray) = _print_matrix(io, parent(A), lookup(A))
 # Labelled matrix printing is modified from AxisKeys.jl, thanks @mcabbot
 function _print_matrix(io::IO, A::AbstractArray{<:Any,1}, lookups::Tuple)
     h, w = displaysize(io)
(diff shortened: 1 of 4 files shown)


sethaxen:
I'll take a look at this later today.

src/array/show.jl
```diff
@@ -73,7 +73,7 @@ function print_name(io::IO, name)
     end
 end
 
-Base.print_matrix(io::IO, A::AbstractDimArray) = _print_matrix(io, parent(A), lookup(A))
+# Base.print_matrix(io::IO, A::AbstractDimArray) = _print_matrix(io, parent(A), lookup(A))
```
sethaxen, line 76: Why is this commented out instead of removed?

src/array/show.jl
```diff
@@ -88,6 +88,7 @@ function _print_matrix(io::IO, A::AbstractArray{<:Any,1}, lookups::Tuple)
     Base.print_matrix(io, A_dims)
     return nothing
 end
+_print_matrix(io::IO, A::AbstractDimArray, lookups::Tuple) = _print_matrix(io, parent(A), lookups)
```
sethaxen, line 91: What does this new method do?

sethaxen (commented):
I don't see any obvious problems, but I did notice that a `_print_matrix` method in `show.jl` is no longer being covered by the test suite (hence the decrease in coverage).

rafaqz:
The `show` related changes are because we were catting together DimArrays i  fragile ways that break now that cat has more checks.

I'll ckean up the coverage  thanks for catching that.

src/array/show.jl
rafaqz: accident, that fixed the coverage too

src/array/show.jl
rafaqz: This shows the matrix with the dimension values appended. It was breaking because the `cat` was out of alignment and we check that now

Issue #500, opened 2023-07-25
`cat` is not type-inferrable


sethaxen:
It seems that when using `cat`, Julia cannot infer the types of the resulting dimensions:

```julia
julia> using DimensionalData, Test

julia> foo(x, y) = cat(x, y; dims=Dim{:foo}([:c, :d]));

julia> da = DimArray(randn(3), Dim{:a})
3-element DimArray{Float64,1} with dimensions: Dim{:a}
 1   0.54291
 2  -0.0289045
 3   0.526266

julia> foo(da, da)
3×2 DimArray{Float64,2} with dimensions: 
  Dim{:a},
  Dim{:foo} Categorical{Symbol} Symbol[:c, :d] ForwardOrdered
   :c          :d
  0.54291     0.54291
 -0.0289045  -0.0289045
  0.526266    0.526266

julia> @inferred foo(da, da)
ERROR: return type DimArray{Float64, 2, Tuple{Dim{:a, DimensionalData.Dimensions.LookupArrays.NoLookup{Base.OneTo{Int64}}}, Dim{:foo, DimensionalData.Dimensions.LookupArrays.Categorical{Symbol, Vector{Symbol}, DimensionalData.Dimensions.LookupArrays.ForwardOrdered, DimensionalData.Dimensions.LookupArrays.NoMetadata}}}, Tuple{}, Matrix{Float64}, DimensionalData.NoName, DimensionalData.Dimensions.LookupArrays.NoMetadata} does not match inferred return type DimArray{Float64, 2, _A, Tuple{}, Matrix{Float64}, DimensionalData.NoName, DimensionalData.Dimensions.LookupArrays.NoMetadata} where _A<:Tuple
```

rafaqz:
Yeah, it doesnt yet know the order or if the lookup is ordered at all. 

You can specify the `Categorical` lookup manually and it should be type stable.

rafaqz:
Yes its fine if you specify `Categorical` with the `order` keyword:

```julia
using DimensionalData.LookupArrays
bar(x, y) = cat(x, y; dims=Dim{:foo}(Categorical([:c, :d]; order=ForwardOrdered())))

julia> @inferred bar(da, da)
3×2 DimArray{Float64,2} with dimensions: 
  Dim{:a},
  Dim{:foo} Categorical{Symbol} Symbol[:c, :d] ForwardOrdered
   :c         :d
  1.6825     1.6825
  0.216932   0.216932
[...]

PR #515, opened 2023-08-16
Breaking: fix cat, again

rafaqz:
Ok, properly this time.

Breaking change for a number of minor reasons. 

Constructed dimensions passed to `dims` will now replace the current dimensiion. Symbols or types will try to use the existing dimension values, or error if they don't make sense.

Passing constructed dimensions that are the wrong size will now break, where before it just used the existing dimensions. As with the last PR, buggy, incorrect lookups are no longer allowed, and will ~~error~~ result in a warning and `cat` on the `parent` array being returned.


sethaxen:
Hm, I think this solution is still problematic, because it is completely fine for a third-party function to `vcat`, `hcat`, or `cat` 2 arrays together, and that code won't care about our dimensions, but because we care about dimensions, we raise an error and forbid our users from applying such a function to these inputs. I think to work around this I would need to special-case `DimArray`s in a number of functions that currently work for all array types.

From that perspective, falling back to the parent array type and raising a warning instead of an error is more useful. I wonder if this type-instability is really a major issue, since that fallback should create a typeunion of size 2, which is usually okay for performance.

An alternative would be to take this approach `Int`(s) are provided for `dims`, since generic Julia code will only provide those. But this causes an inconsistency between `Symbol` and `Int` dims for the same dimensions, which is not ideal.

rafaqz:
Thanks, thats good feedback. I agree the warning and non-dimensional return value are probably the best outcome here.

Part of my reason to try for type stability was #500, but probably its fine compared to the cost of concatenation in most cases. And yes we should get a small union optimisation if its written properly.

rafaqz:
@sethaxen, another question I would like to bounce off you:

If the dimension lookup values don't make sense - say they contain duplicate values or have the wrong order - should we try to keep the dimension wrappers and drop to `NoLookup` ? Or just drop everything and return `cat` on the parent arrays?

PR #516, opened 2023-08-17
Example for the selection of intervals in DimArrays added to the course.

dpabon:


rafaqz:
Thanks!

The recommended way to do that for `Dim{:x}` dimensions is:
```julia
A2[distance=12 .. 35, time=Date(2001, 5) .. Date(2001, 7))]
```

dpabon:
Great! Thanks for the clarification!

I could not find that way in the documentation. Shall I close the PR and open a new one?

rafaqz:
You can also just edit this PR inline

rafaqz:
The comment above now doesnt match the example ;)

dpabon:
Ups… Thanks for the catch! Now should be correct.

docs/crash/course/course.jl
```diff
@@ -150,6 +150,9 @@ mean(A3; dims=Ti)
 
 A[X(12..35), Ti(Date(2001, 5)..Date(2001, 7))]
 
+# To select intervals in DimArrays (e.g. ```A2```) you need to specify ```dimname=a .. b```:
+
+A2[distance=12 .. 35, time=Date(2001, 5) .. Date(2001, 7))]
```
rafaqz, line 155:
```suggestion
A2[distance=12 .. 35, time=Date(2001, 5) .. Date(2001, 7)]
```

PR #521, opened 2023-08-22
add `Not` selector from InvertedIndices.jl

### LINKED ISSUE #365: Allow use of `Not` from InvertedIndices.jl
Being able to do `Not(-20..20)` or `Not(Contains(20.27))` would be pretty useful.

### LINKED ISSUE #519: Not included kwarg
Is there an equivalent to the DatFrames syntax `Not`, so that we can include all values in dimensions except those in the vector/element passed to `Not`?

rafaqz:
Closes #519 and closes #365


test/selector.jl
```diff
@@ -1121,6 +1121,15 @@ end
              3  6  57  60]
     end
 
+    @testset "Not " begin
+        dimz = X(1.0:2:10.0), Ti(1u"s":5u"s":11u"s")
+        A = DimArray((1:5) * (1:3)', dimz)
+        A1 = A[X=Not(Near(5.1)), Ti=Not(1u"s" .. 10u"s")]
```
lazarusA, line 1127:
```suggestion
        A1 = A[X=Not(Near(5.1)), Ti=Not(1u"s" .. 10u"s")]
        A2 = A[Ti=Not(At(1u"s"))]
        A3 = A[X=Not(At([1.0,3.0]))]
        @test lookup(A2, :Ti) == [6u"s", 11u"s"]
        @test lookup(A3, :X) == [5.0,7.0,9.0]
```

src/DimensionalData.jl
```diff
@@ -24,6 +24,7 @@ import Adapt,
        ArrayInterface,
        ConstructionBase, 
        Extents,
+       InvertedIndices,
        IteratorInterfaceExtensions,
        RecipesBase,
        PrecompileTools,
```
lazarusA, line 30: You need to export also `Not` in line 52. Otherwise is not available at the DD level.

lazarusA (commented):
I just added some test using At. And it works :D, please also export `Not` at the DD level. Other than that it looks ready.

PR #560, opened 2023-10-25
used binary search on string indices

bjarthur:
any reason not to use the much faster `searchsorted` instead of `findfirst` for `At` on a dimension whose indices are `String`?  tests pass locally and my application code is 2x faster.

### DIFF
--- a/src/LookupArrays/selector.jl
+++ b/src/LookupArrays/selector.jl
@@ -138,7 +138,7 @@ function at(
     end
 end
 function at(
-    ::Ordered, ::Span, lookup::LookupArray{<:Union{Number,Dates.TimeType}}, selval, atol, rtol::Nothing;
+    ::Ordered, ::Span, lookup::LookupArray{<:Union{Number,Dates.TimeType,AbstractString}}, selval, atol, rtol::Nothing;
     err=_True()
 )
     x = unwrap(selval)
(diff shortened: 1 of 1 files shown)


rafaqz:
No reason I cant remember what even failed to have a type restriction there.

I think `searchsorted ` just needs `<` and  `<=` defined? Can anything be considered  "ordered" without that?

Maybe we can just delete all of those types.

src/LookupArrays/selector.jl
```diff
@@ -138,7 +138,7 @@ function at(
     end
 end
 function at(
-    ::Ordered, ::Span, lookup::LookupArray{<:Union{Number,Dates.TimeType}}, selval, atol, rtol::Nothing;
+    ::Ordered, ::Span, lookup::LookupArray{<:Union{Number,Dates.TimeType,String}}, selval, atol, rtol::Nothing;
```
rafaqz, line 141:
```suggestion
    ::Ordered, ::Span, lookup::LookupArray, selval, atol, rtol::Nothing;
```

src/LookupArrays/selector.jl
```diff
@@ -163,7 +163,7 @@ function at(
         end
     end
 end
-# catch-all for an unordered or non-number index
+# catch-all for an unordered index
 function at(::Order, ::Span, lookup::LookupArray, selval, atol, rtol::Nothing; err=_True())
```
rafaqz, line 167:
```suggestion
function at(::Unordered, ::Span, lookup::LookupArray, selval, atol, rtol::Nothing; err=_True())
```

rafaqz:
Well that failed on `Symbol`... maybe we can go back to just adding `String`, or `AbstractString` and work this out more systematically later.

rafaqz (approved):

rafaqz:
Thanks!

PR #561, opened 2023-10-27
add a custom `SelectorError`exception

rafaqz:
This PR adds a custom `SelectorError` so we get a little more information than a `BoundsError` - it shows the selector used, its value, and the lookup bounds (for ordered lookups) or lookup values (for unordered and categorical lookups).

This is a start on adding more helpful custom errors to DD.


rafaqz:
Apparently changing error types is not a breaking change, so this will be a patch version.

PR #571, opened 2023-12-02
update Makie to 0.20

rafaqz:


Project.toml
```diff
@@ -48,7 +48,7 @@ IntervalSets = "0.5, 0.6, 0.7"
 InvertedIndices = "1"
 IteratorInterfaceExtensions = "1"
 LinearAlgebra = "1"
-Makie = "0.19"
```
sethaxen, line 51: Is it necessary to drop compat for the older version?

sethaxen: Couldn't you then do
```julia
const SurfaceLike = isdefined(Makie, :SurfaceLike) ? Makie.SurfaceLike : Union{Makie.VertexGrid,Makie.CellGrid,Makie.ImageLike}
```

My concern is that the ecosystem may lag a little in adding Makie v0.20 support because it made so many changes, so dropping support if not necessary might block users from installing new DimensionalData versions.

rafaqz: Ok lets do that.

PR #572, opened 2023-12-05
Reset version to 0.25.7

felixcremer:
For some reason there was no registration made with 0.25.7 and therefore all following registration attempts failed.
This should reallow the automerge of the Registrator bot. 
See https://github.com/JuliaRegistries/General/pull/96347 for the latest failing registration.


rafaqz:
Ah thanks

Issue #420, opened 2022-11-17
Interpolate


DanDeepPhase:
Any interest in supporting interpolation along the axes?

I was able to hack it in my code with something like:
```julia
using Interpolations, DimensionalData
import Interpolations.interpolate

interpolate(da::DimArray) = interpolate(Array.(dims(da)), da, Gridded(Linear())) 

xs = ys = 1:10
zs = [rand() for x in xs, y in ys]
M = DimArray(zs, (Y(ys),X(xs)))

itp = interpolate(M)
itp[3.3, 5.1]
[itp[y,x] for y in 3.1:0.35:7.2, x in 2:4] 

```
I saw you use something in rasters for re-gridding, which i think this approach is best suited for. some tweaks would be needed to get reverse ordered vectors working.

There may be a more intrusive / better way to do this by extending or creating a variant of `At()`

rafaqz:
Rasters users `gdalwarp` because it handles projections and has a lot of options. But native interpolation is better eventually.

Unfortunately Interpolations.jl is a very heavy dependency, snd this package is a dep of other packages so we have to keep things lean.

DimensionalDataInterpolations.jl could be useful.

rafaqz:
Revisiting this: with Julia 1.9 and this PR https://github.com/JuliaLang/julia/pull/47695 we can add a weak dependency on Interpolations.jl.

If you want to make a PR that extends Intpolations.jl methods to work on `AbstractDimArray` in a weak dependency in this package, that would be a very useful addition.

kapple19:
I'm happy to look into implementing `DataInterpolationsND.jl` as an extension.

rafaqz:
Yeah, that would be great honestly. 

The tricky part will be the interface, like how do we let users specify interpolation dimension types without having to define custom syntax for it all.

kapple19:
I might approach this some small PRs at a time, it's a bigger task than I anticipated, and I want more features the more I think about it 😅

Like, `DimStack` and `DimTree` support!

But will get the simple things through first, like dispatching correctly to `DataInterpolations.jl` (1D) or `DataInterpolationsND.jl`.

Is it acceptable to `parent` the array and dims in the call to the interpolation constructors? Or is there a better way?

Also last I checked, the interpolator calles accept vector inputs for vector outputs. We want that too?

rafaqz:
> Is it acceptable to parent the array and dims

[...]

PR #586, opened 2023-12-27
add Interpolations.jl extension

rafaqz:
This is an idea for adding an Interpolations.jl extension.

It turns out there isn't a clean way to extend Interpolations.jl methods like `interpolate` or `scale` because really on a `DimArray` you would always want both and have the information for both already in one object.

So the idea here (and what i personally need the most) is to define an interpolation method that lets you interpolate from one `AbstractDimArray` to another in a single command.

We have most of the information we need to choose the interpolation method - users can specify e.g. `Cubic` or `Linear` if they really want to. But `BSpline`/`Gridded` we know from `Regular` / `Irregular` etc.

 We may also want to provide an interpolator object that can interpolate points.
 
 I'll probably leave this here for a while until its clearer what the best strategy is, but comments are welcome.


rafaqz:
Closes #420 

@DanDeepPhase we can implement this pretty easily now. But if you have any comments on syntax or how this should work that would help.

asinghvi17:
Any particular reason this was closed?

rafaqz:
In favour of

https://github.com/rafaqz/DimensionalData.jl/pull/609

Which needs tests, and I forgot about. And I'm not really sure on the best syntax. Feedback appreciated.

asinghvi17:
Did #609 get merged here?  I don't see the log

rafaqz:
It got merged to the branch yes, but the PR seems dead there's no reopen button

asinghvi17:
I guess the PR had to be reopened for the log to refresh...

haakon-e:
Glad to find this PR. It seems the PR as-is returns an array upon interpolation

To add on to how cool this would be, here's a hacked up version that for `DimVector`s (but probably easy to generalize) would return a `DimVector` after interpolation:

```julia
function Interpolations.linear_interpolation(A::AbstractDimVector; kw...)
    function itp(pts...)
        itp = linear_interpolation(DimensionalData.index(dims(A)), A; kw...)
        valtuple = lookup(pts, dims(A))
        intp = itp(valtuple...)
        DimVector(intp, pts)
    end	
end
```

example:
```julia
da = cumsum(rand(X(0:0.1:1)))
itpp = linear_interpolation(da)
da_itp = itpp(X(0:0.01:1))

# plotting
f,a,s = scatter(da_itp, markersize=4, label = "new pts")
scatter!(a, da, label = "old pts")
axislegend()
f
```
image

rafaqz:
It's just a little tricky getting the syntax right, if you want to take over find shing the PR that would help it along

PR #588, opened 2024-01-09
honor :limit kwarg in IOContext

bjarthur:
support TerminalPager.  see https://github.com/ronisbr/TerminalPager.jl/issues/34

these changes work for my narrow use case.  would be good to test more broadly before merging.


rafaqz:
The errors seem unrelated, I'll have a look when I get time

bjarthur:
just force pushed some changes to fix a bug

PR #591, opened 2024-01-17
Add a `groupby` function

### LINKED ISSUE #3: SplitApplyCombine
It would be good to support SplitApplyCombine.jl or something similar, so that complex manipulations could be applied along dimensions in a structured way, say getting mean daily values from hourly values over some timespan.

rafaqz:
Add a `groupby` function (will do aggregate later).

@lazarusA 

@bkamins is there a way we can avoid clashes with DataFrames.jl ? could `groupby` go in DataAPI.jl?

Closes #3


test/groupby.jl
```diff
@@ -0,0 +1,11 @@
+using DimensionalData, Test, Dates
+using DimensionalData: Dimensions
+using DimensionalData: LookupArrays
+using Statistics
+const DD = DimensionalData
+
+# Make a demo DimArray
+A = rand(X(1:0.1:2), Y(1:20), Ti(DateTime(2000):Day(10):DateTime(2003)))
+# Group by month and even/odd Y axis values
+groupmeans = groupby(A, Ti=month, Y=isodd)
```
lazarusA, line 10: I think I should do some comparisons with other libraries, just a few. But, in the meantime the printing, specially here is really too much. 

Maybe for starters, don't print any numbers, and just focus on the output groups and dimensions. Print just that, each output group name and corresponding Dims.

rafaqz: Good idea would be good to have more examples.

Sorry its not really related to this PR its just how we print arrays of arrays 😭

We need to get them to squash like Base does

bkamins:
I am OK to put `groupby` in DataAPI.jl requiring that the first positional argument must have a type constraint to a type defined in a package that adds a method to `groupby`. @nalimilan - OK?

rafaqz:
Thanks, first argument owned by the package is perfect.

docs/crash/course/groupby.jl
```diff
@@ -0,0 +1,66 @@
...
+#
+mean.(groupby(ds, Ti=yearmonth))
+
+#
+mean.(groupby(ds, Ti=hour))
+
+#
+mean.(groupby(ds, Ti=Dates.hour12))
+
+# ### TODO. How do could we incorporate resample? Let's say if we have hour resolution I want to resample every 3,6,12.. hours?
```
rafaqz, line 52: Probably you would use `aggregate` in Rasters.jl for that? I have thought about moving that to DD.

But you could do something like :

```julia
mean.(groupby(ds, Ti=x -> hour(x) ÷ 3 ))
```

src/aggregate.jl
```diff
@@ -0,0 +1,320 @@
...
+
+const SCALE = """
+- `scale`: the aggregation factor, which can be an `Integer`, a tuple of `Integer`
+    for each dimension, or a `Dimension`
+"""
+
+const METHOD = """
+- `method`: a function such as `mean` or `sum` that can combine the
+    value of multiple cells to generate the aggregated cell, or a `Symbol` 
+    `:start`, `:center`, or `:end` or `Locus` like `Start()` `Center()` or `End`
```
nalimilan, line 23: Why not use `first` and `last` instead of `:start` and `:end`? Even `:center` could be replaced with a new function (even if the underlying implemention is more efficient).

rafaqz: That makes sense, we can just define `center`. 

I didnt make them functions because they wouldnt actually be used as functions of the windiw area like `sum`, and because with reversed order lookups they will actually give you the oppodite value to calling `first` on the window.

lazarusA:
what's the plan for aggregate? it will be a groupby operation as well right? 
Latest commit shows a mean over monthly aggregates (groupby month 😄 ).

rafaqz:
I think we will still have aggregate for doing things like aggregate 3 cells to one. It was just too much to get it all right and blocking merging this one, so I separated it out

rafaqz:
@meggart do you want to look at the groupby syntax/implementation and give it the ok for later DiskArrays.jl integration? Just the `groupby.jl` src/test/docs are worth a loook

src/groupby.jl
```diff
@@ -0,0 +1,341 @@
...
+    name::Na
+    metadata::Me
+    function DimGroupByArray(
+        data::A, dims::D, refdims::R, name::Na, metadata::Me
+    ) where {D<:Tuple,R<:Tuple,A<:AbstractArray{T,N},Na,Me} where {T,N}
+        checkdims(data, dims)
+        new{T,N,D,R,A,Na,Me}(data, dims, refdims, name, metadata)
+    end
+end
+function DimGroupByArray(data::AbstractArray, dims::Union{Tuple,NamedTuple};
```
meggart, line 27: This is great. I think with this DimGroupByArray we should have everything we need to make the DiskArrayEngine Wrapper

PR #605, opened 2024-02-01
add an Interfaces.jl extension, and some tests

### LINKED ISSUE #374: Feature request: a way test implementations of AbstractDimArray
Hi @rafaqz, and thanks for creating this package!

I’m working on a package that provides an implementation of an AbstractDimArray along with some additional fields: https://github.com/JuliaAstro/AstroImages.jl/pull/30 
I’d like a way to test that my implementation is correct.
My request is for a way to run this package’s tests, or a subset of them, on the downstream package to check the implementation.

If you think this is useful, I’m happy to lend a hand if you could suggest the right place to get started.

rafaqz:
@gdalle this is the use-case for InterfacesCore. 

I imagine a bunch of packages would want to be structured like this.

Closes #374


gdalle:
How does this package find InterfacesCore?

rafaqz:
Yeah it can't for now and I haven't fiddled with CI to get the dev version. We should get InterfacesCore registered soon

PR #609, opened 2024-02-02
linear interpolation

DanDeepPhase:
This implements a general linear interpolation. Summary:

1. updated module name
2. i went with the "create an interpolator" route, and started with linear_interpolation 
3. You had been implementing a regridding algorithm along some but not all dimensions. It wasn't working when i cloned, and i wasn't able to get it to work, as detailed below
4. its a start, and Interpolations.interpolate() is probably a better function to extend, which should be easy enough. It requires returning the values of the nodes to avoid recursion, which i couldn't remember how to do.

I commented out your existing code because it was failing for me in a few ways. I don't really understand extensions, so could be easy fixes.
1. Can you use extra dependencies (rasters) in the extension? 
2. Does the extension need a TOML at that point?was getting an explicit call for DimensionalData problem which i think was in the const DD =DimensionalData
3. when i try to run test code on the extension, the only functions that work for me are the ones that are extended. so i can run Interpolations.linear_interpolation() but not interp().


rafaqz:
Yeah that's how it works, you can only add to function defined somewhere else, you can "see" functions or types created inside the extension.

Returning an interpolator is probably a better strategy too.

ext/DimensionalDataInterpolations.jl
```diff
@@ -1,48 +1,62 @@
...
-                sitp = scale(itp, map(parent, lookup(shared_dims))...)
-                sitp(map(parent, lookup(dest_dims))...)
-            end
-        end
+#     return DimArray(data, dest_dims)
+# end
+
+function Interpolations.linear_interpolation(A::AbstractDimArray)
+
+    raw_dims = [dim.val.data for dim in dims(A)]
```
rafaqz, line 50:
```suggestion
    raw_dims = map(index, dims(A))
```

ext/DimensionalDataInterpolations.jl
```diff
@@ -1,48 +1,62 @@
...
+
+    isrange(vec) = typeof(vec) <: AbstractRange 
+
+    if all(isrange.(raw_dims))
```
rafaqz, line 54:
```suggestion
    if all(DD.isregular, raw_dims)
```

ext/DimensionalDataInterpolations.jl
```diff
@@ -1,48 +1,62 @@
...
+        itp = interpolate(A, BSpline(Linear()))
+        sitp = scale(itp, raw_dims...)
```
rafaqz, line 56:
```suggestion
        sitp = scale(itp, DD.index(dims(A))...)        
```

DanDeepPhase:
I trimmed it down a lot. The code is pretty straightforward now. Perhaps it should extend interpolate, but i tend to use the convenience functions more. 

For future effort, i could see two nice to have functions:
1. interp_slice(A,dim(val)) = create a slice of a DimArray at an off-index value. I think this is what your interp code was doing. I like the ability to drop "referenced dimensions", like A[:,:,1] 
2. regrid(A, dim(vector), dim2(vector2)) pass an array of coordinates, and interpolate to those values 

not super good with CI, seemed like a compat issue, but it didn't resolve when i added interpolations to the compat list.

rafaqz:
Sorry waa too busy with other things. The Project.toml file is broken, you need an equals sign for the Interpolations version here:

https://github.com/rafaqz/DimensionalData.jl/actions/runs/7786303160/job/21230819886#step:4:46

ext/DimensionalDataInterpolations.jl
```diff
@@ -1,48 +1,13 @@
...
-                sitp(map(parent, lookup(dest_dims))...)
-            end
-        end
-    else
-        error("interpolate is only implemented for regular grids")
-    end
-
-    return DimArray(data, dest_dims)
+function Interpolations.linear_interpolation(A::AbstractDimArray; varargs...)
+    linear_interpolation(DimensionalData.index(dims(A)), A; varargs...)
```
rafaqz, line 6:
```suggestion
function Interpolations.linear_interpolation(A::AbstractDimArray; kw...)
    linear_interpolation(DimensionalData.index(dims(A)), A; kw...)
```

ext/DimensionalDataInterpolations.jl
```diff
@@ -1,48 +1,13 @@
...
 end
 
+function Interpolations.cubic_spline_interpolation(A::AbstractDimArray; varargs...)
+    cubic_spline_interpolation(DimensionalData.index(dims(A)), A; varargs...)
```
rafaqz, line 10:
```suggestion
function Interpolations.cubic_spline_interpolation(A::AbstractDimArray; kw...)
    cubic_spline_interpolation(DimensionalData.index(dims(A)), A; kw...)
```

test/interpolations.jl
```diff
@@ -2,6 +2,22 @@ using DimensionalData
...
 
 f((x1, x2)) = log(x1+x2)
-A = f.(DimPoints((X(1:.1:10), Y(1:.5:20))))
+a = f.(DimPoints((X(1:1:10), Y(1:1.5:20))))
+b = f.(DimPoints((X(1:1:10), Y([1.0,3,7.5,10,15]))))
 to = rand(X(2:.3:7), Y(2:.3:17))
-out = DimensionalData.interp(A; to)
+# out = DimensionalData.interp(A; to)
+
+itp_a = linear_interpolation(a)
```
rafaqz, line 10: We should test that we get the right answer too, with `@test`.  You can just define the ranges `2:0.3:7` etc as separate variables and use them both in defining the `DimArray` and in regular `linear_interpolation`

You'll need `using Test` at the top.

rafaqz (commented):
Looks good, we just need some tests to lock down that it does the right thing.

asinghvi17:
I wonder if this could also use the DimensionalData Makie utility functions...it seems like one would want to shift indices to sample at the center and similar things when interpolating by default.

I do wish that Interpolations.jl respected the array axis interface, but I guess that would be a pretty annoying thing to do.  That being said, for full compatibility we should also be overriding like:
```julia
Interpolations.interpolate(A::AbstractDimArray, interpmode; kw...) = Interpolations.interpolate(DimensionalData.index(dims(A)), A; kw...)
```
or something, to provide general support.

rafaqz (approved):

rafaqz:
Just realised this PR was against the other PR, so I'll reopen it

Beforerr:
Could this be merged into main?

rafaqz:
Do you want to finish the tests?

Beforerr:
Oh I do not realize the test is not finished yet. Maybe I would help in the weekend if I get some time.

rafaqz:
See https://github.com/rafaqz/DimensionalData.jl/pull/609#discussion_r1508792363

rafaqz:
It's also merged into the interpolations branch.

I'm not sure about the syntax either, or if there's a better way to integrate with DD syntax, and I don't have time to look at it

PR #615, opened 2024-02-06
Breaking: better dimensional indexing

rafaqz:
Major overhaul of indexing to remove bugs and better use all the tools we already have.

- `AbstractBasicDimArray` was added, and `AbstractDimIndices` and other objects like `DimSlices` are now in it along with all `AbstractDimArray`. This should give more consistent behavior.
- Tuple{<:Dimension} now works like `CartesianIndex` so `DimIndices` can work like `CartesianIndices`
- `DimIndices` can be used to index like `CartesianIndices` but of course it will work accross any dimension permutions
- `DimKeys` are now called `DimSelectors` (both still exported) because its an array of `<:Dimension{<:Selector}`
- `DimSelectors` works just like `DimIndices`. This means you can use it to interpolate, e.g. with a `Near` selector.
- `AbstractArray{<:Dimension}` and `AbstractArray{<:DimTuple}` work like `AbstractArray{<:CartesianIndex}` but again work on any permutation. They use `mergedims` so you still get a `DimArray` back.
- `mergedims` now accepts a tuple, and ust combines the dim names into a new dim name, if you want that. Indexing uses this to name dimensions after indexing with an AbstractArray.
- `eachslice` on `AbstractDimStack` no uses the `DimSlices` object rather than a Generator. The `DimSlices` object is useful for `groupby` as well, so we may as well have for `eachslice` too.

Breaking: 

- `DimIndices` works for StepRange now and is generally much more like `CartesianIndices`. So it does _not_ hold a LookupArray in its dimensions. Just ranges. This means selectors will not work on it, and should not - its a simple indexing object without lookups.
- LinearIndexing on stack makes sense in terms of dimensions - `st[i]` is now `st[Dimindices(st)[i]]`
- `AbstractDimStack` propagates for longer in indexing - if not all layers are scalars the scalar layers become zero dimensional `fill(x)` even in `getindex`.

### DIFF
--- a/src/DimensionalData.jl
+++ b/src/DimensionalData.jl
@@ -66,7 +66,7 @@ export AbstractDimStack, DimStack
 
 export AbstractDimTable, DimTable
 
-export DimIndices, DimKeys, DimPoints
+export DimIndices, DimSelectors, DimPoints, #= deprecated =# DimKeys
 
 # getter methods
 export dims, refdims, metadata, name, lookup, bounds
@@ -87,6 +87,7 @@ include("name.jl")
 
 # Arrays
 include("array/array.jl")
+include("dimindices.jl")
 include("array/indexing.jl")
 include("array/methods.jl")
 include("array/matmul.jl")
@@ -98,7 +99,6 @@ include("stack/indexing.jl")
 include("stack/methods.jl")
 include("stack/show.jl")
 # Other
-include("dimindices.jl")
 include("tables.jl")
 # Combined (easier to work on these in one file)
 include("plotrecipes.jl")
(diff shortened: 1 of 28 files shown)


rafaqz:
@sethaxen is this close to being ready.

I don't expect a full review as its huge and patchy. But you should be aware of this as it will effect your types.

The two biggest changes for `AbstractDimStack` are:
- Indexing an `AbstractDimStack` always returns another `AbstractDimStack` unless all layers are scalars. Than mean getindex makes `fill(x)` layers sometimes. I think you suggested this should be the behaviour in the past!
- Linear indexing on `AbstractDimStack` is now correct (I hope), all indexing go via `Dimension` wrappers (currently indexing with `Int` will map that over all layers, which is fine for same sized stacks but wrong with mixed layers.)

sethaxen:
Cool! I don't have time to review, but I ran all of my test suites using this PR, and I didn't encounter any new issues.

rafaqz:
Thanks! Great they all pass. Mostly the things that broke were awful or buggy anyway and probably not used.

PR #618, opened 2024-02-09
allow selecting dimensions with predicates

rafaqz:
@meggart I couldn't resist its so little code

One question is should we put the function first so do blocks work, leave it last where the query normally goes in DD, or have both?


meggart:
Good question about changing the signature. I personally would probably leave the query in the end, to not create confusion. Would it be possible dispatch-wise to extend the signature to `query::Callable` instead of `Function` only? No big deal if it does not work for now and then I tend to build custom callable structs, especially for selector things.

PR #631, opened 2024-02-17
Continue main

lazarusA:
fixes docs and add template for groupby docs. And it should fix the current break on docs site.


asinghvi17:
Whoa - this looks great!

Just to keep track here - there are a couple of things that don't work with DocumenterVitepress:

- Nested lists
- Tables
- No line breaks between paragraphs - need to look into that!  Maybe two `\n` after each para?

asinghvi17:
Whoa - this looks great!

Just to keep track here - there are a couple of things that don't work with DocumenterVitepress:

- Nested lists
- Tables
- No line breaks between paragraphs - need to look into that!  Maybe two `\n` after each para?

PR #632, opened 2024-02-17
Refactor to new gh-pages deployment scheme

asinghvi17:


asinghvi17:
This is now mergeable IMO, we can do that whenever - although the docs should probably be revised a bit :D

PR #637, opened 2024-02-19
Make the vitepress path local, see what happens

asinghvi17:
If this works and is merged, it should solve @sethaxen's problem.


asinghvi17:
This is not working yet...

Issue #522, opened 2023-08-25
Stop using internals


rafaqz:
We use a few internal methods here where base functions are hard to extend, like `Base._mapreduce_dim`. We should aim to stop doing that.

PR #645, opened 2024-02-23
Extending Base.stack for DimArrays

brendanjohnharris:


src/array/methods.jl
```diff
@@ -131,15 +131,15 @@ end
     """
     function Base.eachslice(A::AbstractDimArray; dims)
         dimtuple = _astuple(dims)
-        if !(dimtuple == ()) 
+        if !(dimtuple == ())
```
rafaqz, line 134: Would be nice to remove all these whitespace changes so we can see the real changes

src/array/methods.jl
```diff
@@ -542,11 +542,49 @@ end
...
+    _A = Base._dim_stack(newdim, T, t, A)
+
+    if !comparedims(Bool, origdims...;
+        order=true, val=true, warn=" Can't `stack` AbstractDimArray, applying to `parent` object."
+    )
+        return _A
+    end
+
+    newdims = first(origdims)
+    newdims = ntuple(d -> d == newdim ? AnonDim() : newdims[d-(d>newdim)], length(newdims) + 1)
```
rafaqz, line 578: maybe make this a `do` block so its easier to read

src/array/methods.jl
```diff
@@ -542,11 +542,49 @@ end
...
 _cat_warn_string(D, message) = """
 `cat` cannot concatenate `Dimension`s, falling back to `parent` type:
-$message on dimension $D. 
+$message on dimension $D.
 
 To fix for `AbstractDimArray`, pass new lookup values as `cat(As...; dims=$D(newlookupvals))` keyword or `dims=$D()` for empty `NoLookup`.
 """
 
+function Base._typed_stack(::Colon, ::Type{T}, ::Type{S}, A, Aax=_iterator_axes(A)) where {T,S<:AbstractDimArray}
+    origdims = dims.(A)
```
rafaqz, line 551:
```suggestion
    origdims = map(dims, A)
```

src/array/methods.jl
```diff
@@ -542,11 +542,49 @@ end
...
+    )
+        return _A
+    end
+
+    newdims = first(origdims)
+    newdims = ntuple(d -> d == newdim ? AnonDim() : newdims[d-(d>newdim)], length(newdims) + 1)
+    DimArray(_A, newdims)
+end
+
+function Base.stack(dim::Dimension, A::AbstractVector{<:AbstractDimArray}; dims=nothing, kwargs...)
```
rafaqz, line 582: Whats the idea with `dim` as the first argument? 

The tests also don't hit this code.

brendanjohnharris:
Thanks for the suggestions! Went through and cleaned up the PR. 

The `Base.stack(dim, X; dims)` method lets you supply a `Dimension` to assign to the newly created array dimension; not strictly necessary, but more convenient than `set`ting or `rebuild`ing the resulting array to overwrite the new `AnonDim` (is an `AnonDim` a suitable way to treat a newly created dimension, as I've done here?).

Let me know if you have more thoughts on this, happy to work on it further :)

src/array/methods.jl
```diff
@@ -547,6 +547,86 @@ $message on dimension $D.
...
+  ↗ Z 1:2
+└────────────────────────────────────────────────────────────────┘
+
+julia> dims(dc, 3) == Z(1:2)
+true
+julia> parent(dc) == stack(map(parent, [da, db]), dims=3)
+true
+```
+"""
+function Base.stack(dim::Dimension, A::AbstractVector{<:AbstractDimArray}; dims=nothing, kwargs...)
```
rafaqz, line 618: Why not pass `dim` in `dims`?

This method signature is a bit strange , we usually try not to add varients on Base methods, we just allow `dims` to specify named dimensions.

rafaqz:
We try not to change base method signatures besides allowing `dims` to be named. To keep the mental model very simple.

I would just put that new `Dimension` in `dims` like `cat` does. If its a constructed dimension it replaces the existing one. Otherwise Type/Symbol etc just choose them. Just check `d isa Dimension`

You will also need to `format(newdims, A)` afterwards.

rafaqz:
Using a `Pair` sounds like a good solution. I guess putting it as the last dimension would be the majority use-case? (I rarely use `stack`)

Always `AnonDim` will require using `set` all the time with `AnonDim => somedim`. We may as well just put that pair in `stack` and skip the step.

src/array/methods.jl
```diff
@@ -542,11 +542,104 @@ end
...
+  ↗ Z 1:2
+└────────────────────────────────────────────────────────────────┘
+
+julia> dims(dc, 3) == Z(1:2)
+true
+julia> parent(dc) == stack(map(parent, [da, db]), dims=3)
+true
+```
+"""
+function Base.stack(A::AbstractVector{<:AbstractDimArray}; dims=Pair(ndims(A[1]), AnonDim()), kwargs...)
```
rafaqz, line 622:
```suggestion
function Base.stack(A::AbstractVector{<:AbstractDimArray}; dims=Pair(ndims(A[1]) + 1, AnonDim()), kwargs...)
```

?

src/array/methods.jl
```diff
@@ -542,11 +542,104 @@ end
...
+    dims isa Integer && (dims = dims => AnonDim())
+    dims isa Dimension && (dims = ndims(A[1])+1 => dims)
```
rafaqz, line 624:
```suggestion
    if dims isa Integer 
        dims = dims => AnonDim())
    elseif dims isa Dimension 
        dims = ndims(A[1])+1 => dims)
    end    
```

Assigning after `&&` is best avoided as the assignment is hard to see when skimming code

src/array/methods.jl
```diff
@@ -542,11 +542,104 @@ end
...
+        newdims = ntuple(ndims(B)) do d
+            if d == first(dims) # Use the new provided dimension
+                last(dims)
+            else
+                DimensionalData.dims(B, d)
+            end
+        end
+        newdims = format(newdims, B)
+
+        B = rebuild(B; dims=newdims)
```
rafaqz, line 638:
```suggestion
        B = rebuild(B; dims=format(newdims, B))
```

User specified dims are usually incomplete and possibly incorrect

src/array/methods.jl
```diff
@@ -547,6 +547,100 @@ $message on dimension $D.
 To fix for `AbstractDimArray`, pass new lookup values as `cat(As...; dims=$D(newlookupvals))` keyword or `dims=$D()` for empty `NoLookup`.
 """
 
+function Base._typed_stack(::Colon, ::Type{T}, ::Type{S}, A, Aax=_iterator_axes(A)) where {T,S<:AbstractDimArray}
```
rafaqz, line 550: How stable do you think these methods are... could we add a method to use `Base.stack` instead? What do we can from touching these internals?

I know we use internals elsewhere, but we should stop:
https://github.com/rafaqz/DimensionalData.jl/issues/522

src/array/methods.jl
```diff
@@ -547,6 +547,100 @@ $message on dimension $D.
...
+    _A = Base._dim_stack(newdim, T, t, A)
+
+    if !comparedims(Bool, origdims...;
+        order=true, val=true, warn=" Can't `stack` AbstractDimArray, applying to `parent` object."
+    )
+        return _A
+    end
+
+    newdims = first(origdims)
+    newdims = ntuple(length(newdims) + 1) do d
```
rafaqz, line 578: Is this type-stable?

src/array/methods.jl
```diff
@@ -547,6 +547,100 @@ $message on dimension $D.
...
+        if d == newdim
+            AnonDim()
+        else # Return the old dimension, shifted across once if it comes after the new dim
+            newdims[d-(d>newdim)]
+        end
+    end
+    DimArray(_A, newdims)
```
rafaqz, line 585:
```suggestion
    DimArray(_A, format(newdims))
```

src/array/methods.jl
```diff
@@ -547,6 +547,100 @@ $message on dimension $D.
...
+    _A = parent.(A)
+    t = eltype(_A)
+    _A = Base._typed_stack(:, T, t, A)
+
+    if !comparedims(Bool, origdims...;
+        order=true, val=true, warn=" Can't `stack` AbstractDimArray, applying to `parent` object."
+    )
+        return _A
+    else
+        DimArray(_A, (first(origdims)..., AnonDim()))
```
rafaqz, line 561:
```suggestion
        DimArray(_A, format((first(origdims)..., AnonDim()), _A))
```

rafaqz:
@brendanjohnharris any updates? I can also finish this if you don't have time

brendanjohnharris:
Sorry fell off my radar, my bad
How about something much simpler? `Base.stack` now seems to work as expected as long as the input collection is a DimArray, except that if the elements have different dims the resulting stacked array uses the dims of the first element.
This latest commit just adds a `comparedims` check to `Base.stack()` and, if this check fails, returns the stacked parent array (like `cat`).
Then it's not much work at all to construct a `DimVector{<:DimArray}` with the new dimensions as needed

rafaqz:
Perfect, simpler is better!

But we could also allow Dimension `dims` using `dimnum`?

src/array/methods.jl
```diff
@@ -529,6 +529,14 @@ $message on dimension $D.
 To fix for `AbstractDimArray`, pass new lookup values as `cat(As...; dims=$D(newlookupvals))` keyword or `dims=$D()` for empty `NoLookup`.
 """
 
+function check_stack_dims(iter)
+    comparedims(Bool, dims.(iter)...; order=true, val=true, msg=Dimensions.Warn(" Can't `stack` AbstractDimArray, applying to `parent` object."))
+    iter
+end
+Base.stack(iter::AbstractArray{<:AbstractDimArray}; dims=:) = Base._stack(dims, check_stack_dims(iter))
```
rafaqz, line 536:
```suggestion
Base.stack(iter::AbstractArray{<:AbstractDimArray}; dims=:) = Base._stack(dimnum(first(iter), dims), check_stack_dims(iter))
```
    
This should allow passing Dimension/Symbol etc

brendanjohnharris: What about:
```julia
Base._stack(dims::Dimension, iter) = Base._stack(typeof(dims), iter)
Base._stack(dims::Type{<:Dimension}, iter) = Base._stack(dimnum(first(iter), dims), iter)
```
This helps because the default `dims` in Base's `Base.stack` is (oddly) `dims=:`. This has the same effect as `dims=ndims(first(iter))+1`, as far as I can tell, but it doesn't play well with `dimnums` in this context.

rafaqz: Try to avoid dispatch on the underscore as much as possible... And we also need to accept Symbol. I would just define another method. That has dispatch for Colon and everything else uses `dimnum`

test/methods.jl
```diff
@@ -802,6 +802,47 @@ end
...
+    sx = stack(x)
+    @test sx isa AbstractDimArray
+    @test dims(sx) == (dims(first(x))..., dims(x)...)
+
+    @test_throws "ArgumentError" stack(x; dims=4)
+
+    x = DimArray([da, ca], (Dim{:a}(1:2),))
+    sx = stack(x; dims=1)
+    sy = @test_nowarn stack(x; dims=:)
+    sz = @test_nowarn stack(x; dims=X)
```
rafaqz, line 825: What happens with `dims=Z` ?

src/array/methods.jl
```diff
@@ -529,6 +529,53 @@ $message on dimension $D.
...
+        return dim
+    else
+        throw(ArgumentError(LazyString("cannot stack slices ndims(x) = ", ndims(x) + 1, " along dims = ", dim)))
+    end
+end
+_maybe_dimnum(_, ::Colon) = Colon()
+_maybe_dimnum(x, dims::Tuple) = _maybe_dimnum.([x], dims)
+
+"""
+    stack(x::AbstractDimArray; [dims])
```
rafaqz, line 549: We will need to add this to the API docs page next to `cat`

test/methods.jl
brendanjohnharris: Currently, `dims=Z` (where Z is not a dimension in the DimArrays being stacked) falls back to adding the new dimension at the end (so the result is the same as for the default behavior):
```julia
_maybe_dimnum(x, dim) = hasdim(x, dim) ? dimnum(x, dim) : ndims(x) + 1
```

Should we throw an error instead, like when an out-of-range Integer dim is give? I.e.:
```julia
function _maybe_dimnum(x, dim::Int)
    if dim < ndims(x) + 2
        return dim
    else
        throw(ArgumentError(LazyString("cannot stack slices ndims(x) = ", ndims(x) + 1, " along dims = ", dim)))
    end
end
```

rafaqz: I was thinking we should add a dimension like it is currently, but we can set it to be a `Z` dimension

brendanjohnharris: Using the current minimal approach, stacking with a new dimension `Z` would be:
```julia
a = [1 2 3; 4 5 6]
da = DimArray(a, (X(4.0:5.0), Y(6.0:8.0)))
b = [7 8 9; 10 11 12]
db = DimArray(b, (X(4.0:5.0), Y(6.0:8.0)))
x = DimArray([da, db], (Dim{:a}(1:2),))

stack(set(x, :a=>Z)) # Or
set(stack(x), :a=>Z})
```

We could have this behavior be automatic with:
```julia
function Base.stack(iter::AbstractArray{<:AbstractDimArray}; dims=:)
    x = Base._stack(_maybe_dimnum(first(iter), dims), check_stack_dims(iter))
    if !hasdim(x, dims) && Z isa Union{Dimension,Type{<:Dimension}}
        x = set(x, DimensionalData.dims(x)[end] => dims)
    end
    return x
end
```
But I wonder if it could be confusing to have this different behavior for when `dims <: Dimension` and `hasdim(x, dims)` (the new dimension is inserted before `dims` but keeps the original name) versus `!hasdims(x, dims)` (the new dimension is inserted at the end and renamed to `dims`). It also means the return type can't be inferred

rafaqz: > falls back to adding the new dimension at the end (so the result is the same as for the default behavior)

I'm not sure we are understanding each other... My only ask here was that this new dimension is a `Z` dimension if `dims=Z` was used

PR #651, opened 2024-02-25
Breaking: document and tweak stack interface

rafaqz:
Notably this breaks the `AbstrackDimStack` methods:

- `first`
- `last`
- `iterate`
- `length`

These should never have worked how they do (NamedTuple-like). Now they just throw an error - we can do something better and more Array-like with them later.

@sethaxen a few more breaks, hopefully wont break anything that matters.


sethaxen:
It seems I didn't rely on any of these, so these changes don't break any of my tests.

PR #652, opened 2024-02-28
Increase Compat of DataAPI to 1.16

felixcremer:
This is needed because we need the DataAPI.groupby function to be available.
On lower DataAPI versions loading DimensionalData fails.


rafaqz:
Haha how did I forget that its my change 😭

PR #665, opened 2024-03-10
fix nightly bug

### LINKED ISSUE #664: Failing on nightly because uses private methods from Base
x-ref https://github.com/jump-dev/JuMP.jl/pull/3707

https://github.com/rafaqz/DimensionalData.jl/blob/55e8ecf6fd91249caa76af5697eeb66d9f8a40d0/src/Lookups/beginend.jl#L42-L45

rafaqz:
Closes #664


rafaqz:
@odow hopefully this is enough, the printing bug looks like a range/date bug in Base.

PR #675, opened 2024-03-23
remove broken version in vitepress

rafaqz:


docs/src/.vitepress/config.mts
```diff
@@ -15,12 +12,16 @@ export default defineConfig({
   description: "Datasets with named dimensions",
   lastUpdated: true,
   cleanUrls: true,
-  ignoreDeadLinks: true,
```
lazarusA, line 18: ahh... by ignoring the ignoring. But, this should be supported @asinghvi17  (there will be cases with dead links)

asinghvi17: Will see if this works first!  I've pushed a new tag object (we can always delete it later) so we can see if the docs build updates.

PR #682, opened 2024-04-03
Remove .vitepress/theme/index.ts

asinghvi17:
See if this fixes the hydration error which is nuking `stable`...


asinghvi17:
ah this wasn't supposed to be merged, since it breaks the docs build :)

rafaqz:
Haha I missed the "see if" part!

A ** DONT MERGE ** in the title would help

PR #688, opened 2024-04-05
rm icon links

lazarusA:
as less as possible content in landing page. Minimalist hero page.

Same component as in Tyler.jl, which is working.

### DIFF
--- a/.github/workflows/Documenter.yml
+++ b/.github/workflows/Documenter.yml
@@ -43,6 +43,8 @@ jobs:
         uses: julia-actions/cache@v1
       - name: Generate logo
         run: julia --project=docs -e "using Pkg; Pkg.instantiate()"; julia --project=docs/ --color=yes docs/logo.jl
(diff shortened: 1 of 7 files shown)


asinghvi17:
It's probably not the icon links - GeometryOps and SwarmMakie work fine without them, and so did DocumenterVitepress.  My bet is that it's some other change...

lazarusA:
Yes, indeed is none of them. Now, all missing things are fixed, including the `ico` file. 
Also, I tested the following cached files:

Screenshot 2024-04-05 at 21 04 57

they are not getting updated after each new released tag. They should, we need to find a way to clear these files from the cache and generated new ones, the correct ones. It should be an option for `vite` in the config file.... at the moment I don't know how. I will need to look into it.

rafaqz:
Should we merge this?

lazarusA:
is just small fixes. So, yes merging is ok.

PR #694, opened 2024-04-12
Add internal docs for metadata and array behind Extended help wrappers

felixcremer:
This adds docstrings to internal functions in metadata and array which are hidden behind an `Extended help` layer. The idea is to provide the docstring, but to make clear that the functions are internal. 
The docstrings could clearly be improved, they are based on my current understanding from reading the code. 
If you like the overall look of this, I might be going through more of the internal functions in the next weeks to add more internal documentation.

I also added docstrings to `DimVector` and similar types following the docstring examples for AbstractVector. 
In the REPL this would look like this for internal functions:

```julia
help?> DD._dimlength
  Internal function

  ─────────────────────────────────────────────────────────────────────────

Extended help is available with `??`

help?> ?DD._dimlength
  Internal function

  Extended help
  ≡≡≡≡≡≡≡≡≡≡≡≡≡

  _dimlength(dim)

  Return the length of the dimension dim or in case of a Tuple the length
  of every dimension in the Tuple.

```

### DIFF
--- a/src/Lookups/metadata.jl
+++ b/src/Lookups/metadata.jl
@@ -14,6 +14,12 @@ objects to [`set`](@ref) without ambiguity about where to put them.
 abstract type AbstractMetadata{X,T} end
 
 const _MetadataContents = Union{AbstractDict,NamedTuple}
+"""
+Internal type union.
+
+Possible types that can be used as metadata. 
+Contains AbstractMetadata an AbstractDict
+"""
 const AllMetadata = Union{AbstractMetadata,AbstractDict}
 
 Base.get(m::AbstractMetadata, args...) = get(val(m), args...)
@@ -96,6 +102,14 @@ end
 
 # Metadata utils
 
+"""
+Internal function.
+
+# Extended help
+
+    metadatadict(dict)
+Convert all keys of dict to Symbol and wrap it into a standard Dict.
+""" 
 function metadatadict(dict)
     symboldict = Dict{Symbol,Any}()
     for (k, v) in dict
--- a/src/array/array.jl
+++ b/src/array/array.jl
@@ -10,8 +10,24 @@ Only keyword `rebuild` is guaranteed to work with `AbstractBasicDimArray`.
 """
 abstract type AbstractBasicDimArray{T,N,D<:Tuple} <: AbstractArray{T,N} end
 
+""" 
+    AbstractBasicDimVector
+
+Supertype for basic dimensional arrays with one dimension and
+elements of type T. Alias for AbstractBasicDimArray{T,1}.
+"""
 const AbstractBasicDimVector = AbstractBasicDimArray{T,1} where T
+""" 
+    AbstractBasicDimMatrix
+
+Supertype for basic dimensional arrays with two dimension and
+elements of type T. Alias for AbstractBasicDimArray{T,2}.
+"""
 const AbstractBasicDimMatrix = AbstractBasicDimArray{T,2} where T
+"""
+    AbstractBasicDimVecOrMat
+Union type of AbstractBasicDimVector and AbstractBasicDimMatrix
+"""
 const AbstractBasicDimVecOrMat = Union{AbstractBasicDimVector,AbstractBasicDimMatrix}
 
 refdims(::AbstractBasicDimArray) = ()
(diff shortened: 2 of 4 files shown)


src/array/array.jl
```diff
@@ -10,8 +10,24 @@ Only keyword `rebuild` is guaranteed to work with `AbstractBasicDimArray`.
...
 const AbstractBasicDimVector = AbstractBasicDimArray{T,1} where T
+""" 
+    AbstractBasicDimMatrix
+
+Supertype for basic dimensional arrays with two dimension and
+elements of type T. Alias for AbstractBasicDimArray{T,2}.
+"""
 const AbstractBasicDimMatrix = AbstractBasicDimArray{T,2} where T
+"""
+    AbstractBasicDimVecOrMat
```
rafaqz, line 28:
```suggestion
    AbstractBasicDimVecOrMat
    
```

src/array/array.jl
```diff
@@ -58,9 +74,24 @@ Indexing `AbstractDimArray` with non-range `AbstractArray` has undefined effects
...
 const AbstractDimVector = AbstractDimArray{T,1} where T
+""" 
+    AbstractBasicDimMatrix
+
+Supertype for dimensional arrays with two dimension and
+elements of type T. Alias for AbstractDimArray{T,2}.
+"""
 const AbstractDimMatrix = AbstractDimArray{T,2} where T
+"""
+    AbstractDimVecOrMat
```
rafaqz, line 92:
```suggestion
    AbstractDimVecOrMat
    
```

src/array/array.jl
```diff
@@ -163,6 +194,16 @@ Base.similar(A::AbstractDimArray, ::Type{T}, i::Integer, I::Vararg{Integer}) whe
 Base.similar(A::AbstractDimArray, ::Type{T}, I::Tuple{Int,Vararg{Int}}) where T =
     similar(parent(A), T, I)
 
+
+"""
+    MaybeDimUnitRange
```
rafaqz, line 199:
```suggestion
    MaybeDimUnitRange
    
```

src/array/array.jl
```diff
@@ -440,8 +504,22 @@ function DimArray(f::Function, dim::Dimension; name=Symbol(nameof(f), "(", name(
      DimArray(f.(val(dim)), (dim,); name)
 end
 
+"""
+    DimVector{T}
```
rafaqz, line 508:
```suggestion
    DimVector{T}
    
```

src/array/array.jl
```diff
@@ -440,8 +504,22 @@ function DimArray(f::Function, dim::Dimension; name=Symbol(nameof(f), "(", name(
...
+One-dimensional concrete `DimArray` with elements of type `T`.
+This is an alias for DimArray{T,1}.
+"""
 const DimVector = DimArray{T,1} where T
+"""
+    DimMatrix{T}
```
rafaqz, line 514:
```suggestion
    DimMatrix{T}

```

src/array/array.jl
```diff
@@ -440,8 +504,22 @@ function DimArray(f::Function, dim::Dimension; name=Symbol(nameof(f), "(", name(
...
+Two-dimensional concrete `DimArray` with elements of type `T`.
+This is an alias for DimArray{T,2}.
+"""
 const DimMatrix = DimArray{T,2} where T
+"""
+    DimVecOrMat
```
rafaqz, line 520:
```suggestion
    DimVecOrMat
    
```

src/Lookups/metadata.jl
```diff
@@ -96,6 +102,14 @@ end
 
 # Metadata utils
 
+"""
+Internal function.
```
rafaqz, line 106: Does "Internal function" do something or its just a marker?

I'm not used to adding this much documentation, don't we just want this as a code comment?

asinghvi17: IMO this is very nice - when you see it in a stack trace, you can run `?` on it and immediately see what it does.  For people who are good at Julia, not 100% immersed in the DD world, and want to debug an error - this is priceless.

felixcremer: I am wondering, whether the marker "Internal function" is even needed nowadays, because with the public keyword, and the changes in Base handling of the show of non-public functions this is rather handled by Base.

felixcremer:
In Julia 1.11 the Internal function thing and the extended help is not needed anymore because it would look like this:

```julia
help?> ?DimensionalData._dimlength
  │ Warning
  │
  │  The following bindings may be internal; they may change or be removed in future versions:
  │
  │    •  DimensionalData._dimlength

  Internal function

  Extended help
  ≡≡≡≡≡≡≡≡≡≡≡≡≡

  _dimlength(dim)

  Return the length of the dimension dim or in case of a Tuple the length of every dimension in the Tuple.
```

The first part with the | signs in front is inserted from the REPL and the Warning is in yellow, so we could just add the docstring normally to the  non-public functions.

PR #697, opened 2024-04-21
add dimtrait and use it for plots

rafaqz:
This PR means arbitrary dimension names can also plot on the right axes, by associating them with `X`/`Y` etc as a trait.

If anyone has opinions on which names should be assigned to which dimensions (e.g. eleveation to Z) please comment on the PR.

Maybe @lazarusA you have some ideas ?


lazarusA:
if not names are available I would do "X", "Y" and "Z", not elevation, it might be confusing, because it could be also time or levels, but that should be left to the user to decide. Otherwise, the thing coming out from the trait.

rafaqz:
Ahhh maybe your right. People can already choose the dimension in the plot maybe thats enough for now.

PR #700, opened 2024-04-21
Update get_info.md

alex-s-gardner:
change `using DimensionalData.Lookups` to `using DimensionalData.LookupsArrays`


rafaqz:
Lookups is the new syntax, you may just be on an old version.

PR #707, opened 2024-04-30
[WIP] Use DataAPI.metadata instead of redefining it

asinghvi17:


rafaqz:
May need to check we don't have `function metadata end` or `metadata(x) = nothing` lying around

rafaqz:
We have `function metadata end` in interfaces.jl that needs removing. 

And these methods are going to be a piracy problem:

```julia
  [2] metadata(::Tuple{})
     @ DimensionalData.Dimensions ~/.julia/dev/DimensionalData/src/Dimensions/dimension.jl:232
 [15] metadata(ds::Tuple, I)
     @ DimensionalData.Dimensions ~/.julia/dev/DimensionalData/src/Dimensions/dimension.jl:234
 [16] metadata(ds::Tuple, i1, I...)
     @ DimensionalData.Dimensions ~/.julia/dev/DimensionalData/src/Dimensions/dimension.jl:233
 [17] metadata(ds::Tuple)
     @ DimensionalData.Dimensions ~/.julia/dev/DimensionalData/src/Dimensions/dimension.jl:231
 [18] metadata(x)
     @ ~/.julia/dev/DimensionalData/src/Lookups/metadata.jl:107
```

asinghvi17:
Thanks - will take a look today!

asinghvi17:
Ah damn, I just realized that the way that DD handles metadata (`DD.NoLookup`) is not the same way as DataAPI handles metadata (`Nothing`).  I guess this could potentially stay the same for now, and maybe just hook into `DataAPI.metadata` as well for DD native types...

rafaqz:
From discussion: we could instead forward DataAPI.metadata to DimensionalData.metadata so at least they both work.

rafaqz:
Actually thinking about just culling all the DD metadata handling so we use a Dict or NamedTuple or `nothing`.

It will be annoying to check keys but its worth it to use `DataAPI.metadata`

PR #710, opened 2024-05-10
Bump Makie compat + patch version

asinghvi17:


asinghvi17:
```
  Invalid attribute colormap for plot type MakieCore.Plot{Makie.series}.
```
is the current error.  I'm not sure where that attribute is coming from, though.

rafaqz:
Maybe theres a default value somewhere...

lazarusA:
it looks like `series`, not longer has a `colormap` attribute. Which we were using to extended the amount of colours if we had more than 7 lines. https://github.com/rafaqz/DimensionalData.jl/blob/7e6d97a166ff6e5e8de2ca7e39a17f1d565dbdf3/ext/DimensionalDataMakie.jl#L251

(it looks like now we should remove it, and let `Makie.series` do the work)

https://github.com/MakieOrg/Makie.jl/blob/746b03f46282f2385d438b01b486a74ea10f023d/src/basic_recipes/series.jl#L69

PR #711, opened 2024-05-10
Update Makie versions

Datseris:
DD.jl causes lots of downgrades at the moment due to this.


asinghvi17:
Some of the syntax has to change for v0.21.  I'm working on this in #710 :D

rafaqz:
Yeah DD hits a lot of plot types so its not so easy

Issue #552, opened 2023-10-13
Series Makie plot  for more than seven elements is broken


felixcremer:
While running the tests I got the following error. 
I am not sure, whether this is basically a user error and we should change the test or whether we would like to change the behaviour if the number of categories in the other dimension would be higher than the number of colors we have available. 

```
    A2r = rand(Y(10:10:100), X(['a', 'b', 'c']))
ssssssssssssssssssssssssssssssss
julia>     fig, ax, _ = M.series(A2r; labeldim=Y)
ERROR: Not enough colors for number of categories. Categories: 10, colors: 7
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] categorical_colors
    @ ~/.julia/packages/Makie/NrrbP/src/conversions.jl:1222 [inlined]
  [3] categorical_colors(cs::Symbol, categories::Int64)
    @ Makie ~/.julia/packages/Makie/NrrbP/src/conversions.jl:1237
  [4] (::Makie.var"#2121#2129"{Int64})(color::Symbol, scolor::Nothing)
    @ Makie ~/.julia/packages/Makie/NrrbP/src/basic_recipes/series.jl:74
  [5] #map#157
    @ ~/.julia/packages/Makie/NrrbP/src/scenes.jl:185 [inlined]
  [6] map(f::Makie.var"#2121#2129"{Int64}, scene::Combined{Makie.series, Tuple{Vector{Vector{Point{2, Float32}}}}}, arg1::Observable{Any}, args::Observable{Any})
    @ Makie ~/.julia/packages/Makie/NrrbP/src/scenes.jl:182
  [7] plot!(plot::Combined{Makie.series, Tuple{Vector{Vector{Point{2, Float32}}}}})
    @ Makie ~/.julia/packages/Makie/NrrbP/src/basic_recipes/series.jl:72
  [8] plot!(scene::Scene, P::Type{Combined{Makie.series, Tuple{DimensionalData.Dimensions.LookupArrays.NoLookup{Base.OneTo{Int64}}, Matrix{Float64}}}}, attributes::Attributes, input::Tuple{Observable{DimensionalData.Dimensions.LookupArrays.NoLookup{Base.OneTo{Int64}}}, Observable{Matrix{Float64}}}, args::Observable{Tuple{Vector{Vector{Point{2, Float32}}}}})
    @ Makie ~/.julia/packages/Makie/NrrbP/src/interfaces.jl:398
[...]

PR #712, opened 2024-05-12
fixes series and updates Makie compat to latest

lazarusA:
this should close https://github.com/rafaqz/DimensionalData.jl/pull/710, https://github.com/rafaqz/DimensionalData.jl/issues/552


rafaqz:
Something has changed in `rainclouds` too I guess

lazarusA:
volumeslices and series are special cases. They don't fit into the current workflow, since series takes an input a matrix (which usually is assigned to a surface-like call, but now it needs to be a point-like one) and `volumeslices` is actually a `heatmap` call and not a `volume` one.  This hasn't been supported so far. Hence I think we should move with this and open two specific issues for these uses cases.

rafaqz:
Is this good to go?

lazarusA:
Series mutation  method is not supported.  The rest should be fine.

rafaqz:
Is there something that can be done to fix it? Or should I just merge

PR #716, opened 2024-05-17
extend CategoricalArrays

tiemvanderdeure:
Extend CategoricalArrays, so `categorical(x::DimArray)` returns a `DimArray` with categorical values, and common (documented) operations on categorical arrays work on such categorical DimArrays.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -95,4 +95,4 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [targets]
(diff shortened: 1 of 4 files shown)


rafaqz:
Thanks! Good to have

For my understanding is there a reason these things don't "just work" already? Does CategoricalArrays.jl not use `similar` or not accept `AbstractArray` ?

tiemvanderdeure:
They don't work already. `categorical` errors due to an ambiguous dispatch of `copyto!`, `cut` does not error but returns a `CategoricalArray` (without dimensions). All of the other methods dispatch on `CategoricalArray`.

rafaqz:
Ah right, they haven't written it to be extensible at all.

Issue #721, opened 2024-06-07
Would it be useful to have Not() as a selector?


alex-s-gardner:
Found myself wanting for a Not() selector today and was wondering if others would find it useful

rafaqz:
Ehhh... We do?

alex-s-gardner:
Of course there is :-)

So I was trying `Not("april")` instead of `Not(At("april")` I then searched `Not` in the documentation but searching such a common word doesn't return anything useful. I then looked at the dcoumition for Selectors:

 https://rafaqz.github.io/DimensionalData.jl/dev/api/lookuparrays#DimensionalData.Dimensions.Lookups.Selector

and couldn't find anything there either.

using help with DimensionalData doesn't show the  functionality for wrapping a Selector i.e. `Not(Selector("april"))` 
```
help?> Not()
  InvertedIndex(idx)
  Not(idx)

  Construct an inverted index, selecting all indices not in the passed idx.

  Upon indexing into an array, the InvertedIndex behaves like a 1-dimensional collection of the indices of the array that are not in idx.
  Bounds are checked to ensure that all indices in idx are within the bounds of the array — even though they are skipped. The InvertedIndex
  behaves like a 1-dimensional collection of its inverted indices. If idx spans multiple dimensions (like a multidimensional logical mask
  or CartesianIndex), then the inverted index will similarly span multiple dimensions.

  When indexing into a NamedTuple, the InvertedIndex can wrap either a Symbol, a vector of Symbols, or a tuple of Symbols and selects the
  fields of the NamedTuple that are not in idx.
```  

Looking again this morning I found Not() under Lookups:
https://rafaqz.github.io/DimensionalData.jl/dev/selectors#Lookups

rafaqz:
If you saw other places we could document `Not`, feel free to add them.

felixcremer:
Isnt this closed?

PR #726, opened 2024-06-11
Add 'Not` to list in selector.jl

alex-s-gardner:
Not sure if `Not` should be considered a `Selector` or a `Selector Modifier`

Addresses #721

### DIFF
--- a/src/Lookups/selector.jl
+++ b/src/Lookups/selector.jl
@@ -30,8 +30,10 @@ Selectors provided in DimensionalData are:
 - [`Near`](@ref)
 - [`Where`](@ref)
 - [`Contains`](@ref)
-- [`Not`](@ref)
 
+Note: Selectors can be modified using:
+- [`Not`](@ref)
+- [`..`](@ref)
 """
 abstract type Selector{T} end
 
--- a/src/Lookups/selector.jl
+++ b/src/Lookups/selector.jl
@@ -33,7 +33,11 @@ Selectors provided in DimensionalData are:
 
 Note: Selectors can be modified using:
 - [`Not`](@ref)
-- [`..`](@ref)
+And IntervalSets.jl `Interval` can be used instead of `Between`
+- `..`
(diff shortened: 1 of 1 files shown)


rafaqz:
Problem is the others are all `<:Selector` and Not is not!

Can we instead make a note that `..` and `Not` also exist and work like real Selectors?

alex-s-gardner:
> Problem is the others are all `<:Selector` and Not is not!
> 
> Can we instead make a note that `..` and `Not` also exist and work like real Selectors?

So `Not` and `..` are already listed as `Selectors` here: https://rafaqz.github.io/DimensionalData.jl/dev/selectors#Selectors

rafaqz:
Yes...they are selectors in the sense of selecting things. 

But what you are changing is the docstring the abstract supertype `Selector`.... and its _not_ the supertype of `Not` or `..`.

We can either just close this, or add the note, but we cant list them as being actually related to `Selector` in the programming sense, because they are not.

src/Lookups/selector.jl
```diff
@@ -31,6 +31,9 @@ Selectors provided in DimensionalData are:
 - [`Where`](@ref)
 - [`Contains`](@ref)
 
+Note: Selectors can be modified using:
+- [`Not`](@ref)
```
rafaqz, line 35:
```suggestion
- `Not`
```

rafaqz: Ref cant work accross packages...

src/Lookups/selector.jl
```diff
@@ -31,6 +31,9 @@ Selectors provided in DimensionalData are:
...
+- [`..`](@ref)
```
rafaqz, line 36:
```suggestion
And IntervalSets.jl `Interval` can be used instead of `Between`
- `..`
- `Interval`
- `OpenInterval`
- `ClosedInterval`
```

PR #730, opened 2024-06-14
Add tests and fix rand(MersenneTwister, dims)

felixcremer:
This is to add some tests and I fixed the rand with an RNG but without the type information.
I will continue adding to this this morning, but I wanted to already run it on CI.

### DIFF
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -97,7 +97,6 @@ end
     @test reo_s == s
     @test dims(reo_s) == dims(s)
 
-    @test_throws MethodError reorder(da, rand(2,3))
 
     @testset "reorder handles extra dimensions" begin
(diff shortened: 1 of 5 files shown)


felixcremer:
This is good to go, I worked on the other branches rather than these tests.

test/utils.jl
```diff
@@ -97,6 +97,7 @@ end
     @test reo_s == s
     @test dims(reo_s) == dims(s)
 
+    @test_throws MethodError reorder(da, rand(2,3))
```
rafaqz, line 100:
```suggestion
```

Is this needed here? seems unrelated

rafaqz (approved):

test/utils.jl
felixcremer: This is unrelated to the fixes of rand but I actual started this PR to work on test coverage therefore it ended up in there. I actually misunderstood how the dispatch in reorder works. Therefore yes we can just remove it.

PR #739, opened 2024-06-18
Breaking: Materialize `DimArray` or `DimStack` From a Table

### LINKED ISSUE #335: Add a Tables.materializer method
This will help getting `DimArray`/`DimStack` as a return value from operations on tables.

JoshuaBillson:
#### Description:

resolves #335 

This PR aims to let users construct either a `DimStack` or `DimArray` from a table with one or more coordinate columns.

Unlike the existing contructor, rows may be out of order or even missing altogether.

#### Performance:

The algorithm is O(n), requiring two forward passes for each dimension to determine the correct order of rows.

**1000x1000:** 0.005 Seconds

**2000x2000:** 0.025 Seconds

**4000x4000:** 0.108 Seconds

**8000x8000:** 0.376 Seconds

#### Example:
```julia
julia> r = DimArray(rand(UInt8, 1000, 1000), (X, Y));

julia> t = r |> DataFrame |> Random.shuffle!;

julia> restored = DimArray(t, dims(r));

julia> all(r .== restored)
true
```

#### Next Steps:

1. Finalize method signatures.
2. Decide what to do with missing rows. Currently, users may choose a missing value to fill in (`missing` by default).
3. Add support for a `:geometry` column.
4. Write test cases.
5. Update docs.


src/table_ops.jl
```diff
@@ -0,0 +1,77 @@
...
+# Get names of data columns from table
+function _data_col_names(table, dims::Tuple)
+    dim_cols = name.(dims)
+    return filter(x -> !(x in dim_cols), Tables.columnnames(table))
+end
+
+# Determine the ordinality of a set of numerical coordinates
+function _coords_to_ords(coords::AbstractVector, dim::AbstractVector{<:Real})
+    stride = (last(dim) - first(dim)) / (length(dim) - 1)
+    return round.(UInt32, ((coords .- first(dim)) ./ stride) .+ 1)
```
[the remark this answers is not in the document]
JoshuaBillson: Do you mean something like this?

```julia
# Determine the ordinality of a set of regularly spaced numerical coordinates with a starting locus
function _coords_to_ords(
    coords::AbstractVector, 
    dim::Dimension, 
    ::Type{<:Real},
    ::DimensionalData.Start, 
    ::DimensionalData.Regular)
    step = (last(dim) - first(dim)) / (length(dim) - 1)
    return floor.(Int, ((coords .- first(dim)) ./ step) .+ 1)
end

# Determine the ordinality of a set of regularly spaced numerical coordinates with a cenered locus
function _coords_to_ords(
    coords::AbstractVector, 
    dim::Dimension, 
    ::Type{<:Real},
    ::DimensionalData.Center, 
    ::DimensionalData.Regular)
    step = (last(dim) - first(dim)) / (length(dim) - 1)
    return round.(Int, ((coords .- first(dim)) ./ step) .+ 1)
end

# Determine the ordinality of a set of regularly spaced numerical coordinates with an end locus
function _coords_to_ords(
    coords::AbstractVector, 
    dim::Dimension, 
    ::Type{<:Real},
    ::DimensionalData.End, 
    ::DimensionalData.Regular)
    step = (last(dim) - first(dim)) / (length(dim) - 1)
    return ceil.(Int, ((coords .- first(dim)) ./ step) .+ 1)
end

# Determine the ordinality of a set of categorical or irregular coordinates
function _coords_to_ords(coords::AbstractVector, dim::Dimension, ::Any, ::Any, ::Any)
    return map(c -> DimensionalData.selectindices(dim, At(c)), coords)
end

# Preprocessing methods for _coords_to_ords
_coords_to_ords(coords::AbstractVector, dim::Dimension) = _coords_to_ords(coords, dim, eltype(dim), locus(dim), span(dim))
_coords_to_ords(coords::Tuple, dims::Tuple) = Tuple(_coords_to_ords(c, d) for (c, d) in zip(coords, dims))
_coords_to_ords(coords::NamedTuple, dims::Tuple) = _coords_to_ords(Tuple(coords[d] for d in name.(dims)), dims)
```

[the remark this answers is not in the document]
rafaqz: Yep pretty much.

Except use `Contains` for `Intervals` rather than `At`. We could allow users to pass in the selector too, so they can use `Near` or a tolerance with `At`. The default selector can be `nothing` to let use choose specific defaults, or not use selectors at all in the other methods.

(Also note `Tuple(x for x in sometuple)` is not the best pattern for the compiler as it loses the length of the tuple in the iteration, and has to lift it back to a type at runtime. `map` keeps the length so its preferable for small tuples.)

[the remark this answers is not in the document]
JoshuaBillson: I like the idea of letting users choose a selector, but I'm not sure how well it will work in practice. The problem is that `selectindices` is orders of magnitude slower for large regularly spaced dimensions, so it could be a performance footgun. I suppose we could only use it for dimensions caught by the fourth method, but this might result in unexpected behaviour for end users.

JoshuaBillson:
There's still a few questions that need resolving:

1. What do we do with missing rows? This can be handled explicitly in `Rasters.jl`, but `DimArrays` have no concept of missing values. We could let users choose a missing value to write, or we could encode it explicitly with `missing`. We could also disallow the existence of missing rows, but I don't think this is an ideal solution.
2. How should we handle a `:geometry` column? I know that several packages in the Julia ecosystem are using this convention, but I don't have much experience with them. Can we assume `:geometry` will contain tuples of coordinates, or could they also be some sort of geometry like `Meshes.Point`?

src/table_ops.jl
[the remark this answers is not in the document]
rafaqz: Sure. But regular At is pretty strict, they need to be the exact values from the lookup.

It should only be one order of magnitude slower for range dims? But yes vector dims will be slow.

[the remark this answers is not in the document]
JoshuaBillson: Here's the results for 3 000 000 coordinates, where Y only has length 1000:

```julia
julia> @btime DimensionalData._coords_to_ords(coords, dims(r,Y), Int, DimensionalData.End(), DimensionalData.Regular());
  3.832 ms (11 allocations: 22.89 MiB)

julia> @btime DimensionalData._coords_to_ords(coords, dims(r,Y), 1, 1, 1);
  949.727 ms (11 allocations: 22.89 MiB)
```

This would take several seconds to reconstruct a 1000 x 1000 array. It would be far slower for anything of a reasonable size.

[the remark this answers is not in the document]
rafaqz: What are the dims?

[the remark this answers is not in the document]
rafaqz: They should be the same speed... Maybe Base.OneTo is missing optimisations for `searchsorted` in base Julia.

In that case we should stop using it!

There shouldn't be a slow path for ranges

asinghvi17:
I think it's best to test that the geometry column's elements have `GI.PointTrait` - then you can always extract e.g. `GI.x(point)` and the same for `y`, `z`, and `m`.  You can also interrogate the dimension via `GI.is3d, GI.ismeasured`.

rafaqz:
DimensionalData.jl does not depend on GeoInterface

JoshuaBillson:
> DimensionalData.jl does not depend on GeoInterface

Being able to interact with geometries in a generic fashion would help us interop with other packages. `GeoInterface`'s only dependency is `Extents`, which we already depend on. After importing `DimensionalData`, `GeoInterface` loads in 0.008 seconds.

I also see that `Rasters` already depends on `GeoInterface`. We could simply ignore the `:geometry` column in `DimensionalData`, then implement the additional functionality in `Rasters`. However, I think the problem is general enough to be implemented here.

rafaqz:
It's not about deps or timing, it's about clean feature scope. 

The promise here is that Rasters (and YAX) have all the geo deps and features, so non-geo people don't have to worry about them. Some of the biggest contributors here use DD for unrelated fields.

I would ignore point columns here entirely and instead write the code so it's easy for Rasters to handle them. Taking the underscores off a few functions and documenting them as an real interface will help that.

JoshuaBillson:
I've updated the docstrings for both the `DimArray` and `DimStack` constructors. By default, materializers will now use the `Contains` selector for irregular and non-numeric coordinates, with the option to specify an alternative when desired. We're exporting the `restore_array` and `coords_to_index` methods to be used by downstream packages like `Rasters.jl`. I've also written several test cases to show that we can handle tables with out of order and missing rows. If everything looks good, I think we can go ahead and merge this PR.

src/DimensionalData.jl
```diff
@@ -77,13 +77,17 @@ export dimnum, hasdim, hasselection, otherdims
 # utils
 export set, rebuild, reorder, modify, broadcast_dims, broadcast_dims!, mergedims, unmergedims
 
+# table utils
+export restore_array, coords_to_index
```
rafaqz, line 81: Let's not export these, just  document that they can be used as api

src/array/array.jl
```diff
@@ -334,7 +334,7 @@ end
     DimArray <: AbstractDimArray
 
     DimArray(data, dims, refdims, name, metadata)
-    DimArray(data, dims::Tuple; refdims=(), name=NoName(), metadata=NoMetadata())
+    DimArray(data, dims::Tuple; refdims=(), name=NoName(), metadata=NoMetadata(), selector=Contains)
```
rafaqz, line 337: Better to pass a constructed selector `Contains()` that we rebuild. Then `At` can have `atol`

src/array/array.jl
```diff
@@ -429,6 +430,13 @@ function DimArray(A::AbstractBasicDimArray;
     newdata = collect(data)
     DimArray(newdata, format(dims, newdata); refdims, name, metadata)
 end
+# Write a single column from a table with one or more coordinate columns to a DimArray
+function DimArray(table, dims; name=NoName(), selector=DimensionalData.Contains, kw...)
+    indices = coords_to_index(table, dims; selector=selector)
```
rafaqz, line 435:
```suggestion
    indices = coords_to_index(table, dims; selector)
```

src/table_ops.jl
```diff
@@ -0,0 +1,163 @@
...
+julia> coords_to_index(t, dims(d))
+65536-element Vector{Int64}:
+     1
+     2
+     ⋮
+ 65535
+ 65536
+```
+"""
+function coords_to_index(table, dims::Tuple; selector=DimensionalData.Contains)
```
rafaqz, line 69:
```suggestion
function coords_to_index(table, dims::Tuple; selector=DimensionalData.Contains())
```

src/table_ops.jl
```diff
@@ -0,0 +1,163 @@
...
+    ::Type{<:Real},
+    ::DimensionalData.End, 
+    ::DimensionalData.Regular)
+    step = (last(dim) - first(dim)) / (length(dim) - 1)
+    return ceil.(Int, ((coords .- first(dim)) ./ step) .+ 1)
+end
+
+# Determine the ordinality of a set of categorical or irregular coordinates
+function _coords_to_ords(coords::AbstractVector, dim::Dimension, ::Type{T}, ::Any, ::Any, ::Any) where {T<:DimensionalData.Selector}
+    return map(c -> DimensionalData.selectindices(dim, T(c)), coords)
```
rafaqz, line 137:
```suggestion
    return map(c -> DimensionalData.selectindices(dim, rebuild(sel, c)), coords)
```

src/table_ops.jl
```diff
@@ -0,0 +1,163 @@
...
+    coords::AbstractVector, 
+    dim::Dimension, 
+    ::Type{<:DimensionalData.Selector},
+    ::Type{<:Real},
+    ::DimensionalData.Start, 
+    ::DimensionalData.Regular)
+    step = (last(dim) - first(dim)) / (length(dim) - 1)
+    return floor.(Int, ((coords .- first(dim)) ./ step) .+ 1)
+end
+
```
rafaqz, line 110:
```suggestion
```

src/table_ops.jl
```diff
@@ -0,0 +1,163 @@
...
+function _data_col_names(table, dims::Tuple)
+    dim_cols = name(dims)
+    return filter(x -> !(x in dim_cols), Tables.columnnames(table))
+end
+
+# Determine the ordinality of a set of regularly spaced numerical coordinates with a starting locus
+function _coords_to_ords(
+    coords::AbstractVector, 
+    dim::Dimension, 
+    ::Type{<:DimensionalData.Selector},
```
rafaqz, line 103:
```suggestion
    ::DimensionalData.Selector,
```

src/table_ops.jl
```diff
@@ -0,0 +1,163 @@
...
+    ::DimensionalData.Regular)
+    step = (last(dim) - first(dim)) / (length(dim) - 1)
+    return floor.(Int, ((coords .- first(dim)) ./ step) .+ 1)
+end
+
+# Determine the ordinality of a set of regularly spaced numerical coordinates with a central locus
+function _coords_to_ords(
+    coords::AbstractVector, 
+    dim::Dimension, 
+    ::Type{<:DimensionalData.Selector},
```
rafaqz, line 115:
```suggestion
    ::DimensionalData.Selector,
```

src/table_ops.jl
```diff
@@ -0,0 +1,163 @@
...
+    ::DimensionalData.Regular)
+    step = (last(dim) - first(dim)) / (length(dim) - 1)
+    return round.(Int, ((coords .- first(dim)) ./ step) .+ 1)
+end
+
+# Determine the ordinality of a set of regularly spaced numerical coordinates with an end locus
+function _coords_to_ords(
+    coords::AbstractVector, 
+    dim::Dimension, 
+    ::Type{<:DimensionalData.Selector},
```
rafaqz, line 127: We need dispatch on Contains/Near/At here as At should fail if it's not exact or within atol, and Near should use the end value if the data is outside the lookup range.

Contains falls back to exact At for `Points`.

This is a bit annoying to implement but the behaviour should really be the same as for selectors in other places.

src/table_ops.jl
```diff
@@ -0,0 +1,163 @@
...
+    ::Type{<:Real},
+    ::DimensionalData.End, 
+    ::DimensionalData.Regular)
+    step = (last(dim) - first(dim)) / (length(dim) - 1)
+    return ceil.(Int, ((coords .- first(dim)) ./ step) .+ 1)
+end
+
+# Determine the ordinality of a set of categorical or irregular coordinates
+function _coords_to_ords(coords::AbstractVector, dim::Dimension, ::Type{T}, ::Any, ::Any, ::Any) where {T<:DimensionalData.Selector}
```
rafaqz, line 136:
```suggestion
function _coords_to_ords(coords::AbstractVector, dim::Dimension, sel::DimensionalData.Selector, ::Any, ::Any, ::Any)
```

rafaqz (changes requested):
This is good, but a few minor changes are needed for it to be correct. Constructed selectors should be allowed so At can have `atol`. We can just construct selector types at the outer level (they're just filled with `nothing`.

Then there is a bit more dispatch needed to make the fast paths correct for At/Near/Contains with their standard behaviour.

rafaqz:
I've been wondering if we can also detect the dimensions? 

So find the minimum and maximum values and the smallest step (after sorting)? Then check if the step is consistent and choose Regular/Irregular?

rafaqz:
Yeah we probably need a tolerance check.

We could allow specifying if some dimensions are not sorted, like `dims=(X=ForwardOrdered(), Y=ReverseOrdered, C=Unordered())`

src/table_ops.jl
JoshuaBillson: I'm not quite sure that I follow. Are you saying that `At` should throw an error when any of the values fall outside of the tolerance? Should this be the case for all methods dispatching on regular numerical dimensions, or just the method you highlighted? 

If so, then we could first compute the floating point index for each value, then compare each value with the integer index produced by `round`/`ceil`/`floor`. If the difference between the two falls outside of the tolerance for any element, then we throw an error.
```julia
float_ords = ((coords .- first(dim)) ./ step) .+ 1
int_ords = ceil.(Int, float_ords)
invalid_value = findfirst(>(tol), abs.(float_ords .- int_ords))
if !isnothing(invalid_value)
    DD.Lookups._selnotfound(dim, rebuild(sel, coords[invalid_value]))
end
return int_ords
```

In the case of `Near`, we could probably just clamp the indices to `[1, length(dim)]` to handle values outside the lookup range.

rafaqz: > Are you saying that At should throw an error when any of the values fall outside of the tolerance

Yes, and if a value doesn't exactly match at lookup value when there is no tolerance.

Then, `Near` should never error for any numerical/date etc values (clamping may be an efficient way to do that), and `Contains` should error when none of the intervals in the lookup contain the value (so for `Regular` lookups its `Near` with no clamping).

JoshuaBillson:
I've implemented the `guess_dims` method to infer dimensions directly from the data. It should work for both sorted and shuffled data and can handle missing rows. If the data is sorted, you shouldn't need to specify the ordering, since it can be inferred directly. In the case of shuffled data, you will need to give the order as a `Dim => Order` pair. If the names of your coordinate columns can be safely guessed (currently `:X`, `:Y`, `:Z`, `:Ti`, and `:Band`) and the data is sorted, then we should be able to infer the dimensions without any input from the user.

Here's an example of it in action:
```julia
julia> xdims = X(LinRange{Float64}(610000.0, 661180.0, 2560));

julia> ydims = Y(LinRange{Float64}(6.84142e6, 6.79024e6, 2560));

julia> bdims = Dim{:Band}([:B02, :B03, :B04]);

julia> d = DimArray(rand(UInt16, 2560, 2560, 3), (xdims, ydims, bdims));

julia> t = DataFrame(d);

julia> t_rand = Random.shuffle(t);

julia> dims(d)
↓ X    Sampled{Float64} LinRange{Float64}(610000.0, 661180.0, 2560) ForwardOrdered Regular Points,
→ Y    Sampled{Float64} LinRange{Float64}(6.84142e6, 6.79024e6, 2560) ReverseOrdered Regular Points,
↗ Band Categorical{Symbol} [:B02, :B03, :B04] ForwardOrdered

julia> DD.guess_dims(t)
↓ X    Sampled{Float64} LinRange{Float64}(610000.0, 661180.0, 2560) ForwardOrdered Regular Points,
→ Y    Sampled{Float64} LinRange{Float64}(6.84142e6, 6.79024e6, 2560) ReverseOrdered Regular Points,
↗ Band Categorical{Symbol} [:B02, :B03, :B04] ForwardOrdered

julia> DD.guess_dims(t, (X, Y, :Band))
↓ X    Sampled{Float64} LinRange{Float64}(610000.0, 661180.0, 2560) ForwardOrdered Regular Points,
→ Y    Sampled{Float64} LinRange{Float64}(6.84142e6, 6.79024e6, 2560) ReverseOrdered Regular Points,
↗ Band Categorical{Symbol} [:B02, :B03, :B04] ForwardOrdered

julia> DD.guess_dims(t_rand, (X => DD.ForwardOrdered(), Y => DD.ReverseOrdered(), :Band => DD.ForwardOrdered()))
↓ X    Sampled{Float64} LinRange{Float64}(610000.0, 661180.0, 2560) ForwardOrdered Regular Points,
→ Y    Sampled{Float64} LinRange{Float64}(6.84142e6, 6.79024e6, 2560) ReverseOrdered Regular Points,
↗ Band Categorical{Symbol} [:B02, :B03, :B04] ForwardOrdered

julia> DD.guess_dims(t_rand[1:10000,:], (X => DD.ForwardOrdered(), Y => DD.ReverseOrdered(), :Band => DD.ForwardOrdered()))
↓ X    Sampled{Float64} LinRange{Float64}(610000.0, 661180.0, 2560) ForwardOrdered Regular Points,
→ Y    Sampled{Float64} LinRange{Float64}(6.84142e6, 6.79024e6, 2560) ReverseOrdered Regular Points,
↗ Band Categorical{Symbol} [:B02, :B03, :B04] ForwardOrdered
```

As can be seen in the last example, `guess_dims` works even when we select only a small subset of the shuffled rows.

rafaqz:
This is awesome to have, thanks.

It may be a bit before I can fully review and test it but it looks good from a quick skim.

src/stack/stack.jl
```diff
@@ -278,6 +278,7 @@ end
 """
     DimStack <: AbstractDimStack
 
+    DimStack(table, dims; kw...)
```
rafaqz, line 281:
```suggestion
    DimStack(table, [dims]; kw...)
```

src/table_ops.jl
```diff
@@ -0,0 +1,307 @@
...
+ 65535
+ 65536
+```
+"""
+function coords_to_indices(table, dims::Tuple; selector=DimensionalData.Near())
+    return _coords_to_indices(table, dims, selector)
+end
+
+# Find the order of the table's rows according to the coordinate values 
+_coords_to_indices(table, dims::Tuple, sel::DimensionalData.Selector) = _coords_to_indices(_dim_cols(table, dims), dims, sel)
```
rafaqz, line 97:
```suggestion
_coords_to_indices(table, dims::Tuple, sel::DimensionalData.Selector) = 
    _coords_to_indices(_dim_cols(table, dims), dims, sel)
```

src/table_ops.jl
```diff
@@ -0,0 +1,307 @@
...
+end
+
+"""
+    guess_dims(table; kw...)
+    guess_dims(table, dims; precision=6)
+
+Guesses the dimensions of an array based on the provided tabular representation.
+
+# Arguments
+- `table`: The input data table, which could be a `DataFrame`, `DimTable`, or any other tabular data structure.
```
rafaqz, line 111:
```suggestion
- `table`: The input data table, which could be a `DataFrame`, `DimTable`, or any other Tables.jl compatible data structure. 
```

src/array/array.jl
```diff
@@ -429,6 +430,13 @@ function DimArray(A::AbstractBasicDimArray;
     newdata = collect(data)
     DimArray(newdata, format(dims, newdata); refdims, name, metadata)
 end
+# Write a single column from a table with one or more coordinate columns to a DimArray
+function DimArray(table, dims; name=NoName(), selector=DimensionalData.Near(), precision=6, kw...)
+    data = restore_array(table, dims; selector=selector, missingval=missing, name=name, precision=precision)
```
rafaqz, line 435: We should probably do a `Tables.istable` check at this point for a nicer error and stack trace, rather than failing deep in the internals.

src/stack/stack.jl
```diff
@@ -420,5 +421,12 @@ function DimStack(data::NamedTuple, dims::Tuple;
     all(map(d -> axes(d) == axes(first(data)), data)) || _stack_size_mismatch()
     DimStack(data, format(dims, first(data)), refdims, layerdims, metadata, layermetadata)
 end
+# Write each column from a table with one or more coordinate columns to a layer in a DimStack
+function DimStack(table, dims::Tuple; selector=DimensionalData.Contains(), kw...)
+    data_cols = _data_cols(table, dims)
```
rafaqz, line 426: Again we probably need a `Tables.istable` check here

src/table_ops.jl
```diff
@@ -0,0 +1,307 @@
...
+end
+
+# Get names of data columns from table
+function _data_col_names(table, dims::Tuple)
+    dim_cols = DD.name(dims)
+    return filter(x -> !(x in dim_cols), Tables.columnnames(table))
+end
+
+# Determine the ordinality of a set of coordinates
+_coords_to_ords(coords::AbstractVector, dim::DD.Dimension, sel::DD.Selector) = _coords_to_ords(coords, dim, sel, DD.locus(dim), DD.span(dim))
```
rafaqz, line 202: Are all of these DD and DimensionalData qualifiers needed?

src/table_ops.jl
```diff
@@ -0,0 +1,307 @@
...
+```
+"""
+guess_dims(table; kw...) = guess_dims(table, filter(x -> x in Tables.columnnames(table), (:X,:Y,:Z,:Ti,:Band)); kw...)
+guess_dims(table, dims::Tuple; kw...) = map(dim -> guess_dims(table, dim; kw...), dims)
+guess_dims(table, dim; precision=6) = _guess_dims(_get_column(table, dim), dim, precision)
+
+_guess_dims(coords::AbstractVector, dim::DD.Dimension, args...) = dim
+_guess_dims(coords::AbstractVector, dim::Type{<:DD.Dimension}, args...) = _guess_dims(coords, DD.name(dim), args...)
+_guess_dims(coords::AbstractVector, dim::Pair, args...) = _guess_dims(coords, first(dim), last(dim), args...)
+function _guess_dims(coords::AbstractVector, dim::Symbol, precision::Int)
```
rafaqz, line 167: Should this be `AbstractVector{Union{Number,DateTime}` ?

I'm wondering what happens to strings, symbols and other objects that need to go in `Categorical` lookups.

src/table_ops.jl
```diff
@@ -0,0 +1,307 @@
...
+function _guess_dim_span(coords::AbstractVector{<:Real}, ::DD.Ordered, precision::Int)
+    steps = round.((@view coords[2:end]) .- (@view coords[1:end-1]), digits=precision)
+    span = argmin(abs, steps)
+    return all(isinteger, round.(steps ./ span, digits=precision)) ? DD.Regular(span) : DD.Irregular()
+end
+
+function _build_dim(vals::AbstractVector, dim::Symbol, order::DD.Order, ::DD.Span)
+    return Dim{dim}(DD.Categorical(vals, order=order))
+end
+function _build_dim(vals::AbstractVector{<:Real}, dim::Symbol, order::DD.Order, span::DD.Irregular)
```
rafaqz, line 270:
```suggestion
function _build_dim(vals::AbstractVector{<:Union{Number,DateTime}}, dim::Symbol, order::DD.Order, span::DD.Irregular)
```

src/table_ops.jl
```diff
@@ -0,0 +1,307 @@
...
+    return Dim{dim}(DD.Sampled(vals, order=order, span=span, sampling=DD.Points()))
+end
+function _build_dim(vals::AbstractVector{<:Real}, dim::Symbol, order::DD.Order, span::DD.Regular)
```
rafaqz, line 273:
```suggestion
function _build_dim(vals::AbstractVector{<:Union{Number,DateTime}}, dim::Symbol, order::DD.Order, span::DD.Regular)
```

src/table_ops.jl
```diff
@@ -0,0 +1,307 @@
...
+
+_get_column(table, x::Type{<:DD.Dimension}) = Tables.getcolumn(table, DD.name(x))
+_get_column(table, x::DD.Dimension) = Tables.getcolumn(table, DD.name(x))
+_get_column(table, x::Symbol) = Tables.getcolumn(table, x)
+_get_column(table, x::Pair) = _get_column(table, first(x))
+
+_dim_name(x::Symbol) = x
+_dim_name(x::DD.Dimension) = DD.name(x)
+_dim_name(x::Type{<:DD.Dimension}) = DD.name(x)
+_dim_name(x::Pair) = _dim_name(first(x)) => last(x)
```
rafaqz, line 287:
```suggestion
```
Doesn't seem to be used

src/table_ops.jl
```diff
@@ -0,0 +1,307 @@
...
+    elseif issorted(coords, rev=true)
+        return DD.ReverseOrdered()
+    else
+        return DD.Unordered()
+    end
+end
+
+# Estimate the span between consecutive coordinates
+_guess_dim_span(::AbstractVector, ::DD.Order, ::Int) = DD.Irregular()
+_guess_dim_span(::AbstractVector{<:Real}, ::DD.Order, ::Int) = DD.Irregular()
```
rafaqz, line 260:
```suggestion
_guess_dim_span(::AbstractVector{<:Union{Number,DateTime}}, ::DD.Order, ::Int) = DD.Irregular()
```

src/table_ops.jl
```diff
@@ -0,0 +1,307 @@
...
+function _guess_dim_span(coords::AbstractVector{<:Real}, ::DD.Ordered, precision::Int)
```
rafaqz, line 261:
```suggestion
function _guess_dim_span(coords::AbstractVector{<:Union{Number,DateTime}}, ::DD.Ordered, precision::Int)
```

src/table_ops.jl
```diff
@@ -0,0 +1,307 @@
...
+    steps = round.((@view coords[2:end]) .- (@view coords[1:end-1]), digits=precision)
+    span = argmin(abs, steps)
+    return all(isinteger, round.(steps ./ span, digits=precision)) ? DD.Regular(span) : DD.Irregular()
+end
+
+function _build_dim(vals::AbstractVector, dim::Symbol, order::DD.Order, ::DD.Span)
+    return Dim{dim}(DD.Categorical(vals, order=order))
```
rafaqz, line 268: All of these should use `name2dim(dim)` instead of `Dim{dim}` to catch any `@dim` dimensions.

```suggestion
    return rebuild(name2dim(dim), DD.Categorical(vals, order=order))
```

src/array/array.jl
```diff
@@ -429,6 +430,13 @@ function DimArray(A::AbstractBasicDimArray;
     newdata = collect(data)
     DimArray(newdata, format(dims, newdata); refdims, name, metadata)
 end
+# Write a single column from a table with one or more coordinate columns to a DimArray
+function DimArray(table, dims; name=NoName(), selector=DimensionalData.Near(), precision=6, kw...)
```
rafaqz, line 434:
```suggestion
function DimArray(table, dims; name=NoName(), selector=Near(), precision=6, kw...)
```

rafaqz (changes requested):
Looks good, a bunch of minor fixes.

Proably the main questionn is how Number and DateTime is handled

rafaqz:
Any news on this PR @JoshuaBillson ? would be good to have this available

src/table_ops.jl
JoshuaBillson: It should work with almost any type. If the coordinates are non-numerical, then we will internally dispatch on the following methods:

```julia
# Extract all unique coordinates from the given vector
_unique_vals(coords::AbstractVector, precision::Int) = _round_dim_val.(coords, precision) |> unique

# Round dimension value within the specified precision
_round_dim_val(x, ::Int) = x

# Determine if the given coordinates are forward ordered, reverse ordered, or unordered
function _guess_dim_order(coords::AbstractVector)
    if issorted(coords)
        return DD.ForwardOrdered()
    elseif issorted(coords, rev=true)
        return DD.ReverseOrdered()
    else
        return DD.Unordered()
    end
end

# Estimate the span between consecutive coordinates
_guess_dim_span(::AbstractVector, ::DD.Order, ::Int) = DD.Irregular()
```

`_unique_vals` will just return all unique values, where `_round_dim_val` is just the identity function for non-numerical coordinates. 

`_guess_dim_order` should work for anything that can be sorted, which is the case for both `String` and `Symbol`.

`_guess_dim_span` will return `DD.Irregular()` for non-numerical coordinates.

JoshuaBillson:
> Any news on this PR @JoshuaBillson ? would be good to have this available

Sorry, I've been busy with some other projects. I'll try to resolve your suggested fixes this week. Once that's done, I think we just need to update the docs for `DimStack` and `DimArray` and write some more test cases.

src/table_ops.jl
rafaqz: We used to need a try catch for `issorted` in case `<` is not defined for a type

JoshuaBillson: This should let us handle data that can't be sorted:

```julia
function _guess_dim_order(coords::AbstractVector)
    try
        if issorted(coords)
            return DD.ForwardOrdered()
        elseif issorted(coords, rev=true)
            return DD.ReverseOrdered()
        else
            return DD.Unordered()
        end
    catch 
        return DD.Unordered()
    end
end
```

And this should retrieve the span from `Date` and `DateTime` objects:

```julia
function _guess_dim_span(coords::AbstractVector{<:Dates.AbstractTime}, ::DD.Ordered, precision::Int)
    steps = (@view coords[2:end]) .- (@view coords[1:end-1])
    span = argmin(abs, steps)
    return all(isinteger, round.(steps ./ span, digits=precision)) ? DD.Regular(span) : DD.Irregular()
end
```

However, there seems to be a problem with constructing a `LinRange` from `Date` objects:

```julia
julia> vals = [Date("2022-11-16") + Day(i * 7) for i in 0:4];

julia> LinRange(first(vals), last(vals), 5)
5-element LinRange{Day, Int64}:
Error showing value of type LinRange{Day, Int64}:
ERROR: InexactError: Int64(553856.25)
```
Thus, I'm not sure how we should construct a `Dimension` with regularly spaced dates.

rafaqz: A StepRange works for Dates. Probably we should use StepRangeLen instead of LinRange where possible anyway

JoshuaBillson: That works. Do you want to use `StepRangeLen` for numerical coordinates or just dates?

Here's the result:
```julia
julia> xdims = X(LinRange{Float64}(610000.0, 661180.0, 2560));

julia> ydims = Y(LinRange{Float64}(6.84142e6, 6.79024e6, 2560));

julia> bdims = Dim{:Band}([:B02, :B03, :B04]);

julia> tdims = Dim{:Ti}([d1 + Day(i * 7) for i in 0:4]);

julia> d = DimArray(rand(UInt16, 2560, 2560, 3, 5), (xdims, ydims, bdims, tdims));

julia> t = DataFrame(d);

julia> DD.guess_dims(t)
↓ X    Sampled{Float64} 610000.0:20.0:661180.0 ForwardOrdered Regular Points,
→ Y    Sampled{Float64} 6.84142e6:-20.0:6.79024e6 ReverseOrdered Regular Points,
↗ Ti   Sampled{Date} Date("2024-11-18"):Day(7):Date("2024-12-16") ForwardOrdered Regular Points,
⬔ Band Categorical{Symbol} [:B02, :B03, :B04] ForwardOrdered
```

rafaqz: It's probably better for everything, except a bit slower. It didn't exist when I first wrote this package and uses of LinRange are just legacy from that.

src/array/array.jl
```diff
@@ -429,6 +430,35 @@ function DimArray(A::AbstractBasicDimArray;
     newdata = collect(data)
     DimArray(newdata, format(dims, newdata); refdims, name, metadata)
 end
+# Write a single column from a table with one or more coordinate columns to a DimArray
+function DimArray(table, dims; name=NoName(), selector=Near(), precision=6, missingval=missing, kw...)
+    # Confirm that the Tables interface is implemented
+    Tables.istable(table) || throw(ArgumentError("`table` must satisfy the `Tables.jl` interface."))
```
rafaqz, line 436:
```suggestion
    Tables.istable(table) || throw(ArgumentError("`obj` must be an `AbstractArray` or satisfy the `Tables.jl` interface."))
```

People will also hit this method if they do something weird like pass a non-AbstractArray to `DimArray`.

src/array/array.jl
```diff
@@ -334,7 +334,7 @@ end
     DimArray <: AbstractDimArray
 
     DimArray(data, dims, refdims, name, metadata)
-    DimArray(data, dims::Tuple; refdims=(), name=NoName(), metadata=NoMetadata())
+    DimArray(data, dims::Tuple; refdims=(), name=NoName(), metadata=NoMetadata(), selector=Contains())
```
rafaqz, line 337:
```suggestion
    DimArray(data, dims::Tuple; kw...)
```

tiemvanderdeure:
How close is this PR to the finish line? I'm doing some dataframes-heavy things right now and found myself in a situation a few times where this would been super helpful. Haven't looked into the nitty-gritty of the code, but this looks almost ready. I can do some final polishing if that's all it takes to make this available.

rafaqz:
Maybe try it out? I guess it needs a rebase too. I haven't had time or a use case yet to test it so not sure if it's finished

JoshuaBillson:
> Maybe try it out? I guess it needs a rebase too. I haven't had time or a use case yet to test it so not sure if it's finished

I believe it was 90% finished. We mostly need test cases and documentation. We'll also need to rebase, as you mentioned.

One thing I noted when reading the new documentation is the inclusion of `AutoValues()`, `AutoOrder()`, and `AutoSpan()` to automatically infer dimensional lookups from an array of values. Could we reuse this logic here to infer the values, order, and span instead of using custom methods? This would remove redundancy from the codebase and ensure consistency in how lookups are detected.

tiemvanderdeure:
Okay the rebase wasn't too hard.

There are a few method ambiguities - I'll sort those out.

Definitely a good idea to use the automatic detection of lookup types using `format`. It would also be really neat if it were possible to pass things like `X(Sampled(order = ForwardOrdered()))` to force the dimensions to look a certain way, where other fields are inferred from the data. Similar to dimensions how you pass partially formatted dimensions to `rand`

tiemvanderdeure:
Okay a question here is: how should this work for `AbstractArray{<:NamedTuple}`. Since this is both a table and an array?

Maybe we can materialize it as a table if DimArray is called without a dims argument, or if the dims are in also columns, and just make a DimArray{<:NamedTuple} otherwise?

rafaqz:
Can you maybbe demonstrate that with a MWE? I don't think I totally get the problem

tiemvanderdeure:
Now locally I've made it so that it checks if the names of the dimensions are in the columns or not. This ensures that all of this works:
```julia
x = X([:a, :b, :c])
y = Y([10.0, 20.0])
d = Dim{:test}(1.0:1.0:3.0)
dimz = x, y, d
da = DimArray(ones(3, 2, 3), dimz; name=:data)
table = Tables.rowtable(da)
DimArray(table, dimz) # returns a DimArray with x, y, test dimensions
DimArray(table, Z(1:18)) # returns a DimArray{<:NamedTuple} with a Z dimension
```
I'll push it later this afternoon and then we can discuss.

Another case is if you call `DimStack` on a `NamedTuple{<:AbstractVector}`. Now do the elements of the namedtuple represent layers of the stack, or columns of the table?

rafaqz:
Is the NamedTuple not both the dims and the layers like another table would be?

tiemvanderdeure:
> Is the NamedTuple not both the dims and the layers like another table would be?

Yes but there are edge cases where this is breaking. E.g.

```julia
nt = (X = 1:10, data = 1:10)
DimStack(nt, X(1:10))
```
```
julia> DimStack(nt, X(1:10))
┌ 10-element DimStack ┐
├─────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────── dims ┐
  ↓ X Sampled{Int64} 1:10 ForwardOrdered Regular Points
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── layers ┤
  :X    eltype: Int64 dims: X size: 10
  :data eltype: Int64 dims: X size: 10
└─────────────────────────────────────────────
```

Whereas on https://github.com/JoshuaBillson/DimensionalData.jl/pull/2 this returns
```
julia> DimStack(nt, X(1:10))
┌ 10-element DimStack ┐
├─────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────── dims ┐
  ↓ X Sampled{Int64} 1:10 ForwardOrdered Regular Points
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── layers ┤
  :data eltype: Int64 dims: X size: 10
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

```

rafaqz:
Ok so maybe we need to check for the dimensions columns and remove them? And have a keyword to not do that

tiemvanderdeure:
We could add an `astable` keyword for both of these cases (NamedTuple of Vector and Vector of NamedTuple)? Then if that it set to false it would never interpret a table as a table, basically preserving the previous behaviour. And it would default to true, which is to treat tables as tables.

tiemvanderdeure:
I just merged the commits from https://github.com/JoshuaBillson/DimensionalData.jl/pull/2 since we're not hearing from Joshua.

The merge conflicts are a bit tricky, actually. Because of the changes in https://github.com/rafaqz/DimensionalData.jl/pull/990. Git seems to be putting some of the code from this PR in illogical places.

tiemvanderdeure:
If you want to give it a go? Otherwise I can take a closer look at what changed in your PR but I don't know when I will have time

tiemvanderdeure:
There are some method ambiguities - I'll fix those.

Before I do that - I'm wondering if we can use abstract type for many of these constructors like `(::Type{T})(args...; kw...) where T <: AbstractDimArray`. The ambiguities are basically because we do that in some places and not in others. And it will also save us from copying lots of code over to Rasters.

Or does that not make sense for YAX?

rafaqz:
Those constructors are kind of dangerous for ambiguity if you want to change anything, so I'm not sure

tiemvanderdeure:
Okay docs fails because of this:
```
 Error: 5 docstrings not included in the manual:

    DimensionalData.restore_array :: Tuple{AbstractVector, AbstractVector, Tuple, Any}
    DimensionalData.coords_to_indices :: Tuple{Any, Tuple}
    DimensionalData.guess_dims :: Tuple{Any}
    DimensionalData.data_col_names :: Tuple{Any, Tuple}
    DimensionalData.get_column :: Tuple{Any, Type{<:Dimension}}
```

But anyway I think all of these functions should be internal. Should I just remove the docstrings - or make them into comments or something?

(Never seen this before so I'm not 100% about what this is testing)

rafaqz:
Yeah we force docstrings to be in the docs. I think these should just be comments not docstrings

tiemvanderdeure:
All right let's see if this passes now. I think before merging we should add some documentation, though.

Another thing is that this will only work with exactly `DimArray` now, not other abstract dimarrays or abstractdimstacks. We probably want this functionality with rasters and maybe yaxarrays as well, without creating dispatch mayhem. 

I've tried to solve this now by adding a type argument to `dimarray_from_table`, so you could call `dimarray_from_table(Raster, table, dims; kw...)` and the only lines you would have to copy are these:

https://github.com/JoshuaBillson/DimensionalData.jl/blob/aec86a9600d6ceb3af0aaa7dadc0f051c638a87d/src/array/array.jl#L519-L544

rafaqz:
Yes perfect that's why I removed the underscores, those a_from_b methods can become dev api

rafaqz:
@tiemvanderdeure this looks good to merge ?

tiemvanderdeure:
Yeah I think it is! I don't have more, if you've reviewed it and found nothing then do merge :)

tiemvanderdeure:
I think this got merged into main and released. Was that on purpose? It is very slightly breaking.

rafaqz:
It got merged into breaking! which I do want to merge now but it will get a breaking bump

tiemvanderdeure:
Ah I got confused because it is listed in the tag https://github.com/rafaqz/DimensionalData.jl/releases/tag/v0.29.21

Didn't know that it also listed PRs into other branches than main

rafaqz:
Oh I didn't know that either.. thats actually bad

PR #743, opened 2024-06-21
Add dispatch for Observable{AbstractDimArray} for plot!

felixcremer:
This adds a dispatch to plot! function for Observable{AbstractDimArray} to be able to also plot these.
This is not done yet, it would need the same dispatch also for the plot function and it needs some tests

### DIFF
--- a/ext/DimensionalDataMakie.jl
+++ b/ext/DimensionalDataMakie.jl
@@ -62,7 +62,7 @@ end
 for (f1, f2) in _paired(:plot => :scatter, :scatter, :lines, :scatterlines, :stairs, :stem, :barplot, :waterfall)
     f1!, f2! = Symbol(f1, '!'), Symbol(f2, '!')
     docstring = """
-        $f1(A::AbstractDimArray{<:Any,1}; attributes...)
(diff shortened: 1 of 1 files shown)


ext/DimensionalDataMakie.jl
```diff
@@ -161,6 +161,14 @@ for (f1, f2) in _paired(:plot => :heatmap, :heatmap, :image, :contour, :contourf
             # No ColourBar in the ! in-place versions
             return Makie.$f2!(axis, args...; attributes...)
         end
+        function Makie.$f1!(axis, A::Observable{<:AbstractDimMatrix};
```
rafaqz, line 164: Maybe also change all the ones above to also use `AbstractDimMatrix` for consistency, maybe it didn't exist back then...

rafaqz (approved):

ext/DimensionalDataMakie.jl
felixcremer: I changed them to AbstractDimMatrix and AbstractDimVector in the whole file.

PR #744, opened 2024-06-21
hack to fix stable links

lazarusA:
@rafaqz this uses some code from makie, which fixes their `stable` links issues. Let's see if also works here.


rafaqz:
Ahh thanks, wans't sure what to do about that

lazarusA:
we need a new tag. Maybe doing a dummy one, or just wait for a new release 😄 .

rafaqz:
Should I just merge?

PR #756, opened 2024-08-02
fixes leftovers from conflicts

lazarusA:
this should fix the annoying HEAD text in the groupby docstring.


felixcremer (commented):
Looks good to me

felixcremer (approved):

PR #757, opened 2024-08-02
do first and last instead

lazarusA:
do fist and last instead.


felixcremer (approved):
Looks good to me. This would enable OffsetArray lookups to plot correctly.

rafaqz (approved):

rafaqz (approved):

Issue #773, opened 2024-08-20
missing function in plot recipe


xzackli:
`_lookup_to_axis` doesn't appear to be defined anywhere, but is used in the Makie recipe.

https://github.com/rafaqz/DimensionalData.jl/blob/d10d0fc6bd2021487762d93eb711fb8084f7f567/ext/DimensionalDataMakie.jl#L384

rafaqz:
Ahh right I just changed it, we mustn't test plots going through that dispatch.

Do you have an example I can put in the tests?

PR #774, opened 2024-08-20
Changes _surface2 to use the correct conversion trait in Makie

ptiede:
I noticed that the recent change to allow for units seemed to break some of my tests. 
This is related to #773 and also impacted `ImageLike` since `_lookup_to_interval` had been mistyped as `_lookup_to_intervals`. Looking at the code, this wasn't caught in testing because `_surface2` always assumes that the conversion trait is `VertexGrid`.  

This PR tries to fix this by adding an additional argument to `_surface2` so that the specific conversion trait for that plot is used.

### DIFF
--- a/ext/DimensionalDataMakie.jl
+++ b/ext/DimensionalDataMakie.jl
@@ -42,12 +42,12 @@ Labels are found automatically, following this logic:
 
 $(_keyword_heading_doc(f))
 
-- `labeldim`: manual specify the dimension to use as series and get 
+- `labeldim`: manual specify the dimension to use as series and get
     the `labels` attribute from. Can be a `Dimension`, `Type`, `Symbol` or `Int`.
 """
 
 # Only `heatmap` and `contourf` get a colorbar
-function _maybe_colorbar_doc(f) 
+function _maybe_colorbar_doc(f)
     if f in (:heatmap, :contourf)
         """
         - `colorbarkw`: keywords to pass to `Makie.Colorbar`.
@@ -65,13 +65,13 @@ for (f1, f2) in _paired(:plot => :scatter, :scatter, :lines, :scatterlines, :sta
     f1!, f2! = Symbol(f1, '!'), Symbol(f2, '!')
     docstring = """
         $f1(A::AbstractDimVector; attributes...)
-        
+
     Plot a 1-dimensional `AbstractDimArray` with `Makie.$f2`.
 
     The X axis will be labelled with the dimension name and and use ticks from its lookup.
 
     $(_keyword_heading_doc(f1))
-    $AXISLEGENDKW_DOC     
+    $AXISLEGENDKW_DOC
     """
     @eval begin
         @doc $docstring
@@ -96,10 +96,10 @@ function _pointbased1(A, attributes; set_axis_attributes=true)
     args = Makie.convert_arguments(Makie.PointBased(), A2)
     # Plot attribute generation
     user_attributes = Makie.Attributes(; attributes...)
-    axis_attributes = if set_axis_attributes 
-        Attributes(; 
-            axis=(; 
-                xlabel=string(label(dims(A, 1))), 
+    axis_attributes = if set_axis_attributes
+        Attributes(;
+            axis=(;
+                xlabel=string(label(dims(A, 1))),
                 ylabel=DD.label(A),
                 title=DD.refdims_title(A),
             ),
@@ -107,7 +107,7 @@ function _pointbased1(A, attributes; set_axis_attributes=true)
     else
         Attributes()
     end
-    plot_attributes = Attributes(; 
+    plot_attributes = Attributes(;
         label=DD.label(A),
     )
     merged_attributes = merge(user_attributes, axis_attributes, plot_attributes, lookup_attributes)
@@ -124,7 +124,7 @@ for (f1, f2) in _paired(:plot => :heatmap, :heatmap, :image, :contour, :contourf
     f1!, f2! = Symbol(f1, '!'), Symbol(f2, '!')
     docstring = """
         $f1(A::AbstractDimMatrix; attributes...)
-        
+
     Plot a 2-dimensional `AbstractDimArray` with `Makie.$f2`.
 
     $(_keyword_heading_doc(f1))
@@ -133,11 +133,11 @@ for (f1, f2) in _paired(:plot => :heatmap, :heatmap, :image, :contour, :contourf
     """
     @eval begin
         @doc $docstring
-        function Makie.$f1(A::AbstractDimMatrix{T}; 
+        function Makie.$f1(A::AbstractDimMatrix{T};
             x=nothing, y=nothing, colorbarkw=(;), attributes...
         ) where T
             replacements = _keywords2dimpairs(x, y)
-            plottrait = Makie.conversion_trait(Makie.Plot{$f1})
+            plottrait = Makie.conversion_trait(Makie.Plot{$f2})
             A1, A2, args, merged_attributes = _surface2(A, plottrait, attributes, replacements)
             p = if $(f1 == :surface)
                 # surface is an LScene so we cant pass attributes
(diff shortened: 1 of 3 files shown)


asinghvi17 (commented):
Looks like the correct usage to me (but haven't tested it yet)

ptiede:
Ya sorry I'll fix the test failures tonight. I had to run just as a submitted the PR.

asinghvi17:
Ah nevermind, this is actually dangerous ish because you should ideally check whether there is a method for conversion with the plot type first, and only then go to traits.  Makie has that inbuilt but it's more on the Observable pipeline.  Will try to find it and send you a link @ptiede

ptiede:
Ok sounds good!

ptiede:
In the meantime, should I make a separate PR with the `_lookup_to_intervals` correction? I have some tests in some of my packages that are broken because of it.

rafaqz:
Sounds good to me.

asinghvi17:
The invocation is here:
https://github.com/MakieOrg/Makie.jl/blob/2b5931ee92c00603a669d02b8103bff49315024b/src/interfaces.jl#L272

and the main function is
https://github.com/MakieOrg/Makie.jl/blob/2b5931ee92c00603a669d02b8103bff49315024b/src/interfaces.jl#L211-L253

where you probably only want to use the logic that tries to convert using the plot type, checks that, then tries to convert using the conversion trait if the first conversion failed.

ext/DimensionalDataMakie.jl
```diff
@@ -148,44 +149,47 @@ for (f1, f2) in _paired(:plot => :heatmap, :heatmap, :image, :contour, :contourf
...
 
-function _surface2(A, attributes, replacements)
+function _surface2(A, plottrait, attributes, replacements)
     # Array/Dimension manipulation
     A1 = _prepare_for_makie(A, replacements)
     lookup_attributes, newdims = _split_attributes(A1)
     A2 = _restore_dim_names(set(A1, map(Pair, newdims, newdims)...), A, replacements)
-    args = Makie.convert_arguments(Makie.VertexGrid(), A2)
+
+    args = Makie.convert_arguments(plottrait, A2)
```
asinghvi17, line 186: If you know this is a dimarray and you don't care about expanding dimensions, the thing to do here is to try converting with the plottype first, then get the plottrait and convert using that if the specific conversion didn't work.

ptiede: Ok so we are guaranteed that `A` will be an AbstractDimMatrix. 
For you suggestions do you mean I could just do

```julia
Makie.convert_arguments(Plot{func}, A)
```
and check whether this worked? How will I know if this fails? Will it throw an exception or return a special type?

asinghvi17: Specifically this code,
https://github.com/MakieOrg/Makie.jl/blob/2b5931ee92c00603a669d02b8103bff49315024b/src/interfaces.jl#L223-L235

which shows how you can check and recurse based on how many times convert_arguments needs to be applied!

asinghvi17: You would also probably want to `expand_dimensions` first, to get those pesky units out of the way...

asinghvi17: Instead of recursion you can just call convert_arguments again

ptiede: I tried to match what Makie does more closely, but I am now getting a StackOverflow error when the dimensions have units. 

This no longer works.

```julia
A2u = rand(Y(10u"km":10u"km":100u"km"), X(['a', 'b', 'c']))
heatmap(A2u)
```

I am starting to feel that some of these problems may have to do with the extension overloading the `heatmap`, `plot`, `image`, etc. functions. Is there a more Makie way to do this? I know AxisKeysExtra.jl doesn't seem to need to resort to this kind of machinery https://github.com/JuliaAPlavin/AxisKeysExtra.jl/blob/master/ext/MakieExt.jl
To fix all of these issues, I'm starting to feel the need to reimplement some of Makie's conversion pipeline, which is pretty far beyond my comfort level.

ptiede: Looking into this further, Intervals doesn't play nicely with Unitful and Makie. For example,

```julia
x = 1..3
y = 10u"km"..100u"km"
image(x, y, rand(3, 10))
ERROR: MethodError: no method matching to_endpoints(::Tuple{Quantity{Int64, 𝐋, Unitful.FreeUnits{…}}, Quantity{Int64, 𝐋, Unitful.FreeUnits{…}}})

Closest candidates are:
  to_endpoints(::Any, ::Any)
   @ Makie ~/.julia/packages/Makie/aG6xp/src/conversions.jl:372
  to_endpoints(::IntervalSets.ClosedInterval)
   @ Makie ~/.julia/packages/Makie/aG6xp/src/conversions.jl:368
  to_endpoints(::Union{IntervalSets.Interval, AbstractVector})
   @ Makie ~/.julia/packages/Makie/aG6xp/src/conversions.jl:369
  ...

Stacktrace:
  [1] to_endpoints(x::IntervalSets.ClosedInterval{Quantity{Int64, 𝐋, Unitful.FreeUnits{(km,), 𝐋, nothing}}})
    @ Makie ~/.julia/packages/Makie/aG6xp/src/conversions.jl:368
  [2] to_endpoints(x::IntervalSets.ClosedInterval{Quantity{Int64, 𝐋, Unitful.FreeUnits{(km,), 𝐋, nothing}}}, dim::String)
    @ Makie ~/.julia/packages/Makie/aG6xp/src/conversions.jl:375
  [3] convert_arguments(::ImageLike, xs::IntervalSets.ClosedInterval{…}, ys::IntervalSets.ClosedInterval{…}, data::Matrix{…})
    @ Makie ~/.julia/packages/Makie/aG6xp/src/conversions.jl:422
  [4] convert_arguments(::Type{Image}, ::IntervalSets.ClosedInterval{Int64}, ::Vararg{Any}; kw::@Kwargs{})
    @ Makie ~/.julia/packages/Makie/aG6xp/src/conversions.jl:26
  [5] convert_arguments(::Type{…}, ::IntervalSets.ClosedInterval{…}, ::IntervalSets.ClosedInterval{…}, ::Matrix{…})
    @ Makie ~/.julia/packages/Makie/aG6xp/src/conversions.jl:14
  [6] conversion_pipeline(P::Type{…}, used_attrs::Tuple{}, args::Tuple{…}, args_obs::Tuple{…}, user_attributes::Dict{…}, deregister::Vector{…}, recursion::Int64)
    @ Makie ~/.julia/packages/Makie/aG6xp/src/interfaces.jl:223
  [7] conversion_pipeline(P::Type{…}, used_attrs::Tuple{}, args::Tuple{…}, args_obs::Tuple{…}, user_attributes::Dict{…}, deregister::Vector{…})
    @ Makie ~/.julia/packages/Makie/aG6xp/src/interfaces.jl:214
  [8] (Image)(user_args::Tuple{…}, user_attributes::Dict{…})
    @ Makie ~/.julia/packages/Makie/aG6xp/src/interfaces.jl:272
  [9] _create_plot(::Function, ::Dict{Symbol, Any}, ::IntervalSets.ClosedInterval{Int64}, ::Vararg{Any})
    @ Makie ~/.julia/packages/Makie/aG6xp/src/figureplotting.jl:316
 [10] image(::IntervalSets.ClosedInterval{Int64}, ::Vararg{Any}; kw::@Kwargs{})
    @ MakieCore ~/.julia/packages/MakieCore/fkSJS/src/recipes.jl:436
 [11] image(::IntervalSets.ClosedInterval{Int64}, ::IntervalSets.ClosedInterval{Quantity{…}}, ::Vararg{Any})
    @ MakieCore ~/.julia/packages/MakieCore/fkSJS/src/recipes.jl:434
 [12] top-level scope
    @ REPL[82]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

ptiede: Issue opened https://github.com/MakieOrg/Makie.jl/issues/4176

rafaqz: Yeah I just hit this the other day. Thanks for making the issue

rafaqz:
Is this good to go/review?

asinghvi17:
Will look later today, am currently on the road.

rafaqz:
@ptiede I would like to review and merge this, but it has a lot of whitespace changes and thing like commented lines

Is it possible to simplify it to just the code that is actively doing something? Cosmetic changes should happen separately.

rafaqz:
superceded by #793

PR #775, opened 2024-08-20
chore: improve broadcast_dims function to work directly with `Dimension`s

### LINKED ISSUE #748: FR: Make `broadcast_dims` work with `DimArray`s and `Dimension`s
Recording this idea from slack discussions in github.

It would be nice if I could do something like
```julia
broadcast_dims(some_function, A, X, Ti)
```
where `A` is a `DimArray` with dimensions (a subset of) `(X, Ti)`, and `some_function(,xi,ti)` is a function that takes (here) three arguments, where `a` is an elements from `A`, and `xi, ti` are elements from the lookup of each dimension.

Currently, you can do something like:
```julia
lon, ts = X(1:3), Ti(1:5)
A = ones(lon, ts)
broadcast_dims(+, A, DimArray(parent(lon), lon))
```
but it would be neat to do something like 
```julia
broadcast_dims(+, A, X)
```

-----

The most naive implementation I could come up with looks like this:
```julia
function broadcast_dims2(f, As::Union{DimensionalData.AbstractBasicDimArray, DimensionalData.Dimensions.Dimension, Type{<:DimensionalData.Dimension}}...)
	existing_dims = DimensionalData.combinedims(filter(A -> A isa DimensionalData.AbstractBasicDimArray, As)...)
	Bs = map(As) do A
		if A isa DimensionalData.Dimension
			DimArray(parent(A), A)
		elseif A isa Type{<:DimensionalData.Dimension}
			dim = dims(existing_dims, A)
			DimArray(parent(dim), dim)
		else
			A
		end
	end
	broadcast_dims(f, Bs...)
end
```
I had to do it this way because e.g. `combinedims(A, lon)` gives a `StackOverflowError` for me.

With this, I can use both the `Dimension` type (e.g. `X`) or a reference to a concrete dimension (e.g. `lon`)
```julia
broadcast_dims2(+, A, X, Ti)
╭─────────────────────────╮
│ 3×5 DimArray{Float64,2} │
├─────────────────────────┴──────────────────────── dims ┐
  ↓ X  Sampled{Int64} 1:3 ForwardOrdered Regular Points,
  → Ti Sampled{Int64} 1:5 ForwardOrdered Regular Points
└────────────────────────────────────────────────────────┘
 ↓ →  1    2    3    4    5
 1    3.0  4.0  5.0  6.0  7.0
 2    4.0  5.0  6.0  7.0  8.0
 3    5.0  6.0  7.0  8.0  9.0
```

haakon-e:
Closes #748.

### DIFF
--- a/docs/src/broadcast_dims.md
+++ b/docs/src/broadcast_dims.md
@@ -52,3 +52,16 @@ We can see the means of each month are scaled by the broadcast :
 mean(eachslice(data; dims=(X, Y)))
 mean(eachslice(scaled; dims=(X, Y)))
 ````
+
+Broadcasting also works directly over `Dimension`s (or references to them).
+For example, a new `DimArray` can be constructed by broadcasting a function over a set of dimensions:
+
+````@ansi bd
+broadcast_dims(*, x, y)
+````
+
+Existing dimensions can be referenced by name, in which case its lookup values are used:
+
+````@ansi bd
+broadcast_dims(*, data, X)  # or `:X` or `X(:)`
+````
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -108,23 +108,109 @@ function modify(f, index::AbstractArray)
 end
 
 """
-    broadcast_dims(f, sources::AbstractDimArray...) => AbstractDimArray
+    broadcast_dims(f, sources::Union{AbstractDimArray, Dimension, Symbol}...) => AbstractDimArray
 
-Broadcast function `f` over the `AbstractDimArray`s in `sources`, permuting and reshaping
-dimensions to match where required. The result will contain all the dimensions in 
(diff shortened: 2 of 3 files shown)


rafaqz (commented):
Should it also work with DimStack?

haakon-e:
I think so yes -- I'll try to look at this again next week.

haakon-e:
Thanks for pinging me again. Are there use cases for keeping both?
I've beens slowly exploring `@d` to understand how it fits my needs.

An alternative would be to deprecate `broadcast_dims`, if `@d` has all relevant functionality.

rafaqz:
I think there are still contexts to use a function, like we still have `broadcast`. 

Right now especially one of those is on `DimStack`

PR #792, opened 2024-09-01
DDMakieExt: fix trying to set OldAxis values if there is no OldAxis

asinghvi17:
I'm surprised this doesn't seriously break workflows - just encountered this in GeoMakie.


rafaqz:
These are hacks, there need to be real ways to do this in LScene

Issue #791, opened 2024-08-30
implement `@broadcast_dims` macro


haakon-e:
I've really enjoyed `broadcast_dims` since discovering it a few months ago.

Using it a lot, I've found that for some operations, the `broadcast_dims` notation can be a bit obfuscating, and that macros can be used to simplify code that may previously have been relatively complex.

If there's appetite for something like this, I'll open a PR, with a few tests and docs when I have time.

A few examples:
```julia
xs, ys, zs = (X(-180:2.0:-150), Y(-30:2.0:-10), Z(-5000:100.0:0))

a = rand(xs, ys, zs)

d = rand(ys, zs)

# currently, I have to do:
broadcast_dims(/, a, d)

# with my proposal, below, I can do:
@broadcast_dims a / d

# it's implemented to be arbitrarily chained, with certain calls left untouched, e.g.
@macroexpand @broadcast_dims a + .√a / (sqrt.(d) * a.^2)
# expands to:
:(broadcast_dims(+, a, broadcast_dims(/, .√(a), broadcast_dims(*, sqrt.(d), a .^ 2))))
```

<details> <summary> implementation sketch </summary>

```julia
using MacroTools: @capture  # light dependency: https://github.com/FluxML/MacroTools.jl/blob/master/Project.toml
export @broadcast_dims
macro broadcast_dims(expr)
    function process_expr(ex)
        if @capture(ex, f_(a_)) || @capture(ex, a_.b_) || @capture(ex, f_.(args__))
            # don't touch: (1) single arg functions `f(a)`, (2) field access `a.b`, (3) function broadcast `f.(a, b, c)`
            return esc(ex)
        elseif @capture(ex, f_(args__))
            string(f)[1] == '.' && return esc(ex)  # don't broadcast a call like `.√a`
            # recursively check each argument for sub-expressions. 
            args2 = map(process_expr, args)
            return :(broadcast_dims($(esc(f)), $(args2...)))
        else
            # don't touch: anything else
            return esc(ex)
        end
    end

    return process_expr(expr)
end
```
PS: I'm pretty new to macros, but I hope this has decently good "hygiene"

</details>

haakon-e:
ahhh lol I see you might have actually done this in #777

rafaqz:
Yes! It's working well, try it out. There are some examples in the tests.

The trick to the macro is getting every argument into a separate variable so we can extract the dimensions, then permuting them to match before they are used in the broadcast.

[...]

PR #794, opened 2024-09-01
`@d` macro for dimension aware broadcasts

rafaqz:
This PR adds the `@d` macro to transform regular broadcast syntax into dimensional syntax that will permute and reshape dimensions like `broadcast_dims`. It's more powerful than broadcast dims because it works across any complicated nested fused broadcasts you can do in base Julia. 

@haakon-e maybe you want to review this? The code is not too different from the example you posted in #791

### DIFF
--- a/docs/src/api/reference.md
+++ b/docs/src/api/reference.md
@@ -95,10 +95,10 @@ You can turn check the current setting and turn them on or off with these
 methods.
 
 ```@docs
-strict_broadcast
-strict_broadcast!
-strict_matmul
-strict_matmul!
+DimensionalData.strict_broadcast
+DimensionalData.strict_broadcast!
+DimensionalData.strict_matmul
+DimensionalData.strict_matmul!
 ```
 
 Base methods
--- a/docs/src/api/reference.md
+++ b/docs/src/api/reference.md
@@ -76,6 +76,7 @@ For transforming DimensionalData objects:
 set
 rebuild
 modify
+@d
 broadcast_dims
 broadcast_dims!
 mergedims
--- a/docs/src/api/reference.md
+++ b/docs/src/api/reference.md
@@ -84,6 +84,23 @@ unmergedims
 reorder
 ```
 
+# Global lookup strictness settings
+
+Control how strict DimensionalData when comparing [`Lookup`](@ref)s
+before doing broadcasts and matrix multipications.
+
+In some cases (especially `DimVector` and small `DimArray`) checking 
+lookup values match may be too costly compared to the operations.
+You can turn check the current setting and turn them on or off with these
+methods.
+
+```@docs
+strict_broadcast
+strict_broadcast!
+strict_matmul
+strict_matmul!
+```
+
 Base methods
 
 ```@docs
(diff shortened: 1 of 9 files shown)


haakon-e (commented):
Thanks for the tag, it looks like you've put some great effort into this! I like that you thought of 

I will try to go through the code in more detail as I understand it more. For now, I have a few broader comments.

* The macro name `@d` is very short, which can make it convenient to type out. That said, it would be useful to know if other large repositories define a macro with this name, so as to avoid name conflicts. It may also be useful to get feedback from the community on the name (e.g. through slack posts in #geo, #climate, #helpdesk, etc.)
    - I also note that the name is exported, which may increase potential name conflicts.

* As a user, I immediately notice the slight annoyance of having to type out broadcasting dots (e.g. `f.(da1, da2) .+ da3`) in any expression that makes use of `@d`. The code in #791 did not require this.
    - Upon experimenting, (and using `da1, da2, da3, f` from the tests), I see that I:
        - Cannot do: `@d @. da3 * f(da1 / 1, da2)`
        - Can do: `@. @d da3 * f(da1 / 1, da2)`
        - Can do `@. @d da3 * f(da1 / 1, da2) dims = (Z,Y,X)` but `dims` is silently ignored
    - Do you have an example of a scenario where you wouldn't want to broadcast every expression after `@d`?
        - Given that might indeed be useful, perhaps we could follow `@.`'s convention for excluding dots (their docs):
        `If you want to avoid adding dots for selected function calls in expr, splice those function calls in with $. For example, @. sqrt(abs($sort(x))) is equivalent to sqrt.(abs.(sort(x))) (no dot for sort).`

* Before this PR is merged, it would be useful for the new methods to have some more comments and/or docstrings to explain what's going on, which would hopefully improve future maintainability.

* Since this constructs a new `DimArray`, it would be very useful if attributes like `name` can be passed alongside the output `dims`.
    - On this note, it seems that incorrect arguments (like `name`) is currently silently ignored.

Thanks again for this great PR! I'll get back with more comments later as I look at it in more detail.

rafaqz:
Thanks, those are really good observations. I'll make sure passing name and metadata work too. Maybe all unidentified keywords should just go to `rebuild` so for example in Rasters we can pass a new `missingval` and it just works.

I haven't seen `@d` around before, and I think DD is big enough to just claim it these days. Anything longer will be a little verbose for something so common.

I don't think the `@.` macro can be the default behaviour, as sometimes you want to isolate parts that don't broadcast. But we should try and make all of those combinations work. I think some might be hard because what `@.` actually does is pretty weird, as far as I remember. 

I personally never used `@.` but I really wanted to have `@d.` syntax for people that do. Of course it's not parseable as a single macro name.

rafaqz:
Ok @haakon-e this now works with `@.` as the _inner_ macro like `@d @. a * b`. Using it as the outer macro is actually skipping the `@d` mechanisms because it cant find any broadcast calls to wrap. I think it just happens to work anyway in your example.

And all keywords that any DimArray `rebuild` method accepts should also be accepted. `dims` now also sets the lookups values using `set`, so you can change the lookup in the broadcast if you want to.

Also added some comments.

rafaqz:
Ok, gotta merge, other PRs incoming for the breaking version

PR #795, opened 2024-09-02
Make expand_dimensions definitions non ambiguous

asinghvi17:
cc @lazarusA

### DIFF
--- a/ext/DimensionalDataMakie.jl
+++ b/ext/DimensionalDataMakie.jl
@@ -380,6 +380,7 @@ Makie.plottype(::AbstractDimArray{<:Any,3}) = Makie.Volume
 # Makie.expand_dimensions(::Makie.PointBased, y::IntervalSets.AbstractInterval) = (keys(y), y)
 
 # Conversions
+# Generic conversion for arbitrary recipes that don't define a conversion trait
 function Makie.convert_arguments(t::Type{<:Makie.AbstractPlot}, A::AbstractDimMatrix)
     A1 = _prepare_for_makie(A)
     tr = Makie.conversion_trait(t, A)
@@ -390,6 +391,7 @@ function Makie.convert_arguments(t::Type{<:Makie.AbstractPlot}, A::AbstractDimMa
     end
     return xs, ys, last(Makie.convert_arguments(t, parent(A1)))
 end
+# PointBased conversions (scatter, lines, poly, etc)
 function Makie.convert_arguments(t::Makie.PointBased, A::AbstractDimVector)
     A1 = _prepare_for_makie(A)
     xs = parent(lookup(A, 1))
@@ -442,7 +444,7 @@ function Makie.convert_arguments(t::Type{Plot{Makie.volumeslices}}, A::AbstractD
 end
 # # fallbacks with descriptive error messages
 function Makie.convert_arguments(t::Makie.ConversionTrait, A::AbstractDimArray{<:Any,N}) where {N}
-    @warn "$t not implemented for `AbstractDimArray` with $N dims, falling back to parent array type"
+    @warn "Conversion trait $t not implemented for `AbstractDimArray` with $N dims, falling back to parent array type"
     return Makie.convert_arguments(t, parent(A))
 end
 
@@ -488,6 +490,18 @@ function _categorical_or_dependent(A, ::Nothing)
     end
 end
 
+# Check for regular sampling on a lookup, throw an error if not.
+function _check_regular_sampling(l; axis = nothing)
+    if !DD.isregular(l)
+        throw(ArgumentError("""
+        DimensionalDataMakie: The $(isnothing(axis) ? "" : "$axis-axis ")lookup is not regularly spaced, which is required for image-like plot types in Makie.
+        The lookup was:
+        $l
+
+        You can solve this by resampling your raster, or by using a more permissive plot type like `surface`, `contour`, or `contourf`. 
+        """))
+    end
+end
 
 # Simplify dimension lookups and move information to axis attributes
 _split_attributes(A) = _split_attributes(dims(A))
--- a/ext/DimensionalDataMakie.jl
+++ b/ext/DimensionalDataMakie.jl
@@ -138,11 +138,17 @@ for (f1, f2) in _paired(:plot => :heatmap, :heatmap, :image, :contour, :contourf
         ) where T
             replacements = _keywords2dimpairs(x, y)
             A1, A2, args, merged_attributes = _surface2(A, $f2, attributes, replacements)
-            p = if $(f1 == :surface)
+            axis_type = if haskey(merged_attributes, :axis) && haskey(merged_attributes.axis, :type)
+                to_value(type)
+            else
+                Makie.args_preferred_axis(Makie.Plot{$f2}, args...)
+            end
+
+            p = if axis_type <: Union{LScene, Makie.PolarAxis} # TODO: check `Makie.args_preferred_axis` here!
                 # surface is an LScene so we cant pass attributes
                 p = Makie.$f2(args...; attributes...)
                 # And instead set axisnames manually
-                if !isnothing(p.axis.scene[OldAxis])
+                if p.axis isa LScene && !isnothing(p.axis.scene[OldAxis])
                     p.axis.scene[OldAxis][:names, :axisnames] = map(DD.label, DD.dims(A2))
                 end
                 p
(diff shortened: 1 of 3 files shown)


ext/DimensionalDataMakie.jl
```diff
@@ -432,8 +432,13 @@ end
     # These can just forward to the relevant converts.
     Makie.expand_dimensions(t::Makie.PointBased, A::AbstractDimVector) = Makie.convert_arguments(t, A)
     Makie.expand_dimensions(t::Makie.PointBased, A::AbstractDimMatrix) = Makie.convert_arguments(t, A)
-    Makie.expand_dimensions(t::SurfaceLikeCompat, A::AbstractDimMatrix) = Makie.convert_arguments(t, A)
-    Makie.expand_dimensions(t::Makie.CellGrid, A::AbstractDimMatrix) = Makie.convert_arguments(t, A)
```
lazarusA, line 436: as I tested on https://github.com/rafaqz/Rasters.jl/pull/723, it looks like this is not enough to cover the uses cases there.

asinghvi17: No, that's a different issue - the conversions for vertexgrid have to be different than those for cellgrid, and I don't think those are implemented in here yet.  Will look into it.

rafaqz: Can we just delete everything in Rasters? It's an AbstractDimArray anyway...

asinghvi17: Yeah I guess the code is pretty old :D 

The only thing that's really needed is plot I think

asinghvi17: Ah shoot, we can't drop Rasters' Makie because it has the methods to populate missings and stuff.  It's already pretty generic though and everything _should_ be forwarded to DimensionalData...

rafaqz: Ohh right, makes sense. Can we just replace missings there and rewrap as `DimArray`? We will still need all the same dispatch methods to avoid ambiguities.

We also need the dispatch for handling rotated projections and other weird things (or maybe GeoMakie can deal with that?)

asinghvi17: GeoMakie can deal with that to an extent but if you're plotting on a regular axis you still need that.

rafaqz: Yeah, but maybe we need a GeoMakie extension to just plot rotations and similar directly rather than resampling them first

asinghvi17:
I am now routing all `convert_arguments` through the conversion trait for the overridden functions.  Since we don't dispatch on raw plot types for these in convert_arguments anyway, this seems reasonable to me.

This also gets rid of plots not converting correctly.


The currently outstanding issue with this approach, is that updating plots created in this way isn't possible with traditional updating syntax (`plt[1] = new_dimarray`).  You have to pass in an Observable or do the conversion and unwrap it yourself. Both of which are bad ideas.  That's next on the list but would involve cleaning up the whole recipe system anyway.

ext/DimensionalDataMakie.jl
asinghvi17: We could!  It'd add a lot more complexity though, and I wouldn't trust that with e.g. 2d to 3d transformations.

asinghvi17:
Ah, categorical lookups have to be handled like regular ones.

ext/DimensionalDataMakie.jl
```diff
@@ -561,7 +600,18 @@ function _lookup_to_vector(l)
     end
 end
 
+function _lookup_to_vertex_vector(l)
+    if isintervals(l)
+        bs = intervalbounds(l)
+        return @. first(bs) + (last(bs) - first(bs)) / 2
```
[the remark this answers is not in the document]
asinghvi17: Is it?  If so that would be great.   Does shiftlocus also handle converting point sampling to interval sampling?

ext/DimensionalDataMakie.jl
```diff
@@ -464,6 +491,18 @@ function _categorical_or_dependent(A, ::Nothing)
...
 
+# Check for regular sampling on a lookup, throw an error if not.
+function _check_regular_sampling(l; axis = nothing)
+    if !DD.isregular(l)
+        throw(ArgumentError("""
+        DimensionalDataMakie: The $(isnothing(axis) ? "" : "$axis-axis ")lookup is not regularly spaced, which is required for image-like plot types in Makie.
+        The lookup was:
+        $l
+
+        You can solve this by resampling your raster, or by using a more permissive plot type like `surface`, `contour`, or `contourf`.
```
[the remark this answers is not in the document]
asinghvi17: I'm not sure if the GL backends will but let me check

ext/DimensionalDataMakie.jl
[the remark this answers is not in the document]
rafaqz: No you have to do that separatly: `shiftlocus(Start(), set(lookup, Intervals()))`. Points will always turn into `Intervals(Center())` by default.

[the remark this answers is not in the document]
rafaqz: Its good to always use the methods like `shiftlocus` for things like this so we can centralise handling of the edge cases.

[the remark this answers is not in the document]
lazarusA: ok people. I just tested this one in combination with https://github.com/rafaqz/Rasters.jl/pull/723, and things seems to work really well. So, I would like to merge this one, so that we can also fix `Rasters#main`.

lazarusA (approved):
thanks a lot @asinghvi17 for bringing this up to the finish line. Great work!

ext/DimensionalDataMakie.jl
```diff
@@ -464,6 +491,19 @@ function _categorical_or_dependent(A, ::Nothing)
...
+# Check for regular sampling on a lookup, throw an error if not.
+# Here, we assume 
+function _check_regular_or_categorical_sampling(l; axis = nothing)
+    if !(DD.isregular(l) || DD.iscategorical(l))
+        throw(ArgumentError("""
+        DimensionalDataMakie: The $(isnothing(axis) ? "" : "$axis-axis ")lookup is not regularly spaced, which is required for image-like plot types in Makie.
+        The lookup was:
+        $l
+
+        You can solve this by resampling your raster, or by using a more permissive plot type like `heatmap`, `surface`, `contour`, or `contourf`.
```
lazarusA, line 503: @rafaqz I just hit this edge case for raster's volume-like cases. what would be the way to get a regular-spaced raster? a better suggestion is needed here.  Maybe one-liner to resample a given raster `A`?

asinghvi17: Ah yeah a one liner would be good to add there.  I should also add an option to customize the error message so this can be used in volume plots...

asinghvi17:
TODOs on my end are:

- [ ] Special-case `axis` and `figure` attributes so they DON'T get converted to Observables.
- [ ] Better and more configurable error messages when checking regular lookups
- [ ] Use shiftlocus and set when converting lookups to vectors.

ext/DimensionalDataMakie.jl
```diff
@@ -561,7 +601,18 @@ function _lookup_to_vector(l)
     end
 end
 
+function _lookup_to_vertex_vector(l)
+    if isintervals(l)
+        bs = intervalbounds(l)
+        return @. first(bs) + (last(bs) - first(bs)) / 2
```
rafaqz, line 607:
```suggestion
        return parent(shiftlocus(Center(), l))
```

ext/DimensionalDataMakie.jl
```diff
@@ -561,7 +601,18 @@ function _lookup_to_vector(l)
     end
 end
 
+function _lookup_to_vertex_vector(l)
+    if isintervals(l)
+        bs = intervalbounds(l)
```
rafaqz, line 606:
```suggestion
```

lazarusA:
something went wrong with the latest `axis`, `args` removal.

asinghvi17:
OK things look good to me.  @rafaqz feel free to merge whenever.

ext/DimensionalDataMakie.jl
rafaqz: With `image` or `heatmap`? `heatmap` should wirk Maybe we can just warn for `image` and plot anyway?

rafaqz (approved):

ext/DimensionalDataMakie.jl
lazarusA: the issue I found was for something with `X, Y, Ti`, creating a volume, where the axis for Ti is usually irregular, hence failing afterwards to plot, I just want say, treat this as a regular thing as well, it makes sense for daily data, but I'm sure there should be others where we should not do that.

asinghvi17: Aha ok!  Yeah I guess it's a bit restrictive to error all the time.

rafaqz:
Are the `spy` fixes good for merging now too?

asinghvi17:
Yes, I'd say it's better than nothing.  We can only fix the remaining issues on the Makie end.

PR #796, opened 2024-09-03
Update README.md formatting

asinghvi17:
Makes it a bit clearer IMO, but feel free to close/modify if it doesn't end up doing that.


lazarusA (approved):
yes, it seems like there is always room for improvement :D.

Issue #820, opened 2024-10-15
[FR] Support generator expressions for construction


kapple19:
I'd like the syntax:

```
F = DimArray(f(x, y, z) for x in x_vec, y in y_vec, z in z_vec)
```

which will parse that `x`, `y`, and `z` are dimension names.

```
F[x = 3, y = 7, z = -2]
```

This avoids me having to make the intermediate computation, such as

```
F = DimArray(
    [f(x, y, z) for x in x_vec, y in y_vec, z in z_vec],
    (x = x_vec, y = y_vec, z = z_vec)
)
```

which contains a lot of redundant syntax.

rafaqz:
How about `@d f.(X(xvec), Y(yvec), Z(zvec))`? 

That already works and should do the same thing

rafaqz:
What should also be able to work pretty easily is this:
```julia
F = DimArray(f(x, y, z) for x in X(x_vec), y in Y(y_vec), z in Z(z_vec))
```
And this works already for a single dimension:
```julia
julia> DimArray(identity, X(1:5))
╭─────────────────────────────────────────╮
│ 5-element DimArray{Int64,1} identity(X) │
├─────────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────── dims ┐
  ↓ X Sampled{Int64} 1:5 ForwardOrdered Regular Points
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 1  1
 2  2
 3  3
 4  4
 5  5

```

We should let it take multiple arguments.

rafaqz:
Ok this is currently the best option:
```julia
f = sum
julia> f.(DimPoints((X(1:2), Y(1:3), Z(1:4))))
╭─────────────────────────╮
│ 2×3×4 DimArray{Int64,3} │
├─────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────── dims ┐
  ↓ X Sampled{Int64} 1:2 ForwardOrdered Regular Points,
  → Y Sampled{Int64} 1:3 ForwardOrdered Regular Points,
  ↗ Z Sampled{Int64} 1:4 ForwardOrdered Regular Points
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
[:, :, 1]
 ↓ →  1  2  3
 1    3  4  5
 2    4  5  6

```

kapple19:
[...]

Issue #821, opened 2024-10-15
[FR] AlgebraOfGraphics.jl support


kapple19:
For a `v::DimArray`, it would be nice for `AlgebraOfGraphics.data(v)` to work.
`AlgebraOfGraphics.mapping(X, Y)` as well, of course,
and for arbitrary dimension names, e.g. `AlgebraOfGraphics.mapping(:a, :b)`.

rafaqz:
Sounds good.

Personally I use raw Makie.jl so I'm not really across what AOG adds! We should make basic Makie recipes do as much as possible, but if there are extra methods we can add to help AOG I'm happy for someone to write an extension.

Can you maybe explain a bit more what these methods do now and what you would like them to do with a DimArray?

asinghvi17:
Does mapping with symbols not work already?

kapple19:
> Can you maybe explain a bit more what these methods do now and what you would like them to do with a DimArray?

> Does mapping with symbols not work already?

Here's some examples:

```
using DimensionalData
using CairoMakie, AlgebraOfGraphics

P = Dim{:p}

v = DimArray(
    rand(21, 5, 4),
    (X, Y, P);
    name = :RandomData
)

dat = data(v)
scat = vis * visual(Scatter)
line = vis * visual(Series)
heat = dat * visual(Heatmap)

scat * mapping(:x, Y, color = :p) |> draw
scat * mapping(X, Y, color = P) |> draw
scat * mapping(X, P, color = Y) |> draw
line * mapping(X, Y, color = P) |> draw
scat * mapping(X, Y, layout = P) |> draw
scat * mapping(X, Y, row = P) |> draw
scat * mapping(X, Y, col = P) |> draw
heat * mapping(X, Y, layout = P) |> draw
heat * mapping(P, X, layout = Y) |> draw
```

I'll update this post with how the above turns out, just waiting for Makie to compile 😅.

asinghvi17:
It would have to be the exact symbol, so `:X`

kapple19:
> It would have to be the exact symbol, so `:X`

Oh, you're right 😅.

However,

```
scat * mapping(:X, :Y, color = :p) |> draw
```

![image](https://github.com/user-attachments/assets/b45b1797-a3e1-4c3c-a370-bf58a9091d0e)

rafaqz:
> DimArrays aren't long-formatted tabular data

[...]

PR #822, opened 2024-10-15
Maybe Breaking (?): add missing IteratorSize method for dimensions

rafaqz:
Tiny fix so that comprehensions over dimensions return a DimArray. @kapple19 I think this might be even nicer than what you were looking for.

```julia
julia> [(x, y) for x in X(1:5), y in Y(1:3)]
╭─────────────────────────────────────╮
│ 5×3 DimArray{Tuple{Int64, Int64},2} │
├─────────────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────── dims ┐
  ↓ X Sampled{Int64} 1:5 ForwardOrdered Regular Points,
  → Y Sampled{Int64} 1:3 ForwardOrdered Regular Points
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
 ↓ →  1        2        3
 1     (1, 1)   (1, 2)   (1, 3)
 2     (2, 1)   (2, 2)   (2, 3)
 3     (3, 1)   (3, 2)   (3, 3)
 4     (4, 1)   (4, 2)   (4, 3)
 5     (5, 1)   (5, 2)   (5, 3)
```
~~Needs a test.~~ See #820

### DIFF
--- a/src/Dimensions/dimension.jl
+++ b/src/Dimensions/dimension.jl
@@ -286,6 +286,7 @@ Base.first(d::Dimension) = val(d)
 Base.first(d::Dimension{<:AbstractArray}) = first(lookup(d))
 Base.last(d::Dimension) = val(d)
 Base.last(d::Dimension{<:AbstractArray}) = last(lookup(d))
+Base.IteratorSize(d::Dimension{<:AbstractArray}) = Base.IteratorSize(parent(d))
 Base.firstindex(d::Dimension) = 1
 Base.lastindex(d::Dimension) = 1
 Base.firstindex(d::Dimension{<:AbstractArray}) = firstindex(lookup(d))
--- a/test/array.jl
+++ b/test/array.jl
@@ -456,6 +456,11 @@ end
     @test_throws ArgumentError fill(5.0, (X(:e), Y(8)))
 end
 
+
+@testset "generator constructor" begin
+    [(x, y) for x in X(10:10:50), y in Y(0.0:0.1:1.0)]
+end
+
 @testset "ones, zeros, trues, falses constructors" begin
     da = @inferred zeros(X(4), Y(40.0:10.0:80.0; order=ForwardOrdered()); metadata=(a=1, b=2))
     @test eltype(da) <: Float64


kapple19:
That's awesome! Does it work for arbitrary dimensions as per my example?

rafaqz:
Yeah, should work for anything you put in there. It could have worked for years now if I remembered to add that method :sweat_smile:

rafaqz:
This could be considered breaking as its changing the return value of a _lot_ of things

kapple19:
Just to be clear, will the following syntax work?

```
F = DimArray(f(a, b, c) for a in a_vec, b in b_vec, c in c_vec)
```

or do I need to declare my arbitrary `Dim`s?

rafaqz:
Sorry theres not really any way for that to work as is! Its too ambiguous, it would need a macro and those a best saved for when they're really needed. Wrapping with dimensions is what makes all the magic here work because we then own its methods that control the outcomes of a lot of Base julia functions.

So you wont need the `DimArray`, but you need these clunky wrappers. 

```julia
F = [f(a, b, c) for a in Dim{:a}(a_vec), b in Dim{:b}(b_vec), c in Dim{:c}(c_vec)]
```

I always wonder if we should bite the bullet and export `D` as a shortcut to `Dim`, so it would be:

```julia
F = [f(a, b, c) for a in D{:a}(a_vec), b in D{:b}(b_vec), c in D{:c}(c_vec)]
```

Of course you can also do:
```julia
@dim A
@dim B
@dim C
F = [f(a, b, c) for a in A(a_vec), b in B(b_vec), c in C(c_vec)]
```

kapple19:
> Sorry theres not really any way for that to work as is!

That's okay! I understand why, but the existing (prospective) syntax is fine for me since it means I don't have to define multiple intermediate stages, and can instantiate a `DimArray` in one nice swoop.

I can't seem to download the branch, but I'm in no rush, and am looking forward to the merge! Thanks heaps!

rafaqz:
Check your network connection, your just not getting through to github ;)

kapple19:
> I always wonder if we should bite the bullet and export `D` as a shortcut to `Dim`

I am of the opinion that, no. People can just `using DimensionalData: Dim as D`.

Personally I do a lot of `using ... as ...` because I'm pedantic about names. E.g. I always do

```
using DimensionalData: dims as getdims
```

because I like my functions to be verbal, so I would

```
dims = getdims(v::DimArray)
```

rafaqz:
> `dims = getdims(v::DimArray)`

:laughing: yes I clearly have an aversion to putting  `get` on anything, something about it reminds me of OOP days, or its just not aesthetically pleasing. But it would solve a lot, I constantly need to do `DD.dims` inside functions because `dims` is such a common keyword name. But `dims(x)` just looks so much nicer !

rafaqz:
Ohh someone is firewalling something for you! how lucky. I guess `using DimensionalData` is not actually pining github, but julia specific servers

kapple19:
> yes I clearly have an aversion to putting `get` on anything, something about it reminds me of OOP days, or its just not aesthetically pleasing.

I get it. I have similar traumas with MATLAB and its version of OOP. From their case style naming conventions to... everything else.

And don't stress about it, I'm happy to `dims as getdims` 😊.

asinghvi17:
Github was a bit spotty earlier, that might have been the source of your issues.  It should be fine now though so you could try adding the branch again @kapple19

kapple19:
What about the following syntax?

```
V = DimArray(f(a, b) for a in Dim{:a}(1:10), b in Dim{:b}(1:5); name = :V)
```

Is it possible to make that work? The Syntax errors, so is available 👀🥺

kapple19:
Ah I see.

The `DimArray` accepting a generator would, as per my example in the other thread (which I should close to uniquely scope to this thread), bypass the need to construct the `DimArray` with a generator and then `remake` it to assign a `name` value. Especially in motivation for [interop with AlgebraOfGraphics](https://github.com/rafaqz/DimensionalData.jl/issues/821).

rafaqz:
If you use a square brackets generator it will work. But yes we could also accept an uncollected generator.

PR #823, opened 2024-10-15
Allow dimensions to be used as AlgebraOfGraphics selectors

### LINKED ISSUE #821: [FR] AlgebraOfGraphics.jl support

asinghvi17:
This adds AoG selector support to DimensionalData dims, so that things like this just work:

```julia
A3 = DimArray(rand(21, 5, 4), (X, Y, Dim{:p}); name = :RandomData)
data(A3) * visual(Heatmap) * mapping(X, :RandomData, Dim{:p}, layout = Y => nonnumeric) |> draw

# before this, you would have to use and know the specific symbols that the dims are:
data(A3) * visual(Heatmap) * mapping(:X, :RandomData,:p, layout = :Y => nonnumeric) |> draw
```

They actually work anywhere by getting the dimension symbol from `DD.name`, so you can run these on `data(DataFrame(dimtable))` with `X` as well!

Fix #821

### DIFF
--- a/docs/src/plots.md
+++ b/docs/src/plots.md
@@ -21,6 +21,7 @@ and axes in the right places:
 
 ```@example Makie
 using DimensionalData, CairoMakie
+CairoMakie.activate!(type = :svg) # hide
 
 A = rand(X(10:10:100), Y([:a, :b, :c]))
 Makie.plot(A; colormap=:inferno)
@@ -32,13 +33,34 @@ favours the categorical variable for the X axis:
 ```@example Makie
 Makie.rainclouds(A)
 ```
+
+## AlgebraOfGraphics.jl
+
+AlgebraOfGraphics.jl is a high-level plotting library built on top of Makie.jl that provides a declarative "grammar of graphics" for creating complex visualizations. It allows you to construct plots using algebraic operations like `*` and `+`, making it easy to create sophisticated graphics with minimal code.
+
+Any `DimensionalArray` is also a `Tables.jl` table, so it can be used with `AlgebraOfGraphics.jl` directly.  You can indicate columns in `mapping` with Symbols directly (like `:X` or `:Y`), **or** you can use the `Dim` type directly (like `X` or `Y`)! 
+
+```@example Makie
+using DimensionalData, AlgebraOfGraphics, CairoMakie
+CairoMakie.activate!(type = :svg) # hide
+
+A = DimArray(rand(10, 10), (X(1:10), Y(1:10)), name = :data)
+
+data(A) * mapping(X, Y; color = :data) * visual(Scatter) |> draw
+```
+
+
+
+
 ## Test series plots
 
 ### default colormap
+
 ```@example Makie
 B = rand(X(10:10:100), Y([:a, :b, :c, :d, :e, :f, :g, :h, :i, :j]))
 Makie.series(B)
 ```
+
 ### A different colormap
 The colormap is controlled by the `color` argument, which can take as an input a named colormap, i.e. `:plasma` or a list of colours. 
 
--- a/docs/src/plots.md
+++ b/docs/src/plots.md
@@ -36,9 +36,15 @@ Makie.rainclouds(A)
 
 ## AlgebraOfGraphics.jl
 
-AlgebraOfGraphics.jl is a high-level plotting library built on top of Makie.jl that provides a declarative "grammar of graphics" for creating complex visualizations. It allows you to construct plots using algebraic operations like `*` and `+`, making it easy to create sophisticated graphics with minimal code.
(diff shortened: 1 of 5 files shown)


rafaqz:
Can you add a before/after for that example? Just to know what we gain from the extension

asinghvi17:
Before you would just have to put a symbol.  But here you can put the dimension type directly.  You can also select dimensions with e.g. `DD.XDim` if you're not sure what the name is, since on a DimTable this actually queries dimensions.

So something like `data(some_dimarray_lat_lon_axes) * mapping(XDim, YDim. color = :RandomData) * visual(Scatter)` just works.  Even if the dim names are Lon and Lat, if they're subtypes of XDim and YDim the selectors will do the job.

ext/DimensionalDataAlgebraOfGraphicsExt.jl
```diff
@@ -0,0 +1,33 @@
...
+This extension allows DimensionalData `Dimension` types to be used
+as column selectors for AlgebraOfGraphics.jl.
+
+Specifically, this implements the `AoG.select` method for `Columns`
+objects, which is the type that `AlgebraOfGraphics.data` returns.
+
+=#
+
+# The generic selector, to enable this to work even in `DataFrame(dimarray)`
+function AoG.select(data::AoG.Columns, dim::Type{<: DD.Dimensions.Dimension})
```
rafaqz, line 17:
```suggestion
function AoG.select(data::AoG.Columns, dim::DD.Dimensions.DimOrDimType)
```

ext/DimensionalDataAlgebraOfGraphicsExt.jl
```diff
@@ -0,0 +1,33 @@
...
+    name = DD.name(dim)
+    v = AoG.getcolumn(data.columns, Symbol(name))
+    return (v,) => identity => AoG.to_string(name) => nothing
+end
+
+# The specific selector for `DimTable`s.
+# This searches the dimensions of the dimtable for the appropriate dimension,
+# so that e.g. `X` also applies to any `XDim`, and so forth.
+function AoG.select(data::AoG.Columns{<: DD.DimTable}, dim::Type{<: DD.Dimensions.Dimension})
```
rafaqz, line 26:
```suggestion
function AoG.select(data::AoG.Columns{<: DD.DimTable}, dim::DD.Dimensions.DimOrDimType)
```

rafaqz (changes requested):
Nice to allow constructed Dims in too

asinghvi17:
Ah didn't know that was a thing!  Will add it.

docs/src/plots.md
```diff
@@ -32,13 +33,34 @@ favours the categorical variable for the X axis:
 ```@example Makie
 Makie.rainclouds(A)
 ```
+
+## AlgebraOfGraphics.jl
+
+AlgebraOfGraphics.jl is a high-level plotting library built on top of Makie.jl that provides a declarative "grammar of graphics" for creating complex visualizations. It allows you to construct plots using algebraic operations like `*` and `+`, making it easy to create sophisticated graphics with minimal code.
```
lazarusA, line 39: I now that the wording is mainly inspired by ggplot (the grammar, part).  However, when reading https://aog.makie.org/dev/philosophy/ we see that here we are actually doing `Algebra` hence the name of the package .  Maybe rephrase that to something like: `similar to the grammar of graphics in ggplot` or something, without claiming that this one here is `grammar of graphics`.

lazarusA: something I could help with here? docs build correctly... so...

asinghvi17: Thanks for the ping, just addressed this and added some more examples.

test/plotrecipes.jl
```diff
@@ -378,3 +378,33 @@ end
...
+    A1 = rand(X(1:5); name=:test)
+    A1c = rand(X('a':'e'); name=:test)
+
+    @testset "1d, symbol indexing" begin
+        @test_nowarn data(A1) * mapping(:X, :test) * visual(CairoMakie.Lines) |> draw
+        @test_nowarn data(A1c) * mapping(:X, :test) * visual(CairoMakie.Lines) |> draw
+    end
+
+    @testset "1d, dim indexing" begin
+        @test_nowarn data(A1) * mapping(X, Dim{:test}) * visual(CairoMakie.Lines) |> draw
```
asinghvi17, line 396: The test error here is because I'm passing in `Dim{:test}`.  If I don't, this is fine. 

Do we care about this syntax?  I almost want to remove it, since the data name isn't technically a dimension of the array.  At the same time, for an n-dimensional DimArray, the data is the n+1th dimension...

rafaqz: That shouldn't work right? It's not a dimension

lazarusA (approved):
this is going to be so useful, thanks a lot!

rafaqz (approved):

rafaqz:
Thanks!!!!

PR #852, opened 2024-11-07
Add generator syntax as DimArray method

kapple19:
Instead of `rebuild`ing an array comprehension that generates a `DimArray` to e.g. `name` it, this allows the cleaner syntax:

```
julia> DimArray((x, y) for x in X(10:10:50), y in Y(0.0:0.1:1.0); name = :Value)
╭──────────────────────────────────────────────╮
│ 5×11 DimArray{Tuple{Int64, Float64},2} Value │
├──────────────────────────────────────────────┴───────────────────────────────────────────────────────── dims ┐
  ↓ X Sampled{Int64} 10:10:50 ForwardOrdered Regular Points,
  → Y Sampled{Float64} 0.0:0.1:1.0 ForwardOrdered Regular Points
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
  ↓ →  0.0         0.1         0.2         0.3         …  0.7         0.8         0.9         1.0
 10     (10, 0.0)   (10, 0.1)   (10, 0.2)   (10, 0.3)      (10, 0.7)   (10, 0.8)   (10, 0.9)   (10, 1.0)        
 20     (20, 0.0)   (20, 0.1)   (20, 0.2)   (20, 0.3)      (20, 0.7)   (20, 0.8)   (20, 0.9)   (20, 1.0)        
 30     (30, 0.0)   (30, 0.1)   (30, 0.2)   (30, 0.3)      (30, 0.7)   (30, 0.8)   (30, 0.9)   (30, 1.0)        
 40     (40, 0.0)   (40, 0.1)   (40, 0.2)   (40, 0.3)      (40, 0.7)   (40, 0.8)   (40, 0.9)   (40, 1.0)        
 50     (50, 0.0)   (50, 0.1)   (50, 0.2)   (50, 0.3)  …   (50, 0.7)   (50, 0.8)   (50, 0.9)   (50, 1.0)
```

I've added to the documentation that the semi-colon `;` is required to distinguish between the generator syntax dimensions and the keyword arguments.


kapple19:
Is the nightly error my fault?

asinghvi17:
I imagine not, it's failing on my PR as well.

docs/src/dimarrays.md
```diff
@@ -58,6 +58,13 @@ fill(7, X(5), Y(10))
 fill(7, X(5), Y(10); name=:fill, metadata=Dict())
 ```
 
+== generator expression
+
+```@ansi dimarray
+DimArray((x, y) for x in X(10:10:50), y in Y(0.0:0.1:1.0))
```
rafaqz, line 64:
```suggestion
((x, y) for x in X(10:10:50), y in Y(0.0:0.1:1.0))
```

This should work too?

rafaqz: Good to show the simplest version first

kapple19: Ah yes, good point.
I guess I was in the mindset of getting keyword arguments working with a generator constructor.

kapple19: Shall I add an array comprehension syntax example here too?

rafaqz: Oh yeah that's what I meant sorry

kapple19: It works with both actually. The square brackets creates an array, but the round brackets creates a generator.

That said, I should remove the round brackets then, I realise as I type this. 😅

test/array.jl
```diff
@@ -474,6 +474,12 @@ end
 
 @testset "generator constructor" begin
     [(x, y) for x in X(10:10:50), y in Y(0.0:0.1:1.0)]
```
rafaqz, line 476: For all of these we want to test the array size, that dims are actually X and Y etc,  and where a name is passed, test that it gets set.

rafaqz (changes requested):

test/array.jl
kapple19: Ah good point. I'll add tests tomorrow.

src/array/array.jl
```diff
@@ -467,6 +489,8 @@ function DimArray(f::Function, dim::Dimension; name=Symbol(nameof(f), "(", name(
      DimArray(f.(val(dim)), (dim,); name)
 end
 
+DimArray(itr::Base.Generator; kwargs...) = rebuild(collect(itr); kwargs...)
```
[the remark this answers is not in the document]
kapple19: Okay, tomorrow I'll test and work out any kinks tomorrow.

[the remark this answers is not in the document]
kapple19: The suggested change triggers:

```
ERROR: MethodError: similar(::Type{Matrix{Int64}}, ::Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}) is ambiguous.
```

when the following generator example is run

```
(x + y for x in 1:5, y in 1:3)
```

which is blocking development, and the previously working generator syntax method breaks with the same above error.

Do you know what's going on?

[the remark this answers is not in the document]
rafaqz: Ok we're doing some subtle type piracy in `similar` because of the unions. That's unfortunate. We need to rewrite those methods so there is always a Dimensions not in a union in at least one position, so we own the methods and there is no ambiguity. 

Let's make a new issue for that (just your comment text here basically) and fix it later.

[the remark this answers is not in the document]
kapple19: I'm happy to open that new issue if you want.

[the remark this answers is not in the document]
rafaqz: Thanks

PR #854, opened 2024-11-10
make array type printing consistent with base

simeonschaub:
`print_type` was missing a space inside the curly brackets. (My IDE
automatically removed trailing whitespace here, lmk if I should remove
those changes.)


rafaqz:
I think base added that space a few years ago? I kind don't like spaces in types but yeah, let's standardise.

You will need to run the doctests with `fix=true` to pass those tests

simeonschaub:
> I think base added that space a few years ago?

Yes, it was changed in https://github.com/JuliaLang/julia/pull/37085

PR #855, opened 2024-11-11
Respect `axis = (; type = ...)` when type is a LScene

asinghvi17:
Previously not respected.  See test for example.


rafaqz:
Just needs `Makie.LScene`

asinghvi17:
Should be fixed now, that was an oopsie in the tests :)

PR #864, opened 2024-11-18
Fix similar so that new memory is allocated

### LINKED ISSUE #863: similar is broken in `0.29.1`
It looks like the #844 broke similar. 
https://github.com/rafaqz/DimensionalData.jl/blob/910efa3c9dbe13661872f9ed64656aa2a7a693ed/src/array/array.jl#L143-L147

is missing a keyword argument for `data`, so the original array is being passed.

ptiede:
This fixes #863 by allocating new memory in similar.
I also have added a test to ensure this is checked in future versions.


rafaqz:
Good catch. We have far too many `similar` methods... But probably we need this check for every one

Issue #870, opened 2024-11-27
Declare that an `AbstractDimStack` is an `AbstractArray` with namedtuple eltype?


tiemvanderdeure:
After https://github.com/rafaqz/DimensionalData.jl/pull/811 merged, any `DimStack` is now basically an `AbstractArray` that iterates `NamedTuple`s. Should we declare it as such?

It could even be an `AbstractBasicDimArray`, but that would require some changes as a `AbstractDimStack` does not have its dimension in its type, currently.

It would allow us to remove a whole list of array-like methods, and make additional things work, e.g.:

```julia
A = [1.0 2.0 3.0; 4.0 5.0 6.0]
dimz = (X([:a, :b]), Y(10.0:10.0:30.0))
da1 = DimArray(A, dimz; name=:one)
da2 = DimArray(Float32.(2A), dimz; name=:two)
da3 = DimArray(Int.(3A), dimz; name=:three)

s = DimStack((da1, da2, da3))

broadcast!(s, s) do x
    (one = 1, two = 2, three = 3)
end
```

This now fails because `ndims` is not implemented for `DimStack` (an oversight, I assume), but if I just add
`
abstract type AbstractDimStack{K,T,N,L} <: AbstractArray{T,N} end
`
then it just works.

tiemvanderdeure:
I can't find that old PR but I think we're getting really close. It will be a bit of work cleaning up all the method ambiguities, though.

Maybe we could try to see how hard it is to get tests to pass?

rafaqz:
Yeah weird I cant either.  I'm sure I tried that.

But yes just adding the supertype and running tests would be interesting.

PR #871, opened 2024-11-27
Implement `AbstractArray` interface for `AbstractDimStack`

tiemvanderdeure:
See https://github.com/rafaqz/DimensionalData.jl/issues/870

### DIFF
--- a/src/stack/stack.jl
+++ b/src/stack/stack.jl
@@ -25,7 +25,7 @@ To extend `AbstractDimStack`, implement argument and keyword version of
 
 The constructor of an `AbstractDimStack` must accept a `NamedTuple`.
 """
-abstract type AbstractDimStack{K,T,N,L} end
+abstract type AbstractDimStack{K,T,N,L} <: AbstractArray{T,N} end
 const AbstractVectorDimStack = AbstractDimStack{K,T,1} where {K,T}
 const AbstractMatrixDimStack = AbstractDimStack{K,T,2} where {K,T}
 


rafaqz:
Lots of method ambiguities to work through. 

Possibly we can get even more out of this by making it an `AbstractBasicDimArray` ? They are like `AbstractDimArray` but don't necessarily have an AbstractArray parent.

tiemvanderdeure:
Yes I was already working on that. It mean I had to add the DimTuple to the type signature of `AbstractDimStack`, but I don't see why that should be a problem. This is breaking anyway.

tiemvanderdeure:
except for all the ambiguities the only tests that are broken are in `broadcast_dims`. Honestly it's much less bad than I expected

tiemvanderdeure:
Okay I actually managed to get rid of all ambiguities, I think. Most of it I could just delete, but I also explicitly specified that the eltype of a `AbstractDimArray` is a `NamedTuple`, because Aqua was complaining about ambiguous dispatches on `AbstractVector{<:AbstractArray}`. I don't know if there are every cases where this is not the case, though?

I ran some of the tests locally but let's see what they say.

rafaqz:
Yeah its always a `NamedTuple` so fine to limit it. 

One day we can make a `MathyNamedTuple` that `+` etc work on with a scalar. Like a mixed-type static array.

tiemvanderdeure:
I think this should work for the most part now, but would like to suggest one more breaking change.

Currently the way DimArrays are treated in `broadcast_dims` is that it broadcasts for each of the layers, or throws an error if multiple stacks are broadcasted but have different layers. This is inconsistent with the behaviour of a normal `broadcast`, and not at all what I would expect when I broadcast an `AbstractArray{<:NamedTuple}`. I think we should just change it to it broadcasting over `NamedTuple`s. I also think the current behaviour is undocumented.

rafaqz:
The problem is broadcasting over NamedTuple is a very inefficient and annoying way to e.g. multiply all layers of a stack by a scalar, especially if you have mixed size layers.

That change will probably break `mask` or something similar in Rasters.

tiemvanderdeure:
> The problem is broadcasting over NamedTuple is a very inefficient and annoying way to e.g. multiply all layers of a stack by a scalar,

Wouldn't that already require something like `maplayers(l -> l .* 2, s)`. Isn't that already preferred over `broadcast_dims(x -> x * 2, s)`?


> (There are reasons it's not AbstractArray ;)

I know, but I'm positively surprised that deleting a bunch of methods and some small tweaks made tests pass.

> Trying the Rasters tests may help clarify this

I'll give it a try and see how much of is broken on this branch

rafaqz:
Ok maybe a little clearer is multiplying by another raster that is a subset of the dimensions, often a bitmask.

broadcast_dims means the same code works for a Raster and a RasterStack but that will break with your approach.

We will at least need to provide a function that does that.

Maybe or a `bylayer` keyword is a solution?

tiemvanderdeure:
Okay with that example I see the problem more clearly. A keyword is an option but also makes a pretty basic and intuitive function more complicated

rafaqz:
Except `maplayers` won't work on a Raster. Maybe we make it so that just returns a Raster?

`bylayer` will also be a keyword elsewhere, so not the worst addition either

tiemvanderdeure:
Both those changes work well together, I think? So `maplayers` would take `AbstractDimStack`s and/or `AbstractDimArray`s. If there no stacks at all then that would just be `maplayers(f, A...) = f(A...)` and maybe throw a warning. And if it's mixed then we wrap the arrays. 

Then we can define `broadcast_dims` with `bylayer = true` to be just `maplayers(As -> broadcast_dims(f, As...), st).

rafaqz:
Why a warning?

tiemvanderdeure:
We don't have to, I was thinking it could be worth warning because the function doesn't do anything in that case. Anyway, I gave it a shot and you can tell me what you think about that implementation.

Multiple dispatch got complicated because `AbstractDimStack` is a subtype of `AbstractBasicDimArray`, so I just used if-else logic, which still should compile away.

rafaqz:
Right. We just don't want warnings because we'll actually use this on Rasters.

tiemvanderdeure:
Aargh what are these Plots segfaults/compilation errors. So annoying

tiemvanderdeure:
What do you think here @rafaqz? The only thing here that I ended up actually breaking at the user end is the behaviour of `broadcast_dims`, where you can get the old behaviour by setting `bylayer=true`. And I think this has some nice upsides like `vec(stack)` just works and `stack .= f(stack)` just works.

rafaqz:
There are still some conceptual problems, like `mean(stack)`

tiemvanderdeure:
> There are still some conceptual problems

But we should we try to figure those out? I'm sure you have a much better understanding than I do.

> like `mean(stack)`, `sum(stack)` is even worse for mixed size layers.

How so worse? The implementation didn't change at all so I think those should be the exact same?

rafaqz:
sum is currently for the layer. But as an AbstractArray some layers are repeated over missing dimensions so sum over the indexed values would be a much larger number. It's the same now but as our own object that's fine, it's just what it is.

But if we switch to AbstractArray we have to lose the mixed dimensionality in a lot of contexts to be consistent. And I'm not sure it's worth it. Instead we can implement broadcast on AbstractDimStack to be over all layers and always return a mixed size stack like the original. And other things like that

PR #876, opened 2024-12-03
Breaking: include DataType in CategoricalEltypes

tiemvanderdeure:
Does what it says on the tin :)

Specifically designed to make using RasterDataSources climate types as lookups (as in https://github.com/rafaqz/Rasters.jl/pull/826) even more slick.


rafaqz:
> Does what it says on the tin :)

Oh no it's spreading 😂

But yes, great idea. Technically breaking I guess as it's changing the returned type, so let's lump this with another break

Issue #878, opened 2024-12-03
dim-aware broadcast of `Sampled` and `Categorical` dim does not error but probably should


tiemvanderdeure:
MWE:
```julia
using DimensionalData
dv1 = rand(X(1:3))
dv2 = rand(X(Lookups.Categorical([1,2,3])))

bc = @d dv1 .* dv2
dims(bc)
```
returns an `X` dim with `NoLookup`. Probably this should not pass `comparedims` and just throw an error.

Multiplying a `Sampled` with a `Projected` always returns `Sampled` dims, I don't know if this is intentional?
```
dv3 = rand(X(Projected([1,2,3], crs = EPSG(4326))))
crs(@d dv3 .* dv1) # nothing
```

rafaqz:
Projected is `AbstractSampled` so it falls back to Sampled as that's the best we can do in DD. I guess we can set up dispatch in Rasters so that Projected can "win" ?

And the other case is just pragmatic, they are both X at least. 

Maybe we can have strictness levels?

tiemvanderdeure:
So it is intentional? It was just surprising because if any of the other fields are different then it immediately throws an error. So an unordered and an ordered lookup cannot be broadcasted but a sampled and a categorical can.
E.g.
```julia
julia> rand(X(Sampled(1:3, order = Lookups.Unordered()))) .* rand(X(1:3))
ERROR: DimensionMismatch: Lookups do not all have the same order: Unordered(), ForwardOrdered().
Stacktrace:
  [1] _failed_comparedims(t::DimensionalData.Dimensions.Throw{Nothing}, msg_intro::String)
    @ DimensionalData.Dimensions C:\Users\tsh371\.julia\packages\DimensionalData\iHSqC\src\Dimensions\primitives.jl:569
  [2] _orderaction(err::DimensionalData.Dimensions.Throw{Nothing}, a::X{Sampled{…}}, b::X{Sampled{…}})
    @ DimensionalData.Dimensions C:\Users\tsh371\.julia\packages\DimensionalData\iHSqC\src\Dimensions\primitives.jl:566
  [3] #_comparedims2#89
    @ C:\Users\tsh371\.julia\packages\DimensionalData\iHSqC\src\Dimensions\primitives.jl:616 [inlined]
  [4] _comparedims2
    @ C:\Users\tsh371\.julia\packages\DimensionalData\iHSqC\src\Dimensions\primitives.jl:604 [inlined]
  [5] _comparedims2
[...]

PR #879, opened 2024-12-04
add a check for valtypename

tiemvanderdeure:
Solves https://github.com/rafaqz/DimensionalData.jl/issues/878

E.g.
```julia
using DimensionalData, DimensionalData.Lookups
z = Z(1:2)
z2 = Z([1,2])
zcat = Z(Categorical(1:2))

rand(z) .* rand(z2) # still works
rand(z) .* rand(zcat) # now errors
```

With the error message `ERROR: DimensionMismatch: Lookup types for Z do not match: Sampled and Categorical.`

### DIFF
--- a/src/Dimensions/primitives.jl
+++ b/src/Dimensions/primitives.jl
@@ -558,12 +558,18 @@ _valmsg(a, b) = "Lookup values for $(basetypeof(a)) of $(parent(a)) and $(parent
 _dimsizemsg(a, b) = "Found both lengths $(length(a)) and $(length(b)) for $(basetypeof(a))."
 _valtypemsg(a, b) = "Lookup for $(basetypeof(a)) of $(lookup(a)) and $(lookup(b)) do not match."
 _ordermsg(a, b) = "Lookups do not all have the same order: $(order(a)), $(order(b))."
+function _valtypenameaction(a, b)
+    typenamea = Base.nameof(typeof(parent(a)))
+    typenameb = Base.nameof(typeof(parent(b)))
+    return "Lookup types for $(basetypeof(a)) do not match: $typenamea and $typenameb."
+end
 
 @noinline _dimsmismatchaction(err, a, b) = _failed_comparedims(err, _dimsmismatchmsg(a, b))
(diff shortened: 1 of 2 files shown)


rafaqz:
Isn't this too strong? Now `Sampled` and `Projected` will also error. 

Which may not be what we want?

tiemvanderdeure:
Okay probably not. So maybe we want to check if they have the same abstract type. So if both are `AbstractSampled` we should allow it. I'm not sure how to implement that robustly, though.

rafaqz:
Yeah, this is probably why I didn't do it 😅

Like what does the same abstract type really mean when it's generally accepted that you can add intermediate abstract types without a breaking change.

src/Dimensions/primitives.jl
```diff
@@ -616,6 +622,10 @@ end
         isnothing(msg) || _orderaction(msg, a, b)
         return false
     end
+    if valtypename && Base.typename(typeof(parent(a))) != Base.typename(typeof(parent(b)))
```
felixcremer, line 625: We could introduce a function for this check and return false as fallback and return true for combinations that assume to be comparable

rafaqz: Still a bit complicated to know what the function should do. 

We may need to just manually distinguish AbstractSampled from AbstractCategorical, and let `promote_dims` take care of the rest

tiemvanderdeure: Maybe this function could return true for identical types, or for combinations where there is some specialized implementation of `promote_first` (which is I think just for both `AbstractSampled` or two `AbstractCategorical`)

tiemvanderdeure: The reason it returns this NoLookup currently is because nothing else is implemented for `promote_dims`/`promote_first` (I think?), so it would make sense to kind of mirror that.

rafaqz: Yeah. Its a bit odd to have the same thing in two places... I wonder if we can somehow make it share the same code.

Issue #891, opened 2025-01-05
Indexing into a multi-dimensional `DimStack` is broken if any layer is a nested `DimArray`


tiemvanderdeure:
I'll admit I don't make objects like this every day, but definitely a bug:

```julia
a = [rand(X(1:4)) for z in Z(1:2), y in Y(1:2)]
ds = DimStack((;a))
ds1 = ds[1]
typeof(ds1)
```

ds1 is a 0-dimensional DimStack for some reason, where it should be just a NamedTuple with a DimArray

tiemvanderdeure:
Not too hard to find where this one is, it even has a `TODO`

https://github.com/rafaqz/DimensionalData.jl/blob/de6ca6e4fb28431e4ebe3580e0c5b5a982ee1562/src/stack/indexing.jl#L131-L134

I'll have a stab at a fix

PR #892, opened 2025-01-05
Fix indexing of DimStacks with nested DimArrays

tiemvanderdeure:
Solves https://github.com/rafaqz/DimensionalData.jl/issues/891

The logic here is that if getindex returns the eltype of a stack, then it should return a NamedTuple. This should be more robust then checking if any of the returned objects are DimArrays.

I've tried with a few DimStacks and don't see any difference in performance - but we might want to test a bit more rigorously?

### DIFF
--- a/src/stack/indexing.jl
+++ b/src/stack/indexing.jl
@@ -119,7 +119,9 @@ for f in (:getindex, :view, :dotview)
         @propagate_inbounds function $_dim_f(s::AbstractDimStack)
             map(Base.$f, data(s))
         end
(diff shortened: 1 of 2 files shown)


rafaqz (approved):
That does seem more correct

tiemvanderdeure:
I don't know about the docs, but it seems unrelated because there aren't any stacks in those examples?

tiemvanderdeure:
I realized an alternative way  to do this is just to test whether or not `all(hasdim(D, dims(s))) && all(map(d -> d isa Dimension{<:Integer}, D))`. That might be a bit easier on the compiler?

rafaqz:
I think I like the current way better. But either way it would be nice to have a test of the edge cases this fixes

PR #894, opened 2025-01-07
nuke surface! overrides and use correct dimensional arrays.

asinghvi17:


asinghvi17:
Assuming I fixed this somewhere else so closing

rafaqz:
Not sure it was 😅

PR #899, opened 2025-01-11
Matrix lookups

rafaqz:
This PR adds a `MatrixLookup` that allows x/y etc matrices instead of just grid aligned vectors.

It could be extended to ndim arrays, but the main case is matrices and the number of required selector methods explode with each dimension added.

This is a slow `findfirst`/`findmin` algorithm, an extension on NearestNeighbors.jl will be much faster so thats the next step


src/Lookups/selector.jl
```diff
@@ -138,12 +138,11 @@ end
...
     v = val(sel)
     r = round(Int, v)
-    at = atol(sel)
-    if isnothing(at)
+    if isnothing(atol(sel))
         v == r || _selnotfound_or_nothing(err, lookup, v)
     else
         at >= 0.5 && error("atol must be small than 0.5 for NoLookup")
-        isapprox(v, r; atol=at) || _selnotfound_or_nothing(err, lookup, v)
+        isapprox(v, r; atol=atol(at)) || _selnotfound_or_nothing(err, lookup, v)
```
rafaqz, line 145:
```suggestion
        isapprox(v, r; atol=atol(sel)) || _selnotfound_or_nothing(err, lookup, v)
```

src/Lookups/selector.jl
```diff
@@ -138,12 +138,11 @@ end
 function at(lookup::NoLookup, sel::At; err=_True(), kw...)
     v = val(sel)
     r = round(Int, v)
-    at = atol(sel)
-    if isnothing(at)
+    if isnothing(atol(sel))
         v == r || _selnotfound_or_nothing(err, lookup, v)
     else
         at >= 0.5 && error("atol must be small than 0.5 for NoLookup")
```
rafaqz, line 144:
```suggestion
        atol(sel) >= 0.5 && error("atol must be small than 0.5 for NoLookup")
```

PR #900, opened 2025-01-12
Tree

rafaqz:
This PR adds a `AbsractDimTree` and `DimTree` for a mutable, fast-compiled and nested alternative to DimStack.

This object is intended to represent complicated HDF5/NetCDF/Zarr etc projects easily, and to allow easily constructing and modifying them itereratively.

Dimensions are nested in that parent dimensions flow throw to child DimTrees, but can be duplicated in sibling DimTrees. This means DimTree could represent multiple resolutions of X/Y arrays over the same time dimension as needed in cloud optimized pyramids, or multiple matrices that together cover the globle for e.g. oceanography applications like MeshArrays.jl.

DimTree will _not_ have fast data access like DimArray/DimStack. Currently indexing is not defined at all other than `Symbol` for layers, but probably interval selectors should work for spatial/temporal subsetting of the whole tree. I guess this would need to drop groups that fall outside the selection.

Algorithms needing performance can easily convert DimTree to DimStack (droppping parent/child objects) so that access is fast.

### DIFF
--- a/src/DimensionalData.jl
+++ b/src/DimensionalData.jl
@@ -66,6 +66,8 @@ export AbstractDimStack, DimStack
 
 export AbstractDimTable, DimTable
 
+export AbstractDimTree, DimTree
+
 export DimIndices, DimSelectors, DimPoints, #= deprecated =# DimKeys
 
 # getter methods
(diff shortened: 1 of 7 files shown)


asinghvi17:
https://mitgcm.readthedocs.io/en/latest/algorithm/horiz-grid.html may also be an interesting use case - it's 2D grids where the lookup values are the same, but the sampling is different (for variable 1 it's Start, variable 2 it's Center, etc)

rafaqz:
Haha I never imagined that use case.

Why do they go to the trouble of storing all of them?

Issue #745, opened 2024-06-21
friction in dimension-specific `mean` of a `groupby`


haakon-e:
Suppose I define
```julia
julia> times = Ti(Date(2000,1):Month(1):Date(2003,12))

julia> xs = X(1:3)

julia> A = rand(xs, times)
╭──────────────────────────╮
│ 3×48 DimArray{Float64,2} │
├──────────────────────────┴──────────────────────────────────────────────────────────────── dims ┐
  ↓ X  Sampled{Int64} 1:3 ForwardOrdered Regular Points,
  → Ti Sampled{Date} Date("2000-01-01"):Month(1):Date("2003-12-01") ForwardOrdered Regular Points
└─────────────────────────────────────────────────────────────────────────────────────────────────┘
 ↓ →   2000-01-01   2000-02-01  …   2003-11-01   2003-12-01
 1    0.666398             0.484909                0.168768             0.280385
 2    0.00175533           0.17132                 0.323891             0.717875
 3    0.116884             0.503126                0.927827             0.33975
```
Then I want to groupby and average over years. If I just do "vanilla" mean, I get a `DimArray` back:
```julia
julia> mean.(groupby(A, Ti => year))
╭───────────────────────────────╮
│ 4-element DimArray{Float64,1} │
├───────────────────────────────┴──────────────────────────────────────── dims ┐
  ↓ Ti Sampled{Int64} [2000, 2001, 2002, 2003] ForwardOrdered Irregular Points
├──────────────────────────────────────────────────────────────────── metadata ┤
  Dict{Symbol, Any} with 1 entry:
  :groupby => :Ti=>year
└──────────────────────────────────────────────────────────────────────────────┘
 2000  0.500896
 2001  0.605857
 2002  0.450127
 2003  0.52983
```
However, this also averages over `X`. So instead, I specify `mean(A, dims=Ti)`
```julia
julia> B = mean.(groupby(A, Ti => year), dims = Ti)
╭──────────────────────────────────────────────────╮
│ 4-element DimGroupByArray{DimArray{Float64,2},1} │
├──────────────────────────────────────────────────┴───────────────────── dims ┐
  ↓ Ti Sampled{Int64} [2000, 2001, 2002, 2003] ForwardOrdered Irregular Points
├──────────────────────────────────────────────────────────────────── metadata ┤
  Dict{Symbol, Any} with 1 entry:
  :groupby => :Ti=>year
[...]

PR #903, opened 2025-01-14
Breaking: add `combine` method for `groupby` output, fixing `similar` for `AbstractDimStack`

rafaqz:
This PR introduces a `combine` method to make things like this easier:

`combine(f, groupby(dim_array_or_stack, Ti=month); dims=Ti)`

It helps with recomposing grouped output into a single DimArray/DimStack allowing for partial reductions of the grouped content.

Surprisingly hardly any code needed to do this, once I actually fixed `similar` to make more sense on DimStack, and a few bugs in `setindex!` on stacks. I guess in the past it wasn't totally clear how these things should work. But making `combine` just work like this for arrays or stacks probably means the design is correct now.

`similar` now matches `NamedTuple` eltypes as layer types, and keeps the same dimension structure as the original stack (layers missing dimensions are still missing dimensions after `similar`).

### DIFF
--- a/src/groupby.jl
+++ b/src/groupby.jl
@@ -249,7 +249,6 @@ Group some data along the time dimension:
 
 ```jldoctest groupby; setup = :(using Random; Random.seed!(123))
 julia> using DimensionalData, Dates
-
 julia> A = rand(X(1:0.1:20), Y(1:20), Ti(DateTime(2000):Day(3):DateTime(2003)));
 
 julia> groups = groupby(A, Ti => month) # Group by month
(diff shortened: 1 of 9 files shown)


src/groupby.jl
```diff
@@ -447,11 +447,41 @@ end
...
+
+"""
+    combine(f::Function, gb::DimGroupByArray; dims=:)
+
+Combine the `DimGroupByArray` using funciton `f` over the group dimensions.
+
+If `dims` is given, combine only the dimensions in `dims`. The reducing function
+`f` must accept a `dims` keyword.
+"""
+function combine(f::Function, gb::DimGroupByArray{G}; dims=:) where G
```
asinghvi17, line 469: DataAPI.combine?  Since DD owns one of the types...

rafaqz: It doesn't exist yer https://github.com/JuliaData/DataAPI.jl/issues/65 ;)

felixcremer:
Did this close #745?

Issue #914, opened 2025-01-29
Set operator on `DimSelectors`


Beforerr:
I tried to mimic `xarray` behavorior for doing basic math (results only contains intersected coordinates)

```
import Base: -
-(a::DimArray, b::DimArray) = begin
    shared_selectors = intersect(DimSelectors(a), DimSelectors(b))
    invoke(-, Tuple{AbstractArray,AbstractArray}, a[shared_selectors], -b[shared_selectors])
end
```

However the shared_selectors does not keep dimension as original one. Is there anyway to do this?

rafaqz:
Basic math on a `DimArray` works the same as math on a base julia `Array` because `DimArray <: AbstractArray` and we get most of it for free.

Most likely math is more consistently and broadly implemented here than with xarray. Even functions from a third party julia packages we don't know about will usually work.

So `-` is already defined:

```julia
julia> a = rand(X(3), Y(4))
┌ 3×4 DimArray{Float64, 2} ┐
├──────────────────────────┴─────────────────────────────── dims ┐
  ↓ X, → Y
└────────────────────────────────────────────────────────────────┘
 0.356113  0.848334  0.60161    0.401837
 0.722693  0.546814  0.668194   0.0447609
 0.502677  0.820793  0.0470841  0.657557

julia> b = rand(X(3), Y(4))
┌ 3×4 DimArray{Float64, 2} ┐
├──────────────────────────┴─────────────────────────────── dims ┐
  ↓ X, → Y
└────────────────────────────────────────────────────────────────┘
 0.371906  0.209425  0.971638  0.263029
 0.290974  0.489066  0.827357  0.997566
 0.704057  0.636763  0.480227  0.365513

julia> a - b
┌ 3×4 DimArray{Float64, 2} ┐
├──────────────────────────┴─────────────────────────────── dims ┐
  ↓ X, → Y
└────────────────────────────────────────────────────────────────┘
 -0.0157936  0.63891    -0.370027   0.138808
  0.431719   0.0577473  -0.159164  -0.952805
 -0.20138    0.184029   -0.433143   0.292044
```

Additionally, its usually a bad idea to change methods like these yourself.

rafaqz:
What you are talking about is custom operations that doesn't really map to Julia `AbstractArray` interface or the usual meaning of array math.

[...]

PR #915, opened 2025-01-29
DimSelectors indexed with DimSelectors return DimSelectors

rafaqz:
Half of #914

Basically a small bug from dispatching on `AbstractRange` rather than `AbstractArray`


rafaqz:
Fixed in #916

PR #921, opened 2025-01-31
fix `@d` for only `.=`

tiemvanderdeure:
This fixes `@d da1 .= da2`, which would error because `arg` was used but not defined.

Hope this makes sense, I really have no clue what I'm doing here :)


rafaqz:
Ah must be from my recent "fix". Looks right to me

`dump(expr)` and `@macroexpand @d` are good tools here to get your head around things. AST is pretty simple but there's a lot of nesting to wade through.

PR #922, opened 2025-02-01
Document shiftlocus

rafaqz:


rafaqz:
Actually better to just use set and hide this from users

PR #927, opened 2025-02-13
updates to latest DVPress

lazarusA:


lazarusA:
@rafaqz not sure why we don't have the preview button here :(.

rafaqz:
Yeah I don't know how any of that works 😭

rafaqz:
But thanks for the update !

thofma:
I think the `components` stuff is missing from `docs/src`

lazarusA:
they are being copied automatically from DVpress, but users can update them by copying them into a `components` folder, or add new ones into it.

thofma:
thanks for the explanation. I am trying to update to a new DVpress version. I guess https://github.com/rafaqz/DimensionalData.jl/blob/main/docs/src/.vitepress/theme/VersionPicker.vue can be removed?

lazarusA:
oh, yes, good catch, we are not using that file anymore.

PR #928, opened 2025-02-17
Allow BasicDimArrays like DimPoints, DimIndices, etc in DimStack

asinghvi17:
this allows:

```julia
DimStack((; data = da, points = DimPoints(da)))`

very useful if you are trying to calculate e.g slope with stencils, since you then have the coordinates available to you.

### DIFF
--- a/src/stack/stack.jl
+++ b/src/stack/stack.jl
@@ -297,13 +297,13 @@ _maybestack(s::AbstractDimStack, xs::Tuple) = NamedTuple{keys(s)}(xs)
 # Without the `@nospecialise` here this method is also compile with the above method
 # on every call to _maybestack. And `rebuild_from_arrays` is expensive to compile.
 function _maybestack(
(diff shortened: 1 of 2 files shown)


rafaqz:
Hmmm. It doesn't really fit the model of what a DimStack _is_.

There is a rule across DD of no duplication of dimensions. That keeps everything very simple and sane as it can be 

We may need a ProtoDimindices <: AbstractArray etc that has gets turned into DimIndices, like an Array becomes a DimArray when combined with  dimensions.

asinghvi17:
I can see where you're going with that, maybe that makes sense?  But it's very confusing to tell the user that DimPoints is not a DimArray.  We don't necessarily need to promote this workflow but it should either work or error informatively...

rafaqz:
I mean, it literally _cant_ work the way DimStack works. DimPoints is not a DimArray

rafaqz:
This approach isn't really workable, so closing

PR #935, opened 2025-02-24
Add test for copyto! with start indices and for dims tuple in similar

felixcremer:
This adds tests for copyto! with an start index for the source and the destination.

It also adds a  test case for similar with a tuple of dims.


rafaqz:
@JamesWrigley can you look at this? 

It makes me somewhat nervous being close to pythons dumpster fire of dependency management.

We may need to drop it from tests?

JamesWrigley:
I also ran into this recently: https://github.com/JuliaLang/IJulia.jl/pull/1145#issuecomment-2677147781
Will make a PR to fix it.

rafaqz:
Ok @felixcremer you can update against main and this should pass

felixcremer:
I just force pushed a rebased version. Let's see where the tests are going to fail now.

rafaqz (approved):

PR #936, opened 2025-02-25
Make CondaPkg use MicroMamba in the tests

JamesWrigley:
This is the default, it just seems to be required after the latest CondaPkg update.


JamesWrigley:
Nightly appears to be failing for other reasons :sweat_smile:

rafaqz:
Yeah, we ignore nightly!

PR #951, opened 2025-03-13
Fix handling of keyword arguments in sort functions

andreasKroepelin:
Changing a comma to a semicolon makes a fine difference when splatting keyword arguments. Previously, `kw` was not given to `sort(::Matrix, ...)` as keyword arguments but as positional arguments in the form of pairs. Hence, code like
```julia
A = DimArray(rand(5, 5), (X, Y))
sortslices(A, dims = 1, by = sum)
```
to sort a `DimArray` by row-sum did not work before but works now.

### DIFF
--- a/src/array/methods.jl
+++ b/src/array/methods.jl
@@ -176,11 +176,11 @@ end
 # These just return the parent for now
 function Base.sort(A::AbstractDimVector; kw...)
     newdims = (set(only(dims(A)), NoLookup()),)
-    newdata = sort(parent(A), kw...)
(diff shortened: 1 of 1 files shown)


rafaqz:
Wow not sure how that missed testing.

Could you add a test?

andreasKroepelin:
I can add a test but I haven't understood the lookup handling yet. I agree that it would be great to reorder it as well, though.

rafaqz:
Yes I can do the lookup handling another time

PR #957, opened 2025-03-26
Add tests for plot! function without axis

felixcremer:
This adds tests for the plot! function without an axis. 
I am not entirely sure, whether we would need to do the test on every plottype or whether it is enough to test `plot!(A)` for every type of A that we currently test. 

The functions work using GLMakie.


rafaqz:
jjjjjjjjjjjjjjj

asinghvi17:
```
  ArgumentError:     Conversion failed for MakieCore.Scatter (With conversion trait MakieCore.PointBased()) with args: Tuple{StepRange{Char, Int64}, Vector{Float64}} .
      MakieCore.Scatter requires to convert to argument types Tuple{AbstractVector{<:Union{GeometryBasics.Point2, GeometryBasics.Point3}}}, which convert_arguments didn't succeed in.
      To fix this overload convert_arguments(P, args...) for MakieCore.Scatter or MakieCore.PointBased() and return an object of type Tuple{AbstractVector{<:Union{GeometryBasics.Point2, GeometryBasics.Point3}}}.`
```

indicates you are missing a call to expand_dimensions

PR #960, opened 2025-03-28
Add show tests for Begin End and tests for BeginEndRanges

felixcremer:
This adds tests for show of Begin End related concepts and also some missing tests for last, first and step for BeginEndRanges.


rafaqz:
Could also see if it handles printing something with some nested brackets...

felixcremer:
What are the use cases for | and & for Begin and End?
And why does `(1:10)[(End÷2)+1]` give 5? I would have expected to return 6.

rafaqz:
Seems like a bug, should be 6

I don't know, but once upon a time I used bit masking with `&` rather than bounds checks on arrays of length 2^n...

felixcremer:
Fixes #961

test/indexing.jl
```diff
@@ -742,11 +742,18 @@ end
         @test (1:10)[1:End] == 1:10
         @test (1:10)[Begin():End()] == 1:10
         @test (1:10)[Begin+1:End-1] == 2:9
+        @test (1:10)[1+Begin:End-1] == 2:9
         @test (1:10)[Begin()+1:End()-1] == 2:9
         @test (1:10)[Begin:End÷2] == 1:5
         @test (1:10)[Begin|3:End] == 3:10
         @test (1:10)[Begin:End&3] == 1:2
         @test (1:10)[Begin()+1:End()-1] == 2:9
+        @test_broken (1:10)[1+(End÷2)] == 6
```
rafaqz, line 751: Seems like it's doing the +1 first somehow, which would get 5

src/Lookups/beginend.jl
```diff
@@ -51,13 +52,13 @@ _to_index(inds, l::LazyMath{Begin}) = l.f(first(inds))
...
-for f in (:+, :-, :*, :÷, :|, :&)
+for f in (:+, :-, :*, :÷, :|, :&, :max, :min)
     @eval Base.$f(::Type{T}, i::Int) where T <: Union{Begin,End} = LazyMath{T}(Base.Fix2($f, i))
     @eval Base.$f(i::Int, ::Type{T}) where T <: Union{Begin,End} = LazyMath{T}(Base.Fix1($f, i))
     @eval Base.$f(::T, i::Int) where T <: Union{Begin,End} = LazyMath{T}(Base.Fix2($f, i))
     @eval Base.$f(i::Int, ::T) where T <: Union{Begin,End} = LazyMath{T}(Base.Fix1($f, i))
-    @eval Base.$f(x::LazyMath{T}, i::Int) where T = LazyMath{T}(Base.Fix2(x.f ∘ $f, i))
-    @eval Base.$f(i::Int, x::LazyMath{T}) where T = LazyMath{T}(Base.Fix1(x.f ∘ $f, i))
+    @eval Base.$f(x::LazyMath{T}, i::Int) where T = LazyMath{T}(Base.Fix2($f,i) ∘ x.f)
+    @eval Base.$f(i::Int, x::LazyMath{T}) where T = LazyMath{T}(Base.Fix1($f,1) ∘ x.f)
```
rafaqz, line 61:
```suggestion
    @eval Base.$f(i::Int, x::LazyMath{T}) where T = LazyMath{T}(Base.Fix1($f, i) ∘ x.f)
```
?

src/Lookups/beginend.jl
```diff
@@ -51,13 +52,13 @@ _to_index(inds, l::LazyMath{Begin}) = l.f(first(inds))
...
 
-for f in (:+, :-, :*, :÷, :|, :&)
+for f in (:+, :-, :*, :÷, :|, :&, :max, :min)
     @eval Base.$f(::Type{T}, i::Int) where T <: Union{Begin,End} = LazyMath{T}(Base.Fix2($f, i))
     @eval Base.$f(i::Int, ::Type{T}) where T <: Union{Begin,End} = LazyMath{T}(Base.Fix1($f, i))
     @eval Base.$f(::T, i::Int) where T <: Union{Begin,End} = LazyMath{T}(Base.Fix2($f, i))
     @eval Base.$f(i::Int, ::T) where T <: Union{Begin,End} = LazyMath{T}(Base.Fix1($f, i))
-    @eval Base.$f(x::LazyMath{T}, i::Int) where T = LazyMath{T}(Base.Fix2(x.f ∘ $f, i))
-    @eval Base.$f(i::Int, x::LazyMath{T}) where T = LazyMath{T}(Base.Fix1(x.f ∘ $f, i))
+    @eval Base.$f(x::LazyMath{T}, i::Int) where T = LazyMath{T}(Base.Fix2($f,i) ∘ x.f)
```
rafaqz, line 60:
```suggestion
    @eval Base.$f(x::LazyMath{T}, i::Int) where T = LazyMath{T}(Base.Fix2($f, i) ∘ x.f)
```

src/Lookups/beginend.jl
```diff
@@ -1,3 +1,4 @@
+import Base: max, min
```
rafaqz, line 1:
```suggestion
```

It's all explicitly Base

felixcremer:
The changes I did to make (End \div 2) +1 work break the show of Nested LazyMath. 
I am thinking we should switch to showing the  LazyMath in function form, because currently we assume that these are operators and that fails for max and min.
Therefore I would switch the from `Begin+1` to `+(Begin, 1)`.

rafaqz:
Sure. Looks a bit lispy by easier to generalise

felixcremer:
You are right, that looks very lispy  therefore I special cased max and min and we have to add other functions that we would like to make work with Begin End also to the show dispatch. 
I added the parentheses because otherwise the nested cases are a bit unclear. 
One could change that to only happen in the inner function but I am quite happy how it also works for the range. 

It currently looks like this:

```julia
julia>     lplus = Begin+6
(Begin+6)

julia>     lplusr = 6+Begin
(6+Begin)

julia>     ldiv = div(End,2)
(End÷2)

julia>     ldivnest = (End÷2) +1
((End÷2)+1)

julia>     berange = Begin:(End-1)
Begin:(End-1)

julia>     bserange = Begin:3:End
Begin:3:End

julia> max(Begin,2)
max(Begin, 2)
```

felixcremer:
This should be good to go now.

PR #962, opened 2025-03-30
bump AoG

asinghvi17:
Surprised there was no CompatHelper pr

@kapple19


kapple19:
Does CI not run for Julia 1.11?

rafaqz:
Maybe the CompatHelper config is old

PR #971, opened 2025-04-09
Fix dimension ordering when converting a DataArray to DimArray

JamesWrigley:
Previously the code would temporarily store the dimension => lookup mapping in a Dict, but that doesn't maintain insertion order. So later on it was possible for the DimArray to be created with the dimensions in the wrong order, which would cause an exception.

I think it would be good to get this in a release soon-ish since it's quite a serious bug, e.g. if a square matrix is being loaded then the dimensions could get applied in the wrong order and other cases may just fail.

### DIFF
--- a/ext/DimensionalDataPythonCall.jl
+++ b/ext/DimensionalDataPythonCall.jl
@@ -53,20 +53,20 @@ function PythonCall.pyconvert(::Type{DimArray}, x::Py, d=nothing)
 
     dim_names = Symbol.(collect(x.dims))
     coord_names = Symbol.(collect(x.coords.keys()))
-    lookups_dict = Dict{Symbol, Any}()
-    for dim in dim_names
+    lookups_vec = Pair{Symbol, Any}[]
+    for dim in reverse(dim_names) # Iterate in reverse order because of row/col major
         if dim in coord_names
             coord = getproperty(x, dim).data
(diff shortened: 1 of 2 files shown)


felixcremer:
The nightly failures seem to come from ERROR: LoadError: syntax: World age increment not at top level from other packages.

asinghvi17:
It's PrecompileTools.jl, https://github.com/JuliaLang/PrecompileTools.jl/issues/58

fix is on the way

rafaqz:
And Makie is breaking the docs again

asinghvi17:
It's AoG this time 😭

lazarusA:
> And Makie is breaking the docs again

should be easy to fix. I can take a look tomorrow if someone else doesn't beat me to it 😄 .

jpthiele:
Precompile issue in HTTP.jl is fixed in main revision https://github.com/JuliaWeb/HTTP.jl/issues/1218
so the nightly CI issue will be resolved with the next release.

lazarusA:
triggering CI to see if https://github.com/JuliaLang/PrecompileTools.jl/pull/60 and https://github.com/MakieOrg/AlgebraOfGraphics.jl/pull/646 fixed the issues.

rafaqz:
Ok it's just nightly, we can ignore that for now. Thanks everyone I'll merge an bump a patch version

PR #972, opened 2025-04-16
Add support for zero-copy conversions from an xarray to a DimArray

JamesWrigley:
In the process I also refactored the code to rely on PythonCall more, which simplified it quite a lot. One thorny issue is that this could be considered a breaking change since the parent array type changed from `Array` to `PyArray`, which is not a `DenseArray` (mentioned in the new docs). But I feel that the parent array type is a pretty internal thing to change, and since in the worst case it won't break any existing code but just make it slower, maybe we can still put it in a patch release?

### DIFF
--- a/CondaPkg.toml
+++ b/CondaPkg.toml
@@ -1,3 +1,4 @@
 [deps]
 xarray = ""
 numpy = ""
+openssl = "<=julia"
--- a/Project.toml
+++ b/Project.toml
@@ -6,7 +6,6 @@ version = "0.29.16"
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
-OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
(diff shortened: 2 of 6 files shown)


rafaqz:
I'm happy for this not to be breaking, I would assume the output is just some `AbstractArray` wrapped by a `DimArray`

Maybe make it clear in the docs that the behavior is generally guaranteed by semver but the type is not, so this is clear in future.

JamesWrigley:
Roger that, updated the docs in 9448a411.

JamesWrigley:
Blegh, CI was hitting some awful OpenSSL incompatibility:
- https://github.com/JuliaPy/PythonCall.jl/issues/586
- https://github.com/JuliaLang/julia/issues/58170

I attempted to fix it in 208c3a22.

rafaqz:
Yeah, having the python deps in out tests is a little problematic... we have S tier rock-solid package management from Pkg for everything else, but we lose that reliability by inserting a Python package manager into the workflow.

Most devs here including me don't know how to fix problems with python deps, so I am hoping this can kind of be your responsibility!

rafaqz (approved):

JamesWrigley:
Yes I can do that :saluting_face: Luckily it looks like that hack worked.

asinghvi17:
Let's port that fix to Rasters too? Looks like the same issue there.

PR #980, opened 2025-04-23
Enable individual indexing for merged lookups

asinghvi17:
```julia
ds = rand(X(1:10), Y(1:10), Ti(1:10))
m = mergedims(ds, (X, Y) => :space)

m[X(At(1))]
```
previously threw an error, with this PR it just works

What this does is that it introduces a new AbstractMergedLookup supertype that requires that the lookup have dims.  this way we can find all lookups that might be secretly multidimensional (currently just merged lookup and geometry lookup) and pass all matching dims into them directly.  That gets us back a vector index like thing that we can replace the placeholder Colon with that we receive from the split_dims.

This enables geometry lookups in Rasters to work like this as well, and also selecting by extents, crop, etc.

This isn't actually breaking in any way so can be merged for a non breaking release, I think.

Needs tests and benchmarks to check that this does not impact the getindex time, ie it should just compile away

### DIFF
--- a/src/Dimensions/indexing.jl
+++ b/src/Dimensions/indexing.jl
@@ -47,7 +47,7 @@ Convert a `Dimension` or `Selector` `I` to indices of `Int`, `AbstractArray` or 
 @inline function dims2indices(dims::DimTuple, I::DimTuple)
     extradims = otherdims(I, dims) # extra dims in the query, I
     # Extract "multi dimensional" lookups like MergedLookup or Rasters' GeometryLookup
-    multidims = Dimensions.dims(otherdims(dims, I), x -> lookup(x) isa AbstractMergedLookup && !isempty(Dimensions.dims(x, I)))
+    multidims = Dimensions.dims(otherdims(dims, I), x -> lookup(x) isa MultiDimensionalLookup && !isempty(Dimensions.dims(x, I)))
     Isorted = Dimensions.sortdims(I, dims)
     # Warn if any dims from I were not picked up by multidims
     actuallyextradims = otherdims(extradims, x -> any(y -> hasdim(y, x), multidims)) # one way setdiff(extradims, multidims) essentially
@@ -61,11 +61,10 @@ Convert a `Dimension` or `Selector` `I` to indices of `Int`, `AbstractArray` or 
     # things the same but injects the solutions to the merged lookups where
     # available and appropriate.
     return map(dims, one_to_one_idxs) do dim, idx
-        if dim in multidims
+        if hasdim(multidims, dim)
             if !(idx isa Colon)
-                error("I should be a Colon but I am a $idx: dimension $(name(dim))")
+                error("I should be a Colon but I am not: dimension $(name(dim))\nGot\n$idx")
             end
-            # TODO: Unaligned dims are NOT supported in merged lookups!!!!!!!!!
             dims2indices(dim, Dimensions.dims(I, Dimensions.dims(dim)))
         else
             idx
--- a/src/Dimensions/indexing.jl
+++ b/src/Dimensions/indexing.jl
@@ -48,19 +48,20 @@ Convert a `Dimension` or `Selector` `I` to indices of `Int`, `AbstractArray` or 
     extradims = otherdims(I, dims) # extra dims in the query, I
     # Extract "multi dimensional" lookups like MergedLookup or Rasters' GeometryLookup
     multidims = Dimensions.dims(otherdims(dims, I), x -> lookup(x) isa MultiDimensionalLookup && !isempty(Dimensions.dims(x, I)))
-    Isorted = Dimensions.sortdims(I, dims)
     # Warn if any dims from I were not picked up by multidims
     actuallyextradims = otherdims(extradims, x -> any(y -> hasdim(y, x), multidims)) # one way setdiff(extradims, multidims) essentially
     length(actuallyextradims) > 0 && _extradimswarn(actuallyextradims)
     # Run the query for the known one-to-one dimensions
+    Isorted = Dimensions.sortdims(I, dims)
     one_to_one_idxs = split_alignments(dims2indices, unalligned_dims2indices, dims, Isorted) 
-    
+    # Finally, run the query for the multidimensional dims
     # This is basically doing an Accessors.@set on the full_dim_structure
     # one_to_one_idxs is capable of indexing the whole dataset, but 
     # it doesn't know about the merged lookups.  This keeps all one-to-one
     # things the same but injects the solutions to the merged lookups where
     # available and appropriate.
     return map(dims, one_to_one_idxs) do dim, idx
+        # Note that this loop iterates over `dims` so each multidim can only be encountered once
         if hasdim(multidims, dim)
             if !(idx isa Colon)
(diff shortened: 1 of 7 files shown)


src/Dimensions/indexing.jl
```diff
@@ -45,10 +45,32 @@ Convert a `Dimension` or `Selector` `I` to indices of `Int`, `AbstractArray` or
...
+    # one_to_one_idxs is capable of indexing the whole dataset, but 
+    # it doesn't know about the merged lookups.  This keeps all one-to-one
+    # things the same but injects the solutions to the merged lookups where
+    # available and appropriate.
+    return map(dims, one_to_one_idxs) do dim, idx
+        if dim in multidims
+            if !(idx isa Colon)
+                error("I should be a Colon but I am a $idx: dimension $(name(dim))")
+            end
+            # TODO: Unaligned dims are NOT supported in merged lookups!!!!!!!!!
```
rafaqz, line 68:
```suggestion
```

I think they might be actually, this method will just be called again

src/Dimensions/indexing.jl
```diff
@@ -45,10 +45,32 @@ Convert a `Dimension` or `Selector` `I` to indices of `Int`, `AbstractArray` or
...
+    # Run the query for the known one-to-one dimensions
+    one_to_one_idxs = split_alignments(dims2indices, unalligned_dims2indices, dims, Isorted) 
+    
+    # This is basically doing an Accessors.@set on the full_dim_structure
+    # one_to_one_idxs is capable of indexing the whole dataset, but 
+    # it doesn't know about the merged lookups.  This keeps all one-to-one
+    # things the same but injects the solutions to the merged lookups where
+    # available and appropriate.
+    return map(dims, one_to_one_idxs) do dim, idx
+        if dim in multidims
```
rafaqz, line 64:
```suggestion
        if hasdim(multidims, dim)
```
A bit more resilant to change

src/Dimensions/merged.jl
```diff
@@ -1,3 +1,5 @@
+abstract type AbstractMergedLookup{T} <: Lookup{T,1} end
```
rafaqz, line 1: I think this name can be better, maybe `MultiDimensionalLookup`

rafaqz: `multidims` is used as the variable name above...

src/Dimensions/indexing.jl
asinghvi17: ah damn!  You're right

src/Dimensions/merged.jl
asinghvi17: Yeah multidimensional sounds good.

rafaqz:
Had a thought: what happens if e.g. X and and Y are in the regular dimensions, and also in the merged dimensions

asinghvi17:
In the current implementation the merged dimensions are ignored, because X and Y don't make it out of otherdims

We could in theory duplicate those.  But it seems fragile.

src/Dimensions/merged.jl
asinghvi17: should this even be 1D then?

rafaqz: Yeah thats kind of confusting really, everything is still 1d in practice

rafaqz:
Yeah, Its probably best. I just wondered about the behaviour where there are two merged dims and X/Y working on both of them, but not if there is also regular X/Y. That may be slightly confusing.

rafaqz:
I think the MergedLookup docstring need indexing examples so this isn't hidden

asinghvi17:
ok, all relevant tests are passing.

no clue why plots is failing though..are you using some strange backend?

src/Dimensions/indexing.jl
```diff
@@ -45,10 +45,32 @@ Convert a `Dimension` or `Selector` `I` to indices of `Int`, `AbstractArray` or
...
+    # one_to_one_idxs is capable of indexing the whole dataset, but 
+    # it doesn't know about the merged lookups.  This keeps all one-to-one
+    # things the same but injects the solutions to the merged lookups where
+    # available and appropriate.
+    return map(dims, one_to_one_idxs) do dim, idx
+        # Note that this loop iterates over `dims` so each multidim can only be encountered once
+        if hasdim(multidims, dim)
+            if !(idx isa Colon)
+                error("I should be a Colon but I am not: dimension $(name(dim))\nGot\n$idx")
+            end
```
asinghvi17, line 68: Codecov won't cover this because I have no idea how I'd engineer a situation where you get this.  Maybe we can take this out for performance...

rafaqz: Yeah, let's just delete it

rafaqz (approved):

src/Dimensions/indexing.jl
asinghvi17:
```suggestion
```

asinghvi17:
```

julia> using DimensionalData
imp
julia> import DimensionalData as DD

julia> using DimensionalData.Dimensions: _experimental_extent

julia> using Extents

julia> d = rand(X(1:10), Y(1:10), Z(1:10), Ti(1:10), Dim{:something}(1:2))
┌ 10×10×10×10×2 DimArray{Float64, 5} ┐
├────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────── dims ┐
  ↓ X         Sampled{Int64} 1:10 ForwardOrdered Regular Points,
  → Y         Sampled{Int64} 1:10 ForwardOrdered Regular Points,
  ↗ Z         Sampled{Int64} 1:10 ForwardOrdered Regular Points,
  ⬔ Ti        Sampled{Int64} 1:10 ForwardOrdered Regular Points,
  ◩ something Sampled{Int64} 1:2 ForwardOrdered Regular Points
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
[:, :, 1, 1, 1]
 ⋮      ⋱  

julia> dm = mergedims(d, (X, Y) => :space, (Z, Ti) => :valu)
┌ 2×100×100 DimArray{Float64, 3} ┐
├────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────────── dims ┐
  ↓ something Sampled{Int64} 1:2 ForwardOrdered Regular Points,
  → space     MergedLookup{Tuple{Int64, Int64}} [(1, 1), (2, 1), …, (9, 10), (10, 10)] d↓ X, → Y,
  ↗ dvalu      MergedLookup{Tuple{Int64, Int64}} [(1, 1), (2, 1), …, (9, 10), (10, 10)] ↓ Z, → Ti
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
[:, :, 1]
 ⋮      ⋱  

julia> ddm = dims(dm)
(↓ something Sampled{Int64} 1:2 ForwardOrdered Regular Points,
→ space     MergedLookup{Tuple{Int64, Int64}} [(1, 1), (2, 1), …, (9, 10), (10, 10)] (↓ X, → Y),
↗ valu      MergedLookup{Tuple{Int64, Int64}} [(1, 1), (2, 1), …, (9, 10), (10, 10)] (↓ Z, → Ti))

```

PR #989, opened 2025-04-29
Fix missing Colorbar in Makie heatmap

danlooo:
This PR aims to add an otherwise always missing `Colorbar` to a Makie plot of an `DimMatrix`.
Shall we add a `Project.toml` to the test dir and add Makie for testing? See also #957.

```julia
using CairoMakie
using DimensionalData

xs = X(1:10)
ys = Y(1:20)
data = rand(xs, ys)
a = DimArray(data, (xs, ys); name="air_temperature")
p = plot(a)
    
cb = filter(x -> x isa Colorbar,p.figure.content)[1]
cb.label[] == "air_temperature"
```
image

### DIFF
--- a/ext/DimensionalDataMakie.jl
+++ b/ext/DimensionalDataMakie.jl
@@ -165,7 +165,7 @@ for (f1, f2) in _paired(:plot => :heatmap, :heatmap, :image, :contour, :contourf
             end
             # Add a Colorbar for heatmaps and contourf
             # TODO: why not surface too?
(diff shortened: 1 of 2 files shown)


asinghvi17:
Ah, that's what got missed!

Makie is already tested but it would be nice to add a test that `contents(fig.layout)` includes a Colorbar, maybe somewhere here

https://github.com/rafaqz/DimensionalData.jl/blob/d8ee4779f7c41fc152eddc4f14d0c82689f2baba/test/plotrecipes.jl#L260

PR #996, opened 2025-05-02
Run JET and bugfix

rafaqz:
Code coverage is definately getting too low here, need to focus on getting it up closer to full coverage.

Until then, this PR fixes some uncovered code where JET finds bugs.


asinghvi17:
```
Error: 1 docstring not included in the manual:

    DimensionalData.Dimensions.promotedims :: Tuple{Dimension, Vararg{Dimension}}

These are docstrings in the checked modules (configured with the modules keyword)
that are not included in canonical @docs or @autodocs blocks.
```

PR #999, opened 2025-05-05
Remove non imported Dimensions/utils.jl

felixcremer:
The functionality there seems to be moved to dimension.jl 
The file is not included anywhere.


rafaqz:
I think amerge messed something up a while ago

PR #1010, opened 2025-05-09
fix makie axis keywords

### LINKED ISSUE #929: Plot recipes are still eating `axis` kwargs
```julia
using Makie, Rasters
r1 = Raster(rand(X(10), Y(10))
heatmap(r1; axis = (; title = "Your title"))
```
the title doesn't show up

rafaqz:
Remove a bunch of complexity and makes `axis=(; title="foo")` and similar work.

Closes #929

(at the same time I renamed colorbarkw to colorbar to match the other keywords like axis. the previous colorbarkw will still work but is no longer documented)

### DIFF
--- a/ext/DimensionalDataMakie.jl
+++ b/ext/DimensionalDataMakie.jl
@@ -71,13 +71,17 @@ for (f1, f2) in _paired(:plot => :scatter, :scatter, :lines, :scatterlines, :sta
     """
     @eval begin
         @doc $docstring
-        function Makie.$f1(A::AbstractDimVector; axislegendkw=(;), axis=(;), attributes...)
+        function Makie.$f1(A::AbstractDimVector; 
+            axislegendkw=(;), axislegend=axislegendkw, axis=(;), attributes...
+        )
             args, merged_attributes = _pointbased1(A, attributes; axis)
             p = Makie.$f2(args...; merged_attributes...)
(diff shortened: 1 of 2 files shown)


ext/DimensionalDataMakie.jl
```diff
@@ -71,10 +71,9 @@ for (f1, f2) in _paired(:plot => :scatter, :scatter, :lines, :scatterlines, :sta
     """
     @eval begin
         @doc $docstring
-        function Makie.$f1(A::AbstractDimVector; axislegendkw=(;), axis = (;), figure = (;), attributes...)
-            args, merged_attributes = _pointbased1(A, attributes)
-            axis_kw, figure_kw = _handle_axis_figure_attrs(merged_attributes, axis, figure)
-            p = Makie.$f2(args...; axis = axis_kw, figure = figure_kw, merged_attributes...)
+        function Makie.$f1(A::AbstractDimVector; axislegendkw=(;), axis=(;), attributes...)
```
asinghvi17, line 74: might also do the same thing for axislegendkw

asinghvi17 (commented):
No particular comments other than this one

ext/DimensionalDataMakie.jl
rafaqz: ahh I only got half way with that one

Issue #949, opened 2025-03-10
MethodError: no method matching genperm(::Tuple{Base.OneTo{Int64}}, ::Int64)


lazarusA:
the following worked in the previous version. The important bit is that `g_flux` is an ArrayView. Doing `g_flux[:]` works.

Now, not so sure this a a DD issue but rather a a Makie issue :( 

```
g_flux = land_wrapped.fluxes.g_flux
│ # create a time range
│ start_time = DateTime("2025-01-01")
│ end_time = DateTime("2025-01-10")
│ time_interval = start_time:Day(1):end_time
│ # attach a time dimension
│ dd_flux = DimArray(g_flux,  (Ti=time_interval, ); name=:g_flux,)
│ # plot
│ lines(dd_flux; figure = (; size = (600, 300)))
│ ```
│   exception =
│    MethodError: no method matching genperm(::Tuple{Base.OneTo{Int64}}, ::Int64)
│    The function `genperm` exists, but no method is defined for this combination of argument types.
│    
│    Closest candidates are:
│      genperm(::Any, ::AbstractVector{Int64})
│       @ Base permuteddimsarray.jl:86
│      genperm(::NTuple{N, Any}, ::NTuple{N, Int64}) where N
│       @ Base permuteddimsarray.jl:85

```

rafaqz:
Yeah it's hard to know whose fault these things are

Issue #941, opened 2025-03-03
Plot of `Observable{DimArray{<:Quantity}}` does not update when using `@lift`


DylanMMarques:
Hello,

I found that plots of observable of DD do not update when using unitful and `@lift`. For example:
```julia
using GLMakie, Unitful, Unitful.DefaultSymbols, DimensionalData
dd = Observable(DimArray(collect(1V:1V:10V), X(1:10)))
dd_2 = @lift identity($dd)
fig, ax, plt = lines(dd_2)
```

![Image](https://github.com/user-attachments/assets/8efe0e29-9475-441b-9112-5f8cf27ff58c)

```julia
dd[] = DimArray(collect(10V:-1V:1V), X(1:10))
dd_2
Observable(Quantity{Int64, 𝐋^2 𝐌 𝐈^-1 𝐓^-3, Unitful.FreeUnits{(V,), 𝐋^2 𝐌 𝐈^-1 𝐓^-3, nothing}}        [10 V, 9 V, 8 V, 
7 V, 6 V, 5 V, 4 V, 3 V, 2 V, 1 V])
```
`dd_2` is updated correctly but the plot does not update:

![Image](https://github.com/user-attachments/assets/16446079-8e80-4469-a924-13f50cc1e018)

I looked a little bit into the issue and the plot is updated correctly if not using `Quantity`:
```julia
using GLMakie, Unitful, Unitful.DefaultSymbols, DimensionalData
dd = Observable(DimArray(collect(1:1:10), X(1:10)))
dd_2 = @lift identity($dd)
fig, ax, plt = lines(dd_2)
dd[] = DimArray(collect(10:-1:1), X(1:10))
```

Also, it works fine if not using `@lift`. I.e.:
```julia
using GLMakie, Unitful, Unitful.DefaultSymbols, DimensionalData
dd = Observable(DimArray(collect(1:1:10), X(1:10)))
fig, ax, plt = lines(dd)
dd[] = DimArray(collect(10:-1:1), X(1:10))
```
![Image](https://github.com/user-attachments/assets/cfc483d6-6f38-4042-ade6-a3f9f91ace98)

Thanks!

rafaqz:
This must be something to do with how Makie handles quantities... we don't do anything differently between regular numeric lookups and unitful

Do plots update if you manually uses observerable axes of units as the first two arguments? Thats all we are doing underneath.

@SimonDanisch does this make any sense to you?

[...]

Issue #906, opened 2025-01-19
Different plots when using `Observable`


DylanMMarques:
Hello,

The axes are not permuted and reordered when plotting `Observable{DimArray}` creating different plots when using `Observable`.  Also, the axes label are not displayed. 

```julia
using DimensionalData, CairoMakie
dimarray = rand(Y(3), X(7))
fig, ax, plt = series(dimarray)
series(fig[1,2], Observable(dimarray))
```

![Image](https://github.com/user-attachments/assets/6f3dd111-2164-4b9f-bfdc-5fd6f2ab6373)

Probably related with [#724](https://github.com/rafaqz/DimensionalData.jl/issues/724)

rafaqz:
I didn't know that dispatch was even a thing in Makie, doesn't it need to be `series!` ? We are modifying an existing object...

felixcremer:
This might be a dispatch that I did not catch in#743. Does this with a normal array?

felixcremer:
Can you try an Axis as first argument?

DylanMMarques:
`series!` would require a axis. `series` creates an axis in `fig[1,2]`

As far as I am aware, the code I posted above is equivalent to:
```julia 
using DimensionalData, CairoMakie
dimarray = rand(Y(3), X(7))
fig, ax, plt = series(dimarray)
ax2 = Axis(fig[1,2])
series!(ax2, Observable(dimarray))
fig 
```

![Image](https://github.com/user-attachments/assets/29f199d1-fe0c-42ae-9289-db5b926ac451)

 Yes, it does work with a normal array both using `series` and `series!` (not sure if this is what you were asking):
```julia
array = rand(3, 7)
fig, ax, plt = series(array)
ax2 = Axis(fig[1,2])
series!(ax2, Observable(array))
fig 
```

![Image](https://github.com/user-attachments/assets/93b1e4e7-a379-420e-a703-a4b780829530)

rafaqz:
I'm saying it may work, but it shouldn't ;)

Also wondering if this is a Makie bug.

@asinghvi17 should Makie conversions always happen to the contents of an observable too?

asinghvi17:
Yes always, if they don't it's because we override the plot functions...

asinghvi17:
Yes always, if they don't it's because we override the plot functions...

felixcremer:
[...]

PR #1012, opened 2025-05-11
Makie extension to work with `Observable`

DylanMMarques:
Hello, 

This PR aims to improve the support of the Makie extention to use `DimArray` with `Observable`. I tested the changes and it fixes issues:  #949, #941, #929, #906, #805, #800, #719, #564, #949 (I replicated #949 with `lines(DimArray(1:10, (X(1:10),))`, and this PR fixes it)

I made some addional changes that I thought would improve the extension:
- `colorbarkw` renamed to `colorbar`
- `axislegendkw` renamed to `axislegend`
- `colorbar` and `axislegend` can be set to `false` to remove both the colorbar and the legend. 
- `Char` converstion to position: `lines(rand(X('a':'c')))` plots into the x positions `Int.('a':'c')` instead of `axes('a':'c')`. This solves issues of some plots such as: `fig, ax, plt = lines(rand(X('a':'c'))); lines!(ax, rand(X('b':'d'))`
- add additional tests

Regressions:
- `series` of `unitful` does not work as it runs into [this issue](https://github.com/MakieOrg/Makie.jl/issues/4946). Happy to add a special case to handle this while Makie's issue is not fixed.


rafaqz:
Thanks for closing so many issues in one PR!!
I think we may have made some similar changes in #1010, such as dropping the `kw` suffix from colorbar etc, and fixing the swallowed `axis` keyword. I hope it doesnt clash too badly, slightly unfortunate timing.

Are you ok to fix the merge?

Its also good idea to separate Makie and Plots tests, but the bad news is I added a few more Makie tests to plotrecipes.jl in #1010.

Also, moving files and adding code at the same time makes diffs very hard to read, so it may be best to just move it all back and rename in a separate PR.

DylanMMarques:
Thanks, I will try to make those changes and fix the merge in the next few days

rafaqz:
any updates on this? would be good to get all of these changes merged

DylanMMarques:
Sorry for taking this long, I struggled to find time last month. I made some changes to work with the most recent release of Makie. 

A few changes: 
- `series` of `Unitful` works fine now
- To work with Makie v0.24, I had to rename the kwarg `x` and `y` to `x_dim` and `y_dim` (it run in a error with `ComputePipeline.jl`). For example, for `heatmap`:
`heatmap(rand(X(10), Y(10)); x_dim = Y, y_dim = X) `

I am not sure if x_dim, y_dim, and z_dim are the best kwarg for this. Maybe the kwarg could follow `Makie` convention for example  `dim1_conversion`. Do you have any suggestion?

rafaqz:
So Makie calls them dim1, dim2, dim3 ?

Hmm x_dim is a little clearer to me, as I'm not sure which axis dim1 would be in 2d or 3d plots... but hard to know whats best.

DylanMMarques:
x_dim, y_dim is clearer to me as well and matches better with xlabel, ylabel, etc. For unitful support, Makie calls it dim1_conversion (for x axis), dim2_conversion (for y axis), and dim3_conversion (for z axis) - not sure why.

I also added categorical_dim to rainclouds, boxplot, violin. I was just thinking that maybe the labeldim from series could also be relabelled categorical_dim for consistency (or label_dim)? 

Alternatively, I could also remove the "_" of the new ones to avoid breaking more code and keeping the interface consistent

felixcremer:
We could also name it vertical_dim, horizontal_dim and depth_dim to make clear, in which direction the dimension is plotted.

rafaqz:
I think x and y is a little more generic. Maybe also remove the underscore to match Makie syntax, so xdim, ydim ?

DylanMMarques:
I think that xdim, ydim, and zdim are the best options as well.

I will set to categoricaldim for rainclouds (etc) and put back as labeldim for series. I will implement these changes later today

rafaqz:
Ok let me know when it's good to merge

DylanMMarques:
I am happy for the PR to be merged. After it is merged, I can open a PR to separate the Plots and Makie tests as `@safetestset` does not work well in the same file I think.

rafaqz (approved):

rafaqz:
Thanks!

PR #1015, opened 2025-05-27
ups for DVPress 0.2

lazarusA:
updates to DocumenterVitepress 0.2, plus some stars.

https://github.com/rafaqz/DimensionalData.jl/pull/1002 broke a lot of `doctests`.


tiemvanderdeure:
> don't know what that is? or where to use it 😄 .

it updates the docstrings with their actual output (according to the Documenter docs)

lazarusA:
> Hm I already got a test failure there. Seems `doctest` added `Dates.` everywhere even though there is `using Dates` in the docstring

no worries, I can take care of that.

PR #1016, opened 2025-05-27
fix doctests

tiemvanderdeure:
I just used `doctest(DD; fix = true)`, let's see if it works.


tiemvanderdeure:
All of this was merged into https://github.com/rafaqz/DimensionalData.jl/pull/1015

Issue #937, opened 2025-02-25
Warnings on world age problems on nightly


felixcremer:
On the current nightly CI we get a lot of these warnings for accessing some things prior to their world age

```julia
  WARNING: Detected access to binding `Dimensions.X` in a world prior to its definition world.
│    Julia 1.12 has introduced more strict world age semantics for global bindings.
│    !!! This code may malfunction under Revise.
│    !!! This code will error in future versions of Julia.
```

This is something that we would have to look into before julia 1.12 will be released. 
Unfortunately I don't see from these warnings where the access happens in the code.

rafaqz:
Interesting. We can always define those without the macro

sethaxen:
If I comment out this line then all the warnings go away: https://github.com/rafaqz/DimensionalData.jl/blob/87a0b40fd2d4fa541161c729dd0a30ddc5cf9d88/src/DimensionalData.jl#L120

asinghvi17:
This is breaking GO tests on nightly too

rafaqz:
Hopefully we just need to move the definition backwards or forwards in the load order? I haven't had time to try

PR #1026, opened 2025-06-23
no precompile

rafaqz:
For #937 

Maybe we don't need precompilation anyway, mostly people wont have these exact types. But it does make `show` faster...

But I don't know how or have time to solve the world age problem otherwise.


rafaqz:
Seems to help but CairoMakie also fails on nightly so we don't get far

tiemvanderdeure:
Closing now that this is fixed by #1027

PR #1027, opened 2025-06-24
Do not rely on age-dependent name lookup to find the base type

### LINKED ISSUE #937: Warnings on world age problems on nightly

yuyichao:
The information is in the type and does not require a name lookup to find

Fix #937

### DIFF
--- a/src/Lookups/utils.jl
+++ b/src/Lookups/utils.jl
@@ -75,7 +75,7 @@ where `Dim{:x}` is different from `Dim{:y}`.
     if T isa Union
         T
     else
-        getfield(parentmodule(T), nameof(T))
+        Base.unwrap_unionall(T).name.wrapper
     end
 end
 


tiemvanderdeure:
This replaces https://github.com/rafaqz/DimensionalData.jl/pull/1026, right?

yuyichao:
Yes

rafaqz:
Ah right same problem from ConstructionBase.jl

We used to consider those internals

PR #1028, opened 2025-06-28
update Makie and CairoMakie

rafaqz:


rafaqz:
Closed on favour of #1031

PR #1031, opened 2025-06-30
bump makie, cairomakie

haakon-e:
julia tests pass locally, except for micromamba (related to python) for which the environment failed to activate, so hopefully those tests run on the github runner.

only tested CairoMakie 0.15, so conservatively only added that to the compat, as my primary goal is to use Makie@0.24.

Once this is merged, I'll add a corresponding PR to Rasters.jl, which needs a version bump in GeometryBasics, in addition to this PR being merged.


haakon-e:
ah, saw too late that you already did this in https://github.com/rafaqz/DimensionalData.jl/pull/1028. Sorry for the duplicate PR.

rafaqz:
No worries! I don't have time to fix the error, whatever it is so if you want to push it through that would be extremely helpful

rafaqz:
Ok this was solved elsewhere

PR #1033, opened 2025-07-03
Test plot! functions without axis argument

felixcremer:
This adds tests for the plot!(dimarray) functionality for all plotting types for which the function with an axis argument works. 

This superseeds #957.


felixcremer:
The nightly failure is still https://github.com/JuliaLang/julia/issues/57959

rafaqz (approved):

PR #940, opened 2025-02-27
Overload rechunk from DiskArrays

felixcremer:
This is dependent on the corresponding PR from DiskArrays https://github.com/JuliaIO/DiskArrays.jl/pull/237 
This simply forwards the functions to the parent of the DimArray.

PR #1035, opened 2025-07-04
Overload DiskArrays.mockchunks

felixcremer:
This is the replacement of #940 . This overloads the DiskArrays.mockchunks function to work on AbstractDimArray by  forwarding this to the parent. 
This removes the method for DimStacks because this would need a bit more thought whether it actually make sense to change the chunking structure of all underlying arrays. Especially for datasets that have different number of dimensions.

### DIFF
--- a/ext/DimensionalDataDiskArraysExt.jl
+++ b/ext/DimensionalDataDiskArraysExt.jl
@@ -92,7 +92,7 @@ end
 end
 
 @static if isdefined(DiskArrays, :mockchunks)
-    DiskArrays.mockchunks(x::Union{AbstractDimStack,AbstractDimArray}, chunks) = 
+    DiskArrays.mockchunks(x::AbstractDimArray, chunks) = 
         modify(A -> DiskArrays.mockchunks(A, chunks), x)
 end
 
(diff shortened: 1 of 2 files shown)


rafaqz:
Would it make sense to actually link dimensions names/wrappers and chunks?

Then DimStack would also work

tiemvanderdeure:
But the layers of a DimStack could have different chunk patterns? Or would we then rechunk those already?

rafaqz:
They could, and this would force them to pretend to be the same! Thats actually really useful.

felixcremer:
So what so you think would be a good interface?

mockchunks(dimarray, NamedTuple)

Or how does view specify the dimensions?
 Would you remove the method which doesn't specify the dimension names? 

We should then implement it also for a DimTree.

rafaqz:
Usually we allow both  `NamedTuple` and `Tuple{Vararg{Dimension}}` because its easier programatically (with type stability) to make a tuple of dim wrappers like `map(rebuild, dims, chunks)` than it is to make a `NamedTuple`.

DimTree would probably need to be for only the dimensions defined in the bottom branch/trunk but yes we should do that too.

rafaqz:
Lets just merge this for now so the basic form exists, and do the dims version later

PR #1043, opened 2025-07-19
Delete CairoMakie compat entry

asinghvi17:
There is no point to having it because it is managed by the makie compat


rafaqz:
I think we need compat even for test extras

asinghvi17:
Ah for Aqua? Can have it skip though

rafaqz:
Seems complicated special casing. What's the reason?

Issue #925, opened 2025-02-07
compatibility with LazyArrays


tiemvanderdeure:
It could be nice to make dimensionaldata and LazyArrays.jl work together. It would probably need an extension. Currently everything I've tried with the `@d` macro returns errors somewhere and I don't really understand how their `@~` macro works. But just to demonstrate some potential use cases I've implemented a wrapper to generate a DimArray with a LazyArray parent:

```julia
using DimensionalData, LazyArrays, BenchmarkTools

_maybe_reshape(x::AbstractDimArray, dims) = parent(DimensionalData._maybe_insert_length_one_dims(x, dims))
_maybe_reshape(x, _) = x

function LazyDimArray(a)
    dimargs = filter(x -> x isa AbstractDimArray, a.args)
    isempty(dimargs) && error("No dimensional arguments found")
    dest_dims = DimensionalData.combinedims(dimargs...)
    _args = map(x -> _maybe_reshape(x, dest_dims), dimargs)
    data = LazyArray(@~ a.f.(_args...))
    rebuild(first(dimargs); data, dims = dest_dims)
end
```

Something like this could be a neat way of getting some really big dimarrays with minimal overhead:

```julia
da1 = rand(X(1:1000))
da2 = rand(Y(1:1000))

@btime sum(LazyDimArray(@~ $da1 .* $da2))
@btime sum(@d $da1 .* $da2)
A = parent(da1); B = parent(da2)'
@btime sum(LazyArray(@~ $A .* $B))
```
```
  105.847 μs (2 allocations: 96 bytes)
  1.399 ms (5 allocations: 7.63 MiB)
  105.815 μs (0 allocations: 0 bytes)
```

tiemvanderdeure:
Ideally I think something like `LazyArray(@d @~ da1 .* da2)` would work. But I think we might need some additional dispatches.

What `@~` does is basically it stops the materialize. So `@~ A .* B` is a `Base.Broadcast.Broadcasted`

rafaqz:
I don't think `@d` and `@~` will compose correctly, why not the other way around

tiemvanderdeure:
In Rasters, yes. But now that `@d` makes broadcasting across dims so easy, I think this is also a pretty logical use case.

I agree this could potentially be confusing, though.

rafaqz:
And YAX...

We can totally integrate with LaxyArrays, but there is no way lazy will be a built in keyword for the d macro that means "LazyArrays".

[...]

PR #1051, opened 2025-07-31
Implement Broadcast.instantiate function

tiemvanderdeure:
Define `Broadcast.instantiate` to more closely mirror the structure of base broadcasting. `instantiate` checks that dimensions align etc., so we don't need as many other methods (this gets rid of the `materialize!` dispatch completely)

Let's see if tests pass

### DIFF
--- a/src/array/broadcast.jl
+++ b/src/array/broadcast.jl
@@ -73,11 +73,11 @@ BroadcastStyle(a::Style{Tuple}, ::DimensionalStyle{B}) where {B} = DimensionalSt
         length(axes) == length(ds) || 
             throw(ArgumentError("Number of broadcasted dimensions $(length(axes)) larger than $(ds)"))
         axes = map(Dimensions.DimUnitRange, axes, ds)
-    else # bc already has axes which might have dimensions
+    else # bc already has axes which might have dimensions, e.g. when assigning to a DimArray
         axes = bc.axes
         Base.Broadcast.check_broadcast_axes(axes, bc.args...)
         ds = dims(axes)
-        isnothing(ds) || _comparedims_broadcast(A, bdims..., ds)
+        isnothing(ds) ? _comparedims_broadcast(A, bdims...) : _comparedims_broadcast(A, ds, bdims...)
     end
     return Broadcasted(bc.style, bc.f, bc.args, axes)
 end
--- a/src/array/broadcast.jl
+++ b/src/array/broadcast.jl
@@ -88,6 +88,7 @@ function Broadcast.copy(bc::Broadcasted{DimensionalStyle{S}}) where S
     data = copy(_unwrap_broadcasted(bc))
     # unwrap AbstractDimArray data
     data = data isa AbstractDimArray ? parent(data) : data
+    data isa AbstractArray || return data
     return rebuild(A; data, dims = dims(bc.axes), refdims=refdims(A), name=Symbol(""))
 end
 
--- a/src/array/broadcast.jl
+++ b/src/array/broadcast.jl
@@ -62,9 +62,9 @@ BroadcastStyle(::DimensionalStyle{A}, b::Style{Tuple}) where {A} = DimensionalSt
 @inline function Broadcast.instantiate(bc::Broadcasted{<:DimensionalStyle{S}}) where S
     A = _firstdimarray(bc)
     bdims = _broadcasted_dims(bc)
+    _comparedims_broadcast(A, bdims...)
     if bc.axes isa Nothing
         axes = Base.Broadcast.combine_axes(map(_unwrap_broadcasted, bc.args)...)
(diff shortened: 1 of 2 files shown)


test/broadcast.jl
```diff
@@ -227,8 +227,7 @@ end
 
     @test_throws DimensionMismatch z .= ab .+ ba
     @test_throws DimensionMismatch z .= ab .+ ac
-    # Maybe this should work...
-    # @test_throws DimensionMismatch a_ .= ab .+ ac
+    a_ .= ab .+ ac
```
tiemvanderdeure, line 230: I don't totally understand what I've changed that affects this, but putting an AnonDim now just "absorbs" all the other dimensions such that something like a_ .* ab .* ac works

tiemvanderdeure: Scratch that, `a_ .* ab .* ac` already works on main - now `a_ .= ab .* ac` also does so the behaviour is more consistent

tiemvanderdeure:
Hm I can't reproduce these errors locally. I need to look into what's going on with the JLArrays.

EDIT: it doesn't work for GPUArrays <v11.0 because the `backend` doesn't call `parent` whereas now it does through `KernelAbstractions`

test/broadcast.jl
```diff
@@ -246,7 +245,7 @@ end
 
     @test_throws DimensionMismatch z .= ab .+ ba
     @test_throws DimensionMismatch z .= ab .+ ac
-    @test_throws DimensionMismatch a_ .= ab .+ ac
+    a_ .= ab .+ ac
```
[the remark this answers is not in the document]
rafaqz: Ah right yes that should all error.

You rarely intentionally use AnonDim it's the outcome of some functions that add dimensions for some reason and we don't know what they are. Mostly edge cases I can't even remember where...

src/array/broadcast.jl
```diff
@@ -49,58 +50,51 @@ function DimensionalStyle(a::BroadcastStyle, b::BroadcastStyle)
...
+    return Broadcasted(bc.style, bc.f, bc.args, axes)
 end
 
-function Base.copyto!(dest::AbstractArray, bc::Broadcasted{DimensionalStyle{S}}) where S
-    fda = _firstdimarray(bc) 
-    isnothing(fda) || _comparedims_broadcast(fda, _broadcasted_dims(bc)...)
-    copyto!(dest, _unwrap_broadcasted(bc))
+function Base.similar(bc::Broadcasted{<:DimensionalStyle{S}}, ::Type{T}) where {S,T}
+    A = _firstdimarray(bc)
+    rebuild(A; data = similar(_unwrap_broadcasted(bc), T), dims = dims(axes(bc)), name = NoName())
```
tiemvanderdeure, line 83: This follows the current logic, even though I'm not totally sure if it really makes sense to set name to noname, but preserve metadata and refdims.

tiemvanderdeure:
Okay tests pass! I had hoped I would be able to simplify broadcasting a lot more, but it is just a lot of complicated machinery. Nothing breaking er immediately bug-fixing here, but now errors with non-compatible dimensions are always in `instantiate`. It gets us some of the way of lazy broadcasting with LazyArrays https://github.com/rafaqz/DimensionalData.jl/issues/925

src/array/broadcast.jl
rafaqz: I guess the thinking is the content the name refers to must be new, but the dimensional context of the refdims is unchanged. But yeah, probably they should all be removed. But in a separate breaking PR

src/array/broadcast.jl
```diff
@@ -62,44 +63,38 @@ BroadcastStyle(a::Style, ::DimensionalStyle{B}) where {B} = DimensionalStyle(a,
...
-    Base.Broadcast.materialize!(style, parent(dest), bc)
-    return dest
+function Base.similar(bc::Broadcasted{DimensionalStyle{S}}, ::Type{T}) where {S,T}
+    A = _firstdimarray(bc)
+    rebuild(A; data = similar(_unwrap_broadcasted(bc), T), dims = dims(axes(bc)))
 end
 
-function Base.similar(bc::Broadcast.Broadcasted{DimensionalStyle{S}}, ::Type{T}) where {S,T}
-    A = _firstdimarray(bc)
-    rebuildsliced(A, similar(_unwrap_broadcasted(bc), T, axes(bc)...), axes(bc), Symbol(""))
```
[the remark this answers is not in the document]
rafaqz: Hmm yeah I guess the output can never have less or smaller dimensions than the input, so the slicing is redundant?

test/broadcast.jl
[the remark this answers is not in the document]
tiemvanderdeure: Okay so here let me preserve the old behaviour, then we can make breaking changes in a separate PR

rafaqz (approved):

PR #1058, opened 2025-08-01
Remove Makie override for figure-axis-plot

asinghvi17:
specifically `plot(dd)` should just use the Makie pipeline, we are not doing anything different


asinghvi17:
Ah I think this captures some kwargs

PR #1061, opened 2025-08-02
Fix vcat for Datetime

### LINKED ISSUE #1060: Could not `vcat` two Time dimension.
Minimal example to reproduce

```{julia}
t1 = Ti(DateTime(2000):Day(1):DateTime(2001)-Day(1))
d1 = rand(t1)
t2 = Ti(DateTime(2001):Day(1):DateTime(2002))
d2 = rand(t2)
vcat(d1, d2)
```

stacktrace

```
ERROR: MethodError: no method matching isapprox(::DateTime, ::DateTime)
The function `isapprox` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  isapprox(::Missing, ::Any; kwargs...)
   @ Base missing.jl:90
  isapprox(::Any, ::Missing; kwargs...)
   @ Base missing.jl:91
  isapprox(::Polynomials.AbstractPolynomial{T}, ::S; kwargs...) where {S, T}
   @ Polynomials ~/.julia/packages/Polynomials/893kl/src/common.jl:1350
  ...

Stacktrace:
 [1] (::DimensionalData.var"#290#291"{…})(l::Sampled{…})
   @ DimensionalData ~/.julia/packages/DimensionalData/omXy6/src/array/methods.jl:424
 [2] map(f::DimensionalData.var"#290#291"{UnionAll, Day}, t::Tuple{Sampled{…}})
   @ Base ./tuple.jl:355
 [3] _check_cat_lookups(::Type, ::Regular{…}, ::Sampled{…}, ::Vararg{…})
   @ DimensionalData ~/.julia/packages/DimensionalData/omXy6/src/array/methods.jl:415
 [4] _check_cat_lookups
   @ ~/.julia/packages/DimensionalData/omXy6/src/array/methods.jl:409 [inlined]
 [5] check_cat_lookups
   @ ~/.julia/packages/DimensionalData/omXy6/src/array/methods.jl:400 [inlined]
 [6] vcat(::DimVector{…}, ::DimVector{…})
   @ DimensionalData ~/.julia/packages/DimensionalData/omXy6/src/array/methods.jl:374
 [7] top-level scope
   @ REPL[232]:1
```

Until https://github.com/JuliaLang/julia/issues/56879 feature is added, I am just wondering what is the best way to solve this issue.

felixcremer:
This fixes #1060 by making an equality test before making an approx test. Because the equality works for datetimes. 

This adds the example from #1060 as a test case.

### DIFF
--- a/src/array/methods.jl
+++ b/src/array/methods.jl
@@ -421,21 +421,28 @@ function _check_cat_lookups(D, ::Regular, lookups...)
             @warn _cat_warn_string(D, "step sizes $(step(span(l))) and $s do not match")
             return false
         end
-        if hasmethod(isapprox, (typeof(lastval+s), typeof(first(l))))
-            if !(lastval + s ≈ first(l))
-                @warn _cat_warn_string(D, "`Regular` lookups do not join with the correct step size: $(lastval) + $s ≈ $(first(l)) should hold")
-                return false
-            end
-        else
-            if lastval + s != first(l)
-                @warn _cat_warn_string(D, "`Regular` lookups do not join with the correct step size: $(lastval) + $s == $(first(l)) should hold since isapprox is not defined")
-                return false
-            end
-        end
+        _check_cat_step_join(D, lastval, first(l), s) || return false
         lastval = last(l)
         return true
     end |> all
 end
+
+function _check_cat_step_join(D, lastval::Number, firstval::Number, steplen)
+    if !(lastval + steplen ≈ firstval)
+        @warn _cat_warn_string(D, "`Regular` lookups do not join with the correct step size: $(lastval) + $steplen ≈ $firstval should hold")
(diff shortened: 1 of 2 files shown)


rafaqz:
But it will still have a weird error for all AbstractTime when they don't match? It should be a warning.

felixcremer:
I added a check whether isapprox is defined for the types and a fallback to isequal. I didn't want to special case DateTime because there might be other types which have isequal bot not isapprox. Also this seemed like the easiest place to add this without duplicating the other checks.

rafaqz:
I don't love using reflection like that.

We could instead dispatch on Number for `isapprox` and AbstractTime or whatever for dates and do the exact check.

rafaqz (approved):

PR #1062, opened 2025-08-03
Remove rtol from At selector

### LINKED ISSUE #895: Delete rtol in At selector, it's not used

felixcremer:
This removes rtol from the At selector. Let's see whether this passes CI.

This closes #895


felixcremer:
I don't understand the purpose of the tuple version of At. 
Now that the normal At constructor has only two parameters, this function won't end, because it will just call itself. 
And if I remove this function I get a lot of test failures because this  functionality is explicitely tested or at least assumed. 

This is the function definition:

At(a, b; kw...) = At((a, b); kw...)

rafaqz:
At, Near, Contains etc with a Tuple let you specify a range by selecting start and stop values. Which can be a useful alternative to intervals.

felixcremer:
Should we remove the two argument overload of At and just indicate that you should use a Tuple for the interval behaviour?

rafaqz:
We shouldn't need to? Just need to tweak the constructors to not loop

rafaqz:
Ok we need to remove the default constructor so you can only set atol as a keyword. so maybe the keyword version goes in the struct calls the fully typed `At{T,X}`. Then we need a `Construction Base.constructorof` def for `At`

felixcremer:
I tried putting the keyword method in the Constructor and then I get a different StackOverflowError. 
I don't understand how the internals work and what we would need for the ConstructionBase.constructorof

rafaqz:
No worries, will finish it

felixcremer:
Thanks for taking over. I get really a bit lost in the internals of DimensionalData there.

Issue #1063, opened 2025-08-04
comparedims(d1,d2) result depends on order of dimensions


felixcremer:
The result of comparedims when one of the arrays has more dims than the other depends on the order of the dimensions in the larger array. 
I would have expected that this comparison is independent on the order of the dimensions. 

```julia
julia> r1 = rand(X([1,3,4]))
┌ 3-element DimArray{Float64, 1} ┐
├────────────────────────────────┴───────────────────────── dims ┐
  ↓ X Sampled{Int64} [1, …, 4] ForwardOrdered Irregular Points
└────────────────────────────────────────────────────────────────┘
 1  0.171079
 3  0.166104
 4  0.807168

julia> r2 = rand(Y(1:10),X([1,3,4]))
┌ 10×3 DimArray{Float64, 2} ┐
├───────────────────────────┴───────────────────────────── dims ┐
  ↓ Y Sampled{Int64} 1:10 ForwardOrdered Regular Points,
  → X Sampled{Int64} [1, …, 4] ForwardOrdered Irregular Points
└───────────────────────────────────────────────────────────────┘
  ↓ →  1         3          4
  1    0.569142  0.373322   0.762811
  2    0.29472   0.540092   0.546435
  3    0.81794   0.336375   0.86178
  ⋮                         
  7    0.391197  0.862263   0.625704
  8    0.943477  0.0393667  0.948477
  9    0.10773   0.0180314  0.185559
 10    0.283775  0.286469   0.673191

julia> r3 = rand(X([1,3,4]), Y(1:10))
┌ 3×10 DimArray{Float64, 2} ┐
├───────────────────────────┴────────────────────────────── dims ┐
  ↓ X Sampled{Int64} [1, …, 4] ForwardOrdered Irregular Points,
  → Y Sampled{Int64} 1:10 ForwardOrdered Regular Points
└────────────────────────────────────────────────────────────────┘
 ↓ →  1         2         3         4         5          6         7         8         9          10
 1    0.083804  0.105279  0.818215  0.516143  0.0913519  0.467079  0.508764  0.572529  0.0145466   0.111746
 3    0.545093  0.642441  0.880817  0.123372  0.815497   0.773596  0.124326  0.310236  0.868969    0.976927
 4    0.606889  0.719435  0.20014   0.558182  0.544476   0.452265  0.979472  0.90835   0.877468    0.506196

julia> DD.Dimensions.comparedims(r2,r1)
basetypeof(a) = Y
[...]

PR #1064, opened 2025-08-04
compare only commondims when adding branch to dimtree

tiemvanderdeure:
Closes https://github.com/rafaqz/DimensionalData.jl/issues/1063

### DIFF
--- a/src/tree/tree.jl
+++ b/src/tree/tree.jl
@@ -182,7 +182,8 @@ function _adddims!(dt::AbstractDimTree, newdims::Tuple)
         setfield!(dt, :dims, newdims)
     else
         # If there are already dims, check they match before we do anything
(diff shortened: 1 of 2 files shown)


tiemvanderdeure:
@felixcremer I think this should solve your original issue?

felixcremer:
We should add the tree example to the tests.

tiemvanderdeure:
> We should add the tree example to the tests.

I just switched 2 dims in one of the existing tests. I checked that it errors on main

felixcremer:
This fixes my initial dim Tree issue.

tiemvanderdeure:
Maybe this is actually a common enough pattern we could define a `comparecommondims`?

rafaqz:
Will merge for now we can add the other method later if you want

PR #1070, opened 2025-08-10
remove meta description from homepage?  maybe that fixes it

asinghvi17:


rafaqz:
What is "it" that this fixes? ;)

asinghvi17:
The fancy homepage is broken

rafaqz:
Oh no!

lazarusA:
maybe push also a rebuild for the `stable` docs ? And yes, Vitepress frontmatter doesn't support that !

rafaqz:
@lazarusA We can bump another patch version

PR #1072, opened 2025-08-13
resolve breaking changes to broadcast

tiemvanderdeure:
My attempt to resolve the different changes to broadcast from https://github.com/rafaqz/DimensionalData.jl/pull/917and https://github.com/rafaqz/DimensionalData.jl/pull/1051.

I really don't think we need a specific broadcast style for basicdimarrays. `similar` can just handle that and we can just default to `DimensionalStyle{DefaultArrayStyle}`

Had to make a tiny tweak to `similar` to make this work for other AbstractDimArrays than just DimArray. Now `similar(raster, Float64, axes(raster)` will return a Raster


tiemvanderdeure:
Hm this test I think fails because I merged some changes to tests but not to the code (it fails on time vcat which is from a recent PR I'm quite sure?) It became a bit of a mess - how do you think we should resolve this @rafaqz

rafaqz:
I would start again and just do a `git revert` on the squash merge commit from the broadcast PR. That will simply revert all the changes from that commit.

Reading this might help:
https://www.atlassian.com/git/tutorials/undoing-changes/git-revert

tiemvanderdeure:
Superseded by https://github.com/rafaqz/DimensionalData.jl/pull/1073

Let's just revert the whole broadcast PR - then merge main into breaking - then branch from breaking and figure out a clean way to do it that doesn't break anything.

Issue #1077, opened 2025-08-15
behaviour of `similar` with axes argument is inconsistent


tiemvanderdeure:
MWE:
```julia
using DimensionalData, Rasters
ra = Raster(rand(X(1:10), Y(1:10), Z(1:1)); metadata = Dict("A" => 1), name = :myraster)[Z = 1]

ra2 = similar(ra, Int)
ra3 = similar(ra, Int, axes(ra))
ra4 = identity.(ra2)

# All of these are false
name(ra2) == name(ra3) # Symbol("") and NoName()
refdims(ra2) == refdims(ra3) # Z and ()
metadata(ra2) == metadata(ra3) # Dict and nothing
typeof(ra3) # DimArray not raster!
```

Making this consistent is not hard but will technically be breaking.

Can we fit in some changes here before the next breaking release https://github.com/rafaqz/DimensionalData.jl/pull/946?

rafaqz:
Sounds good! Ra3 is because its hitting the fallback for non AbstractDimArray and getting the dims from our sneaky dimensional axes. But it should be calling `rebuild` on `ra`.

That code is some pretty wild nested evals already, let me know if you have trouble

rafaqz:
Was this all fixed in #1082 ?

tiemvanderdeure:
Basically, this now gives:
```julia
name(ra2) == name(ra3) == name(ra4) # true - all Symbol("")
refdims(ra2) == refdims(ra3) # false - Z and (), but hard to fix?
metadata(ra2) == metadata(ra3) == metadata(ra4) # true - all keep metadata
typeof(ra3) # is now a Raster!
```

So the only outstanding thing is the refdims - but I don't think the `similar` method with axes really can retain refdims, since the axes might have completely different dims (including some identical to the refdims).

Then there is the question of whether metadata should be retained after broadcasting, but maybe that is for a different issue.

PR #1082, opened 2025-08-15
use rebuild for similar of dimarray with new axes

tiemvanderdeure:
closes https://github.com/rafaqz/DimensionalData.jl/issues/1077#issuecomment-3190958095

With new axes it never makes sense to keep refdims - since the axes might have new dimensions. For the metadata I've now opted to keep this since this is what happens for `similar(da, Int)` - but then we could also choose to always drop metadata. I'm not sure

### DIFF
--- a/src/array/array.jl
+++ b/src/array/array.jl
@@ -262,7 +262,6 @@ function _similar(A::AbstractDimArray, T::Type, shape::Tuple;
     shape isa Tuple{Vararg{Dimensions.DimUnitRange}} || return data
     rebuild(A; data, dims, refdims, name, metadata, kw...)
 end
-
 function _similar(A::AbstractArray, T::Type, shape::Tuple; kw...)
     data = similar(parent(A), T, map(_parent_range, shape))
     shape isa Tuple{Vararg{Dimensions.DimUnitRange}} || return data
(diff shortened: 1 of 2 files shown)


tiemvanderdeure:
I might actually be more in favour of by default dropping metadata. Behaviour like this is a bit weird.

```
using DimensionalData, Rasters
ra = Raster(rand(X(1:10), Y(1:10), Z(1:1)); metadata = Dict("A" => 1), name = :myraster)
ra2 = Raster(rand(X(1:10), Y(1:10), Z(1:1)))

ra .+ ra2 # has metadata
ra2 .+ ra # has no metadata
```

But I don't know what the total consequences of a change like this would be

src/array/array.jl
```diff
@@ -254,6 +254,14 @@ for s1 in (:(Dimensions.DimUnitRange), :MaybeDimUnitRange)
         end
     end
 end
+function _similar(A::AbstractDimArray, T::Type, shape::Tuple;
```
rafaqz, line 257:
```suggestion

function _similar(A::AbstractDimArray, T::Type, shape::Tuple; 
```

rafaqz:
I'm really not sure how not copying metadata plays out for Rasters.jl lets leave it for now until we have a more coherent overall metadata strategy.

`similar` probably want a test too :)

tiemvanderdeure:
> `similar` probably want a test too :)

It is really extensively tested - and had to make some changes so this change is tested as well

rafaqz:
I mean there is no test of what you changed here

(I guess it's hard because in the case it's a DimArray either way - we probably need to use a test AbstractDimArray)

PR #1084, opened 2025-08-20
Forward name keyword in groupby

felixcremer:
This should just forward the name keyword in the groupby function, so that it can be changed in the highest level.


rafaqz:
Does it need a CHANGELOG.md entry? is it really breaking?


I guess it does anyway just for added funcitonality

felixcremer:
It is not breaking I just had it for some reason already done on the breaking branch. And it seemed easier to PR against breaking than having to cherry pick the change. I added a changelog entry and force pushed the branch, because I had to rebase it to get the CHANGELOG file.

PR #1086, opened 2025-08-21
feat: Add CoordArray dimensional arrays with coordinate support

Beforerr:
## Summary

This PR introduces CoordArray, a new dimensional array type that extends AbstractDimArray to support both dimension  and non-dimension coordinates.

## Key Features

🚀 Intuitive Construction Syntax

```julia
da = CoordArray(
    data, 
    (time_dim, x_dim, y_dim, z_dim),
    coords=(; 
        latitude = ((X, Y), lat_grid),
        longitude = ((X, Y), lon_grid),
        elevation = ((Z,), elevation)
    ),
    name="temperature"
)
```

🔄 Automatic Coordinate Slicing

When indexing a CoordArray, coordinates are automatically sliced to match: subset = da[X(1:2), Y(2:3), Ti(At(DateTime(2020,1,5)))] # coordinates are automatically sliced: lat_grid[1:2, 2:3], etc.

image

Co-Authored-By: Claude noreply@anthropic.com

### DIFF
--- a/src/DimensionalData.jl
+++ b/src/DimensionalData.jl
@@ -62,6 +62,8 @@ export Begin, End
 
 export AbstractDimArray, DimArray
 
+export CoordArray
+
 export AbstractDimVector, AbstractDimMatrix, AbstractDimVecOrMat, DimVector, DimMatrix, DimVecOrMat
 
 export AbstractDimStack, DimStack
@@ -101,6 +103,7 @@ include("array/methods.jl")
 include("array/matmul.jl")
 include("array/broadcast.jl")
 include("array/show.jl")
+include("coordarray/coordarray.jl")
 # Stacks
 include("stack/stack.jl")
 include("stack/indexing.jl")
--- a/src/coordarray/coordarray.jl
+++ b/src/coordarray/coordarray.jl
@@ -1,0 +1,117 @@
+using DimensionalData: to_indices, slicedims, dimnum, show_main, show_after, print_block_separator
+
+"""
+    CoordArray <: AbstractDimArray
+
+A dimensional array type that supporting both dimension coordinates and non-dimension coordinates.
+
+## Example
+
+```julia
+using DimensionalData
+
+# Create dimension coordinates
+time_dim = Ti(DateTime(2020, 1, 1):Day(1):DateTime(2020, 1, 10))
+x_dim = X(1:5)
+y_dim = Y(1:3)
+z_dim = Z(1:4)
+
+# Create non-dimension coordinates
+lat = rand(5, 3)  # 2D coordinate
+lon = rand(5, 3)  # 2D coordinate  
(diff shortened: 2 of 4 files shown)


rafaqz:
Hang on a bit... we really try not to introduce new array types, because then your functionality is not available to Rasters.jl or YAXArrays.jl or the other major dependents of this package. Basically `AbstractDimArray` has to be able to do everything (A large fraction of our users and contributors rarely use `DimArray` let alone another new array type).

It also wont work in DimStack or DimTree without an explosion of new CoordStack, CoordTree etc etc objects.

So we need to build this functionality into the structure of the dimensions, then its generic.

I do like how simple the `coords` field makes this. 

But if we can do it with just the `dims` Tuple then it works in all dependent packages automatically keeping a very clean interface. Thats really the number 1 priority for the package.

This means inverting the relationships in your coords objects and stuffing the extra dimensions into fields on the lookups of the main `dims` Dimensions. And in fact this kind of thing already works, see ArrayLookup.


(And I want to say otherwise this is a great contribution I'm honestly sorry to put the brakes on it like this!! But this package really differentiates itself from AxisArrays.jl, AxisKeys.jl and the other options by intentionally acting as the abstract infrastructure for an ecosystem of other packages with more specific goals. The original reason to write DD at all was that I couldn't extend the array or lookup types in the other package to add my own fields and method dispatches for functionality, plotting, etc specific to spatial data in Rasters.jl)

Beforerr:
Yes I understand. I am also concerned about the design here (so put into a draft and open space for discussion).

I have tried to play with `ArrayLookup` a little bit, but found it hard to make it work without some tricks (like mimicing a vector as a matrix) and honestly do not know how to put all coordinates into dimension.

I am wondering why it won't work with DimStack or DimTree, since it's just another `AbstractDimArray`.

rafaqz:
Yeah, the interface on these things could be improved. Did you read the tests?

As for why it won't work, both DimStack and DimTree decompose their content DimArrays to the parent objects, the dims Tuple, and lists if dims and metadata for each layer.

The reason for this is consistency: there can only be one source of truth for a dimension. 

For example, if one layer has cords and the other does not, what would indexing with coord values mean? And what dimensions columns go into `DataFrame(stack)`?

There are endless problems like this if DimStack is a collection of DimArray

tiemvanderdeure:
Should we maybe mark CoordArray as experimental (similar to DimTree) so we can change the behaviour more easily in the future? Understand we have to merge at some point but it looks like the discussion wasn't 100% resolved.

Beforerr:
I think now I have a better idea to build CoordArray on top of DimStack instead and this hopefully could solve the issues raised above. I just do not have time to implement it :(

tiemvanderdeure:
Looks like @rafaqz reverted this again

Issue #513, opened 2023-08-08
implement some chainrules


lazarusA:
maybe defining a few chainrules for DD, similar to 

https://github.com/mcabbott/AxisKeys.jl/blob/master/src/chainrules.jl

would be useful for a wider adoption of DD.

sethaxen:
I'd be happy to help with this. While ChainRulesCore was originally intended to be a lightweight dep, it's more commonly now used as a weak dep in an extension.

lazarusA:
hi @sethaxen, any updates on this? 😄.

PR #1089, opened 2025-08-21
adds ChainRulesCore extension

lazarusA:
should close https://github.com/rafaqz/DimensionalData.jl/issues/513


lazarusA:
Well, I guess AI did a pretty good job 😄 , it needs a second pair of expert eyes on the primitives being used @rafaqz  

Also,

Screenshot 2025-08-21 at 22 13 02

yes, and then ?

Screenshot 2025-08-21 at 22 13 30

I still need to test this for my use case 😄 for the final test !

rafaqz:
Do you think it needs the other fields, like name, refdims, metadata?

lazarusA:
probably, let's add tests with those fields as well, although all of them are kwargs?

rafaqz:
Right, should we just use ` rebuild` instead?

Then we can switch it to AbstractDimArray at the same time

lazarusA:
Ok, I did compare with the README example from ForwardDiff: https://github.com/JuliaDiff/ForwardDiff.jl

```julia
using ForwardDiff

f(x::Vector) = sin(x[1]) + prod(x[2:end])
x = vcat(pi/4, 2:4)
ForwardDiff.gradient(f, x)
```

## DimensionalData
and now with DD, it works without errors and same output.

```julia
using DimensionalData
using ChainRulesCore

f(x::DimensionalData.DimArray) = sin(x[1]) + prod(x[2:end])
x = DimensionalData.DimArray(vcat(pi/4, 2:4), (:angle))
ForwardDiff.gradient(f, x)
ForwardDiff.hessian(f, x)
```

and with Zygote also works:

```julia
using Zygote

Zygote.gradient(f, x)
Zygote.hessian(f, x)
```

also with Raster type works

## Rasters

```julia
using Rasters
using DimensionalData
using ChainRulesCore

x = DimensionalData.DimArray(vcat(pi/4, 2:4), (:angle))
ras = Raster(x)
f(x::Raster) = sin(x[1]) + prod(x[2:end])
ForwardDiff.gradient(f, ras)
# with zygote
Zygote.gradient(f, ras)
Zygote.hessian(f, ras)
```
and the YAXArray type

## YAXArrays

```julia
using YAXArrays
using DimensionalData
using ChainRulesCore

yax = YAXArray(vcat(pi/4, 2:4))

f(x::YAXArray) = sin(x[1]) + prod(x[2:end])

g = ForwardDiff.gradient(f, yax)
g |> collect

# with zygote
gz = Zygote.gradient(f, yax)
gzh = Zygote.hessian(f, yax)
```

rafaqz:
I guess I don't totally understand if its ok that all `AbstactDimArray` end up as the standard `DimArray` with only `data` and `dims` fields propagated, or if there are any consequences to that?

I'm ok to merge anyway if its working for you as-is

lazarusA:
Yeah, don't know either. I guess the goal is to specify what is the actual data. Let's move forward, I guess, and iterate over if things don't work out 😕

PR #1094, opened 2025-08-31
fix: improve dimension mismatch error messages for broadcasting

Beforerr:
- Fix DimUnitRange.show to display concise axis information instead of  verbose lookup data that can be thousands of characters long
- Fixes issue #1093 where large dimension arrays would crash I/O with  extremely verbose error messages

### DIFF
--- a/test/dimunitrange.jl
+++ b/test/dimunitrange.jl
@@ -1,5 +1,5 @@
 using DimensionalData
-using DimensionalData.Dimensions: Dim, DimUnitRange
+using DimensionalData.Dimensions: Dim, DimUnitRange, basetypeof
 using OffsetArrays
 using Test
 
@@ -14,7 +14,7 @@ axdims = [
     @test parent(r) === ax
     @test dims(r) === r.dim
     @test dims((r, r2)) === (r.dim, r2.dim)
-    @test sprint(show, "text/plain", r) == "$(DimUnitRange)$((r.range, r.dim))"
+    @test sprint(show, "text/plain", r) == "$(basetypeof(r.dim))($(r.range))"
     @test length(r) == length(ax)
     @test !isempty(r)
     @test first(r) == first(ax)


src/Dimensions/dimunitrange.jl
```diff
@@ -27,7 +27,8 @@ end
 # this is necessary to ensure that keyword syntax for DimArray works correctly
 Base.Slice(r::DimUnitRange) = Base.Slice(parent(r))
 
-Base.show(io::IO, r::DimUnitRange) = print(io, DimUnitRange, (r.range, r.dim))
+# Base.show(io::IO, r::DimUnitRange) = print(io, DimUnitRange, (r.range, r.dim))
```
rafaqz, line 30:
```suggestion
```

src/Dimensions/primitives.jl
```diff
@@ -559,7 +559,7 @@ end
 Throw() = Throw(nothing)
 
 _dimsmismatchmsg(a, b) = "$(basetypeof(a)) and $(basetypeof(b)) dims on the same axis."
-_valmsg(a, b) = "Lookup values for $(basetypeof(a)) of $(parent(a)) and $(parent(b)) do not match."
+_valmsg(a, b) = "Lookup values for $(basetypeof(a)) of $(_limit_sprint(parent(a))) and $(_limit_sprint(parent(b))) do not match."
```
rafaqz, line 562: Where is `_limit_sprint` defined ?

src/Dimensions/primitives.jl
```diff
@@ -559,7 +559,7 @@ end
 Throw() = Throw(nothing)
```
Beforerr, line 560: _valmsg(a, b) = "Lookup values for $(basetypeof(a)) of $(parent(a)) and $(parent(b)) do not match."

src/Dimensions/primitives.jl
Beforerr: _limit_sprint(x) = sprint(show, x; context = :limit => true)

src/Dimensions/primitives.jl
Beforerr, line 560: _limit_sprint(x) = sprint(show, x; context = :limit => true)

Beforerr, line 560:
```suggestion
_limit_sprint(x) = sprint(show, x; context = :limit => true)
```

src/Dimensions/dimunitrange.jl
```diff
@@ -27,7 +27,7 @@ end
 # this is necessary to ensure that keyword syntax for DimArray works correctly
 Base.Slice(r::DimUnitRange) = Base.Slice(parent(r))
 
-Base.show(io::IO, r::DimUnitRange) = print(io, DimUnitRange, (r.range, r.dim))
+Base.show(io::IO, r::DimUnitRange) = print(io, basetypeof(r.dim), "(", r.range, ")") # concise presentation
```
rafaqz, line 30: Its a bit off making it look like a Dimension rather than DimUnitRange. But it is concise...

Beforerr: Yeas, it is hard decision.... This form looks more like the parent, so kinda of prefer it.

PR #1096, opened 2025-09-09
adds selectors rrule

lazarusA:
previous PR https://github.com/rafaqz/DimensionalData.jl/pull/1089 did not contained a rule for selectors, which is needed to slice data in certain dimensions ( targets ). This fixes that.

With this PR now this also works:

```Julia
using Zygote, ChainRulesCore, DimensionalData
ar = rand(3,3)
A = DimArray(ar, (Y([:a,:b,:c]), X(1:3)));
grad = Zygote.gradient(x -> sum(x[Y=At(:b)]), A)
```

this should help us close https://github.com/EarthyScience/EasyHybrid.jl/issues/139

### DIFF
--- a/ext/DimensionalDataChainRulesCoreExt.jl
+++ b/ext/DimensionalDataChainRulesCoreExt.jl
@@ -37,16 +37,17 @@ end
 #! rrule for keyword getindex with selectors
 function ChainRulesCore.rrule(::typeof(getindex), A::DD.AbstractDimArray; kwargs...)
     dimsA = dims(A)
-    indices = ntuple(i -> begin
+    selectors = ntuple(i -> begin
         dim = dimsA[i]
         key = name(dim)
         if haskey(kwargs, key)
-            # Convert selector/Colon to actual indices along dimension
-            DD.Lookups.selectindices(dim, kwargs[key])
+            kwargs[key]  # the Selector passed by the user
         else
             Colon()
         end
     end, length(dimsA))
+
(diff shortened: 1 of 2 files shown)


rafaqz:
I think this isn't totally right... Don't we need a more general call with `dims2indices`?


So this and similar things would work?

```julia
using Zygote, ChainRulesCore, DimensionalData
ar = rand(3,3) 
A = DimArray(ar, (Y([:a,:b,:c]), X(1:3))); 
grad = Zygote.gradient(x -> sum(x[Y(At(:b)])), A)
```

lazarusA:
not sure about `dims2indices`. This is the thing we found to give us back the corresponding indices `DD.Lookups.selectindices(dim, kwargs[key])` when calling on a selector. 

Goal. Given `A[Y=At(:b)]`, what are the corresponding indices in the array? In the case above is `A[2, :]`. 

How can I rewrite

```Julia
indices = ntuple(i -> begin
        dim = dimsA[i]
        key = name(dim)
        if haskey(kwargs, key)
            # Convert selector/Colon to actual indices along dimension
            DD.Lookups.selectindices(dim, kwargs[key])
        else
            Colon()
        end
    end, length(dimsA))
```
using `dims2indices`?

lazarusA:
> I think this isn't totally right... Don't we need a more general call with `dims2indices`?

ok, now I'm using `dims2indices`, and of course also works 🥳 !

PR #1109, opened 2025-10-09
Restrict CI to Python 3.13

JamesWrigley:
Preemptive fix since I ran into this on another project. I assume it's because of the free-threading changes.


JamesWrigley:
AFAICT the test failures are unrelated.

felixcremer:
These test failures should be fixed on main.

tiemvanderdeure:
yeah if you rebase it should work!

PR #1110, opened 2025-10-09
Fix tests no julia 1.12

tiemvanderdeure:
Replacing === with isequal should do the trick.

From the documentation of NaN:
```
Always use isnan or isequal for checking for NaN. Using x === NaN may give unexpected results:
```


tiemvanderdeure:
All green! ✅

felixcremer:
Lgtm

PR #1116, opened 2025-10-14
fix `sum(da; dims = :notadim)`

tiemvanderdeure:
Does what it says on the tin.

Reducing along a dimension that is not in the array does nothing and should not error


rafaqz (commented):
Nice

PR #1117, opened 2025-10-15
Standardise Ext

rafaqz:
This is trivial but has been annoying for ages.

Every extension name should end in "Ext"!

Hope this does't wipe out anyones PR, let me know if so


tiemvanderdeure (approved):
LGTM

PR #917, opened 2025-01-30
Breaking: `preservedims` in tables

rafaqz:
Forgot to PR this a while ago

Closes #896 

@asinghvi17 want to review?

PR #1118, opened 2025-10-15
Implement `Base.instantiate` - take 2

tiemvanderdeure:
I'm having another go at this, I'll run Rasters tests as well this time to make totally sure we're not breaking anything.

#917 introduced a `BasicDimensionalStyle` which I don't think we need at all because we can just let `similar` dispatch do the work.

### DIFF
--- a/src/array/broadcast.jl
+++ b/src/array/broadcast.jl
@@ -96,9 +96,9 @@ function Base.similar(bc::Broadcasted{<:DimensionalStyle{S}}, ::Type{T}) where {
     similar(A; data, dims = dims(axes(bc)))
 end
 
-@inline function Base.materialize!(::S, dest, bc::Broadcasted) where S<:DimensionalStyle
+@inline function Base.materialize!(::DimensionalStyle, dest, bc::Broadcasted)
     # check dimensions
-    bci = Broadcast.instantiate(Broadcasted{S}(bc.f, bc.args, axes(dest)))
+    bci = Broadcast.instantiate(Broadcasted(bc.style, bc.f, bc.args, axes(dest)))
     # unwrap before copying
     Base.copyto!(_unwrap_broadcasted(dest), _unwrap_broadcasted(bci))
     return dest


tiemvanderdeure:
I tried Rasters tests on this branch locally and it doesn't seem to break anything. But Rasters tests are broken right now, and there are some other things on this branch that broke it, so I'm not 100% sure.

rafaqz:
yeah, maybe lets fix those tests first. I guess its the threading on 1.12?

tiemvanderdeure:
> yeah, maybe lets fix those tests first. I guess its the threading on 1.12?

The threading is one thing and then the other breaking changes, like removing `index` and constructors for `AbstractDimArray`

tiemvanderdeure:
I just ran Rasters test to locally with this branch, https://github.com/rafaqz/Rasters.jl/pull/1030, https://github.com/rafaqz/Rasters.jl/pull/1029 and some very small local changes to dispatch on `Raster{T}(undef, ...)`.

The only failure is with some I/O things with Rasters filed with strings - I think that might be because I'm on windows and I am sure it has nothing to do with this PR.

So this is safe to merge AFAIC

rafaqz:
Hmm I will wait until you check on linux we do use broadcasts in write

tiemvanderdeure:
Okay actually the same test does pass on main, so I have to investigate a little more.

The thing is I branched this of breaking so there are a whole bunch of breaking changes in here and any of them might cause an error.

This is the test that is failing:

https://github.com/rafaqz/Rasters.jl/blob/9404d72372ca7ddff65097ecc34c61cf21ce2826/test/sources/ncdatasets.jl#L446-L451

tiemvanderdeure:
It was actually a bug (though pretty niche) - see https://github.com/JuliaIO/DiskArrays.jl/issues/284

rafaqz:
Thats what I thought it would be 😂

tiemvanderdeure:
Okay just ran Rasters tests with this branch and they all pass!!

PR #1119, opened 2025-10-16
Breaking: Include refdims as columns in DimTable

### LINKED ISSUE #884: Make `refdims` table columns
This could help a bunch of use cases like AoG plotting

See #862

sethaxen:
This PR adds a keyword `refdims` to `DimTable`, defaulting to the refdims of the input. All refdims are included as table columns. Fixes #884

Potentially breaking changes:
- `dims` field added to `DimTable`
- `refdims` columns included by default

## Example

```julia
julia> using DimensionalData, Tables

julia> using DimensionalData, Tables

julia> ds = DimStack((; x=DimArray(ones(5, 3), (X, Y)), y=DimArray(zeros(5), X)))  # current behavior unchanged
┌ 5×3 DimStack ┐
├──────────────┴───────────────────── dims ┐
  ↓ X, → Y
├────────────────────────────────── layers ┤
  :x eltype: Float64 dims: X, Y size: 5×3
  :y eltype: Float64 dims: X size: 5
└──────────────────────────────────────────┘

julia> Tables.columntable(DimTable(ds))  # without refdims, current behavior unchanged
(X = [1, 2, 3, 4, 5, 1, 2, 3, 4, 5, 1, 2, 3, 4, 5], Y = [1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3], x = [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0], y = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0])

julia> Tables.columntable(DimTable(ds[Y(3)]))  # refdims included by default
(X = [1, 2, 3, 4, 5], Y = [3, 3, 3, 3, 3], x = [1.0, 1.0, 1.0, 1.0, 1.0], y = [0.0, 0.0, 0.0, 0.0, 0.0])

julia> Tables.columntable(DimTable(ds[Y(3)]; refdims=()))  # manually opt out of refdims being included
(X = Base.OneTo(5), x = [1.0, 1.0, 1.0, 1.0, 1.0], y = [0.0, 0.0, 0.0, 0.0, 0.0])

julia> Tables.columntable(DimTable(ds; refdims=(Z(1:2),)))  # manually add new refdims (all columns repeat)
(X = [1, 2, 3, 4, 5, 1, 2, 3, 4, 5  …  1, 2, 3, 4, 5, 1, 2, 3, 4, 5], Y = [1, 1, 1, 1, 1, 2, 2, 2, 2, 2  …  2, 2, 2, 2, 2, 3, 3, 3, 3, 3], Z = [1, 1, 1, 1, 1, 1, 1, 1, 1, 1  …  2, 2, 2, 2, 2, 2, 2, 2, 2, 2], x = [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0  …  1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0], y = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0])
```

### DIFF
--- a/src/tables.jl
+++ b/src/tables.jl
@@ -90,7 +90,7 @@ To get dimension columns, you can index with `Dimension` (`X()`) or
 - `layersfrom`: Treat a dimension of an `AbstractDimArray` as layers of an `AbstractDimStack`
     by specifying a dimension to use as layers.
 - `refdims`: Additional reference dimensions to add to the table, defaults to reference
-    dimensions of the table source.
+    dimensions of the table source. Use `refdims=()` for none.
 
 # Example
 


src/tables.jl
```diff
@@ -168,6 +194,8 @@ for func in (:dims, :val, :index, :lookup, :metadata, :order, :sampling, :span,
 
 end
 
+_dims(t::DimTable) = getfield(t, :dims)
```
sethaxen: Is it at all weird that `dims(t::DimTable)` will return fewer dims than the actual dims included in the table (because it just forwards to the parent)?

src/tables.jl
```diff
@@ -58,6 +59,8 @@ To get dimension columns, you can index with `Dimension` (`X()`) or
 # Keywords
 - `mergedims`: Combine two or more dimensions into a new dimension.
 - `layersfrom`: Treat a dimension of an `AbstractDimArray` as layers of an `AbstractDimStack`.
+- `refdims`: Additional reference dimensions to add to the table, defaults to the reference
+    dimensions of the input.
```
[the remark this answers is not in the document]
rafaqz: @sethaxen CF standards has its own refdims concept so Rasters.jl Raster can load with refdims already in place.

src/tables.jl
rafaqz: I didn't even know it did that. We can fix these things and merge to breaking instead?

rafaqz: Can we not just do `refdims(dt::DimTable) = refdims(parent(dt))` ? Maybe I'm missing something

sethaxen: I guess it depends on what `dims(::DimTable)` and `refdims(::DimTable)` should mean. With this PR, a user can provide arbitrary `refdims` to the `DimTable`, so they may not even be in the parent. But should `refdims(::DimTable)` return those user-provided `refdims` (currently only stored in `(::DimTable).dims`) or return those of the parent? Should `dims(::DimTable)` return just the dims of the parent, or should it return all dimensions corresponding to columns (with this PR, also includes user-provided refdims).

tiemvanderdeure: These shenanigans are exactly why I would prefer the refdims argument were true/false. If you can provide arbitrary refdims then the dimtable is no longer just a lazy iterator over an actual array/stack but its own thing

tiemvanderdeure: > Should `dims(::DimTable)` return just the dims of the parent, or should it return all dimensions corresponding to columns (with this PR, also includes user-provided refdims).

But again, without looking at the parent there is no way of knowing for sure which columns are data columns and which are dim columns. And if we are using the dimensions of the parent to return the dimensions of the dimtable, but we're not forwarding and instead using otherdims or something then it all becomes very convoluted

rafaqz:
@sethaxen the conflicts against breaking are pretty extreme, sorry about that.

But we should try getting it merged there for the next breaking release (which is very soon)

sethaxen:
> @sethaxen the conflicts against breaking are pretty extreme, sorry about that.
> 
> But we should try getting it merged there for the next breaking release (which is very soon)

Currently the default behavior is to not add refdims to columns. What makes this breaking? And, if this will be considered breaking anyways, should I then restore the default behavior being to include refdims in columns?

rafaqz:
Oh right I misunderstood. So yes its fine to merge to main.

But: then there will be the other problem that I cant easily do the merge because neither side of the Tables.jl changes are my code, and its a lot.

If I merge to main can you commit to merging main into breaking soon afterwards? I will make sure its only your changes to merge.

sethaxen:
> Oh right I misunderstood. So yes its fine to merge to main.
> 
> But: then there will be the other problem that I cant easily do the merge because neither side of the Tables.jl changes are my code, and its a lot.
> 
> If I merge to main can you commit to merging main into breaking soon afterwards? I will make sure its only your changes to merge.

Oh I see what you mean. Oof, both PRs rewrite the tables tests. But most of the changes to tests in this PR are I think it will actually be easier to redo this PR onto breaking. Okay with me opening a new PR against breaking with these changes (with default being to include refdims as columns)?

rafaqz:
Yes changing the default would be fine on breaking. 

But this PR is already moved to breaking so no need for a new one.

src/tables.jl
```diff
@@ -78,15 +84,17 @@ To get dimension columns, you can index with `Dimension` (`X()`) or
...
 # Keywords
 - `mergedims`: Combine two or more dimensions into a new dimension.
-- `preservedims`: Preserve one or more dimensions from flattening into the table. 
+- `preservedims`: Preserve one or more dimensions from flattening into the table.
     `DimArray`s of views with these dimensions will be present in the layer column,
     rather than scalar values.
 - `layersfrom`: Treat a dimension of an `AbstractDimArray` as layers of an `AbstractDimStack`
     by specifying a dimension to use as layers.
+- `refdims`: Additional reference dimensions to add to the table, defaults to reference
+    dimensions of the table source.
```
rafaqz:
```suggestion
- `refdims`: Additional reference dimensions to add to the table, defaults to reference
    dimensions of the table source. Use `refdims=()` for none.
```

rafaqz (approved):

PR #1124, opened 2025-10-28
Extend matmul with DimUnitRange axes

### LINKED ISSUE #1122: *(::AbstractTriangular, ::DimArray) discards dimensions of triangular matrix
An `UpperTriangular(::AbstractDimMatrix)` has `DimUnitRange` axes. However, multiplying it by a `DimVector` results in compatibility of the inner dimension not being validated and of the outer dimension being discarded, e.g.

```julia
julia> using DimensionalData, LinearAlgebra, Test

julia> A = UpperTriangular(DimMatrix(randn(5, 5), (X, Y)));

julia> B = DimVector(randn(5), Y);

julia> axes(A)
(X(Base.OneTo(5)), Y(Base.OneTo(5)))

julia> A * B  # should this have dim X?
┌ 5-element DimArray{Float64, 1} ┐
├────────────────────────── dims ┤
  ↓ AnonDim Base.OneTo(1)
└────────────────────────────────┘
   1     -0.192035
 #undef   3.22561
 #undef   0.765636
 #undef   2.12322
 #undef   0.239097

julia> A' * B  # should this error?
┌ 5-element DimArray{Float64, 1} ┐
├────────────────────────── dims ┤
  ↓ AnonDim Base.OneTo(1)
└────────────────────────────────┘
   1     0.00633928
 #undef  0.147744
 #undef  0.0511726
 #undef  1.29501
 #undef  1.95363
```

I wonder if some special-casing can be performed on the following lines to check if the `AbstractMatrix` has `DimUnitRange` axes, and if so, to use those:

https://github.com/rafaqz/DimensionalData.jl/blob/6db30de4b2e1fc7f8611b7e1dc3f89dc02c78598/src/array/matmul.jl#L114-L124

Also, for some reason the multiplication is performed twice here, and the `AnonDim` is given a lookup with a different length than the dimension length. I wonder if the following is more sensible:

```julia
function DimensionalData._rebuildmul(A::AbstractMatrix, B::AbstractDimVector)
    ax1, ax2 = axes(A)
    if ax2 isa Dimensions.DimUnitRange
        isstrict = DimensionalData.strict_matmul()
        Dimensions.comparedims(Dimensions.dims(ax2), first(Dimensions.dims(B)); 
            order=isstrict, val=isstrict, length=false
        )
    end
    newdata = A * parent(B)
    if newdata isa AbstractArray
        out_dim = if ax1 isa Dimensions.DimUnitRange
            Dimensions.dims(ax1)
        else
            Dimensions.AnonDim(ax1)
        end
        DimensionalData.rebuild(B, newdata, (out_dim,))
    else
        newdata
    end
end
```

With this method, we get
```julia
julia> @inferred A * B
┌ 5-element DimArray{Float64, 1} ┐
├────────────────────────── dims ┤
  ↓ X
└────────────────────────────────┘
 -0.192035
  3.22561
  0.765636
  2.12322
  0.239097

julia> A' * B
ERROR: DimensionMismatch: X and Y dims on the same axis.
...
```

sethaxen:
Extends matmul methods to construct an `AbstractDimArray` whenever one argument is an `AbstractDimArray` and the other has `DimUnitRange` axes. Fixes #1122

### DIFF
--- a/test/matmul.jl
+++ b/test/matmul.jl
@@ -83,7 +83,9 @@ end
         @testset "$T" for T in special_types
             x = T(ones(5,5))
             @test dims(x * da) isa Tuple{<:AnonDim, Dim{:b}}
+            @test lookup(dims(x * da, 1)) == axes(x, 1)
             @test dims(da * x) isa Tuple{Dim{:a}, <:AnonDim}
+            @test lookup(dims(da * x, 2)) == axes(x, 2)
             @test typeof(x * da) <: DimArray
             @test typeof(da * x) <: DimArray
             @test parent(da' * x) == parent(da)' * x
@@ -96,7 +98,9 @@ end
         @testset "$T" for T in special_types
             x = T(ones(5,5))
             @test dims(x * dv) isa Tuple{<:AnonDim}
+            @test lookup(dims(x * dv)[1]) == axes(x, 1)
             @test dims(dv' * x) isa Tuple{<:AnonDim,<:AnonDim}
+            @test lookup(dims(dv' * x)) == (axes(dv, 2), axes(x, 2))
             @test typeof(x * dv) <: DimArray
             @test typeof(dv' * x) <: DimArray
             @test parent(dv' * x) == parent(dv)' * x
@@ -107,6 +111,7 @@ end
             @test x * dv === 15.0
             @test typeof(dv * x) <: DimArray
             @test dims(dv * x) isa Tuple{<:Dim{:vec},AnonDim}
+            @test lookup(dims(dv * x, 2)) === axes(x, 2)
             @test dv * x == vcat(x, x, x, x, x)
         end
 


sethaxen:
One coincidental behavior change is that before this PR `dim(fill(1)) === nothing`. After this PR `dim(fill(1)) == ()`. Not certain if this has any bad side effects.

tiemvanderdeure:
I guess there isn't really any answer to if `fill(1)` has dimensional axes

Maybe we can just add a `dims(x::AbstractArray{<:Any, 0}) = nothing` to not make this breaking

rafaqz (approved):

rafaqz:
~~Yeah lets special case it thats slightly undexpected.~~

Actually, fixing this will create ambiguities everywhere so we have to define `dims` twice for `AbstractDimArray`. Lets see if it causes any real problems first.

PR #1126, opened 2025-11-03
ft: add gridded interpolation to DimensionalData

haakon-e:
This pull request adds support for gridded interpolation of `AbstractDimArray`s using `Interpolations.jl`, while ensuring that the new functionality is only available when `Interpolations.jl` is loaded. The main changes introduce a new extension module, user-facing API, and integration points.

**New gridded interpolation support:**

* Added the `DimensionalDataInterpolationsExt` extension module (`ext/DimensionalDataInterpolationsExt.jl`) that defines a `DimGriddedInterpolation` type for gridded interpolation of `AbstractDimArray`s, a `gridded_interpolate` constructor, and an `interpolate!` method for in-place interpolation. This module also includes comprehensive documentation and a custom display method.

---

This PR is inspired by previous efforts and discussions to support interpolation in `DimensionalData`, including 
- #420, old discussion on interpolations in DD.
- #586, #609 an early draft for adding `Interpolations.jl` as an extension.
- #1091, adds an extension for 1D interpolation of `DimVector`s using `DataInterpolations.jl`.

I'm very open to discussing the API, bikeshed names and details, etc.
This is very much a first draft that hopefully can stimulate some discussion and feedback before I put in the additional effort to add docs, tests, and clean up the interface and edge-cases. That said, I have used a version of this for a while for an ongoing project, including a variant that works for `DimStack`s, but haven't gotten around to trying to get it merged into `DimensionalData` itself.

There's a few caveats to this draft, including:
- in-place interpolation requires equal ordering of the dimensions of the original and target array. I can't see an obvious work-around that would
    - retain allocation-free in-place interpolation
    - keep the code simple
- I needed to use an additional package, [`UnrolledUtilities`](https://clima.github.io/UnrolledUtilities.jl/dev/), to ensure that out-of-place interpolation is type-stable. Notably, it is used to move scalar-like dimensions to `refdims` so that the returned `DimArray` only contains vector-like dimensions. Unfortunately, I think this means end-users would have to explicitly load both packages for the extension to load. 
    - A workaround for this would be to add `UnrolledUtilities` as an ordinary dependency (it's quite light, after all and potentially useful elsewhere).


rafaqz:
Oh no... Im working on integrating DataInterpolationsND.jl here! (And working on that package quite a lot too)

Mostly because it will allow arbitrary mixed interpolation types by dimension, and will be faster and more flexible (e.g. GPU compatible) than Interpolations.jl due to KernelAbstractions.jl core, and becaise Im going to optimise the kernels pretty hard over the next week or two.

And unfortunately I also think we should go with one specific interpolations integration to minimise future maintenance.

haakon-e:
Ah gotcha. Well in that case, please feel free to ping me once you've got a PR ready. I'd be happy to provide feedback and discuss how it could work for some use cases I would find valuable.
I agree one extension makes most sense. I'll use this for myself until we get the DataInterpolationsND-version merged!

rafaqz:
Think I'm back on board with using Interpolations.jl.

But I want to make this a lot more flexible and have everything by dimension. I will have a go at refactoring this PR to see how easy it is to do.

Issue #412, opened 2022-09-27
Define broadcast on `DimStack`


rafaqz:
We should be able to broadcast over all layers of a `DimStack` in the same call, as with a `DimArray`.

The objects broadcasted over could be `NamedTuple` from all layers, but more interesting may be to wrap this in a wrapper object like `StackSlice{T<:NamedTuple} <: Number` so that we can use math on all layers at once.

PR #1127, opened 2025-11-03
Preserve dimensions when broadcasting dimstack

tiemvanderdeure:
After we implemented collect for dimstacks, broadcasting dimstacks is now possible but returns an array.

This one line change makes it so broadcasting operations will return DimArrays with dimensions preserved:

E.g.
```
using DimensionalData
a = rand(X(1:10), Y(1:10))
b = rand(X(1:10))
ds = DimStack((; a, b))
myfun(x::NamedTuple) = x.a + x.b
myfun.(ds)
```

We could improve on this by making the generator lazy as a AbstractDimArrayGenerator. Then we could even specialize some dispatch so that broadcasting over RasterStacks would return Rasters.

Closes https://github.com/rafaqz/DimensionalData.jl/issues/412

### DIFF
--- a/src/stack/stack.jl
+++ b/src/stack/stack.jl
@@ -627,3 +627,5 @@ Base.eltype(::Type{Base.SkipMissing{T}}) where {T<:AbstractDimStack{<:Any, NT}} 
 
 @generated _nonmissing_nt(NT::Type{<:NamedTuple{K,V}}) where {K,V} =
     NamedTuple{K, Tuple{map(Base.nonmissingtype, V.parameters)...}}
+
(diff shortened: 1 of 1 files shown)


rafaqz:
Ok great that works now! @mrke this means we can do those big broadcasts over a `RasterStack` rather than arguments.

@tiemvanderdeure does it not already return Raster from RasterStack? that should already be built in with detection of the lookup types with `dimconstructor` ?

rafaqz:
Lets add the lazy generator later, its not breaking just an implementation detail. I'll merge

PR #1128, opened 2025-11-03
Add the `D` parameter to `AbstractDimStack`

rafaqz:
Adds the `D` (dims) parameter to `AbstractDimStack` so it can be wrapped by `<: AbstractDimArrayGenerator` objects without needing messy constructors.

See https://github.com/rafaqz/DimensionalData.jl/pull/1127/files#r2488160067


tiemvanderdeure:
This one needs to be in the changelog for sure. Agree that it will make other code much cleaner but it also affects every implementation of AbstractDimStack

PR #1130, opened 2025-11-04
Add _show_data function to skip data loading during show

felixcremer:
This is to make sure, that DiskArrays don't getindex into data during show. 
I added a fallback to true so that for non-DiskArray datasets this would not change in behaviour. 

I thought about using isdisk, but I wasn't sure, where I could add this check because we don't have isdisk available in DimensionalData and in the extension it is too late. 

I also added a test, that checks that DiskArrays based DimArrays don't access the data.

### DIFF
--- a/test/makie.jl
+++ b/test/makie.jl
@@ -744,12 +744,12 @@ end
     volumeslices!(ax, A3abc; zdim=:a)
     volumeslices!(A3abc;zdim=:a)
 
-    "LScene support"
+    #LScene support 
     f, a, p = heatmap(A2ab; axis=(; type=LScene, show_axis=false))
     @test a isa LScene
     @test isnothing(a.scene[OldAxis])
 
-    "Colorbar support"
+    #Colorbar support
     fig, ax, _ = plot(A2ab; colorbar=(; width=50))
     colorbars = filter(x -> x isa Colorbar, fig.content)
     @test length(colorbars) == 1


felixcremer:
I applied your suggestions and fixed the show test. 
Locally I get the following unrelated test failure n the "inference" testset:

```julia
julia> x = DimArray(randn(2, 3, 4), (X, Y, Z));

julia> foo(x) = maximum(x; dims=(1, 2))
foo (generic function with 1 method)

julia> @inferred foo(x)
ERROR: return type DimArray{Float64, 3, Tuple{X{DimensionalData.Dimensions.Lookups.NoLookup{Base.OneTo{Int64}}}, Y{DimensionalData.Dimensions.Lookups.NoLookup{Base.OneTo{Int64}}}, Z{DimensionalData.Dimensions.Lookups.NoLookup{Base.OneTo{Int64}}}}, Tuple{}, Array{Float64, 3}, DimensionalData.NoName, DimensionalData.Dimensions.Lookups.NoMetadata} does not match inferred return type DimArray{Float64, 3, D, Tuple{}, Array{Float64, 3}, DimensionalData.NoName, DimensionalData.Dimensions.Lookups.NoMetadata} where D<:Tuple
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:44
 [2] top-level scope
   @ REPL[30]:1
```

I also get this on DimensionalData 0.29.24

felixcremer:
The last commit was needed because the string before a multiple instance assignment fails now after https://github.com/JuliaLang/julia/pull/59742. I don't see why we wanted to have a string there anyway so I converted it to proper comments.

PR #1131, opened 2025-11-05
Add a DimStackArray generator

tiemvanderdeure:
Adds `DimStackArray`, a wrapper to lazily represent a dimstack as an array.

Use this for broadcasting DimStacks

### DIFF
--- a/docs/src/api/reference.md
+++ b/docs/src/api/reference.md
@@ -40,6 +40,7 @@ hasdim
 AbstractDimStack
 DimStack
 maplayers
+DimStackArray
 ```
 
 ## DimTree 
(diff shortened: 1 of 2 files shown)


tiemvanderdeure:
I think this will work as it is. It would be nice to make it so that broadcasting a RasterStack would return a Raster and not a DimArray. We'd probably need a dispatch somewhere in `similar` to make that work.

Also need to add some tests for broadcasting a dimstack

rafaqz:
Doesnt it already work? `similar` has `dimconstructor` calls

tiemvanderdeure:
But then `dimconstructor` defaults to `DimArray` so I think no unless there are projected lookups?

rafaqz:
Yeah it only works with Projected lookups but thats most of the time

rafaqz:
Just need that docs entry fo DimStackArray and we can merge

rafaqz (approved):

PR #1132, opened 2025-11-05
remove deprecated LookupArrays

rafaqz:


felixcremer:
You have missed the import of LookupArrays in the Makie extension. Apart from that LGTM

PR #1148, opened 2025-12-09
feat: Add reindex function for coordinate-based array reindexing

### LINKED ISSUE #1147: support for reindex
I am wondering if we would like something like what `xarray.reindex` does https://docs.xarray.dev/en/stable/generated/xarray.DataArray.reindex.html.

I have working codes that support something like this, but I am not sure if I should make a PR.

The interface is something like this below

```
using DimensionalData, Dates

times = DateTime(2020, 1, 1):Day(1):DateTime(2020, 1, 10)
da = DimArray(1:10, Ti(times))

# Reindex to new times
new_times = DateTime(2020, 1, 1, 12):Day(1):DateTime(2020, 1, 10, 12)
reindex(da, Ti(new_times); method=:bfill)
```

One thing is that `Base` has `reindex` method for handling subarray.

Beforerr:
Implements `reindex` function to transform arrays to new coordinate values with multiple interpolation methods (:nearest, :linear, :ffill, :bfill). Supports multi-dimensional arrays and DimensionalData objects with configurable fill values and tolerance parameters.

Close #1147


Beforerr:
The test failure seemed unrelated and the pr passed the related testset locally.

PR #1162, opened 2026-01-07
Breaking: Make reduction methods error when passed missing dimensions

### LINKED ISSUE #1154: Some reduction methods don't error on wrong dimension names
MWE:
```julia
# Does not error
mean(rand(X(10)); dims=:foo)

# Does not error
sum(rand(X(10)); dims=:foo)

# Does error with 'ArgumentError: (X,) dims were not found in object'
std(rand(X(10)); dims=:foo)
```

I did not audit all of the methods to check their behaviour. I think DD should always throw an error if an invalid dimension name is passed.

JamesWrigley:
`dimnum()` already errors when invalid dimensions are passed so I just needed to modify the reduction methods to pass the users arguments directly to it, and fix the error message.

Also, it seems that some unrelated tests recently started failing :eyes: I attempted a fix for the NearestNeighbors.jl extension with Claude in ca67455f (I assume it broke in the last NN release) but I'm not really confident it's correct. Some inference tests were broken too, I couldn't figure out how to fix them so I just disabled them in 6c47f947.

Fixes #1154.

### DIFF
--- a/test/makie.jl
+++ b/test/makie.jl
@@ -1,6 +1,7 @@
 using DimensionalData, Test, Dates
 using AlgebraOfGraphics
 using CairoMakie
+using CairoMakie: ComputePipeline
 using ColorTypes
 using Unitful, Unitful.DefaultSymbols
 import Distributions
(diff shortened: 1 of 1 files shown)


rafaqz:
Thanks!

Issue #1046, opened 2025-07-23
easy ways to cut off some latency


tiemvanderdeure:
Inspired by a long slack discussion in gripes, I just ran a quick snoop_compile on this package, and found this one maybe problematic method.

Our definition of `Base.convert(T::Type{<:AbstractString}, name::AbstractName) = convert(T, string(name))` seems particularly bad:
https://github.com/rafaqz/DimensionalData.jl/blob/d9ac254358c6e689a3de3daa2c320a977bfaabfc/src/name.jl#L8C1-L8C87

Do we need that at all? It leads to:

```
inserting convert(T::Type{<:AbstractString}, name::DimensionalData.AbstractName) @ DimensionalData ~/.julia/dev/DimensionalData/src/name.jl:8 invalidated:
   mt_backedges:  1: signature Tuple{typeof(convert), Type{String}, Any} triggered MethodInstance for setindex!(::Dict{String, Base.UUID}, ::Any, ::Any) (0 children)
                  2: signature Tuple{typeof(convert), Type{String}, Any} triggered MethodInstance for setindex!(::Dict{String, Union{Bool, String}}, ::Any, ::Any) (0 children)
                  3: signature Tuple{typeof(convert), Type{String}, Any} triggered MethodInstance for setindex!(::Dict{String, Union{Nothing, String}}, ::Any, ::Any) (0 children)
                  4: signature Tuple{typeof(convert), Type{String}, Any} triggered MethodInstance for setindex!(::Dict{String, Nothing}, ::Nothing, ::Any) (0 children)
                  5: signature Tuple{typeof(convert), Type{String}, Any} triggered MethodInstance for setindex!(::Dict{String, Union{Nothing, Tuple{Base.PkgId, String}}}, ::Any, ::Any) (0 children)
                  6: signature Tuple{typeof(convert), Type{String}, Any} triggered MethodInstance for push!(::Vector{String}, ::Any) (0 children)
                  7: signature Tuple{typeof(convert), Type{String}, Any} triggered MethodInstance for setindex!(::Dict{String, Pkg.Types.Compat}, ::Any, ::Any) (0 children)
                  8: signature Tuple{typeof(convert), Type{String}, Any} triggered MethodInstance for (::Base.var"#cvt1#1"{Tuple{Revise.PkgData, String}, <:Tuple{Revise.PkgData, Any}})(::Int64) (0 children)
[...]

PR #1170, opened 2026-01-25
Move Adapt and ArrayInterface support into package extensions

JamesWrigley:
The benefits are a bit more marginal here, going from ~0.175s to ~0.169s. Though it should help a bit more on network filesystems where accessing files is more expensive.

Part of #1046.


JamesWrigley:
(bump)

rafaqz (approved):

PR #1184, opened 2026-04-05
:zap: (Change to _noname(::AbstractString) to support Substring)

arafune:
* Summary: Change _noname(::String) to _noname(::AbstractString) to support SubString.

* Reason: Currently, broadcasting operations (e.g., isfinite.(...)) fail with a MethodError if the array name is a SubString. Using AbstractString follows Julia's best practices and resolves this issue.

* Reproduce
```julia
using DimensionalData
da = DimArray(rand(5), (X(1:5),); name=SubString("test", 1, 2))
isfinite.(da) # This fails without the fix
```

### DIFF
--- a/src/array/array.jl
+++ b/src/array/array.jl
@@ -323,7 +323,7 @@ _arraytype(::Type{Bool}) = BitArray
 
 # Keep the same type in `similar`
 _noname(A::AbstractBasicDimArray) = _noname(name(A))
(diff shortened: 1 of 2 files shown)


tiemvanderdeure:
Can you add a test?

tiemvanderdeure (approved):
LGTM

PR #1196, opened 2026-05-11
Specialize Base.sizeof() for AbstractDimArray

JamesWrigley:
Previously `sizeof(::DimArray)` would report the size of the `DimArray` struct rather than the underlying array. Reporting the size of the array is consistent with `Array`.


JamesWrigley:
Nightly failure is https://github.com/JuliaPy/CondaPkg.jl/issues/205.

JamesWrigley:
(bump)

PR #1206, opened 2026-07-14
Add constructor for DimArray{T,N}

### LINKED ISSUE #1202: `DimArray{T,N}(undef, dims...)` fails although `Array{T,N}(undef, ...)` works
On DimensionalData v0.30.1 (Julia 1.12), the `undef` constructor is only defined for the
  single-parameter form `DimArray{T}`, not the two-parameter form `DimArray{T,N}`. This is
  asymmetric with `Base.Array`, which accepts both.

  ```julia
  using DimensionalData

  # Base.Array accepts either the {T} or {T,N} form:
  Array{Float64}(undef, 2, 3)      # ✓
  Array{Float64,2}(undef, 2, 3)    # ✓

  # DimArray only accepts the {T} form:
  DimArray{Float64}(undef, X(1:2), Y(1:3))     # ✓
  DimArray{Float64,2}(undef, X(1:2), Y(1:3))   # ✗ MethodError
  ```

  The last line throws:

  ```
  ERROR: MethodError: no method matching (DimArray{Float64, 2, D, R, A} where
  {...})(::UndefInitializer, ::X{...}, ::Y{...})
  ```

  This used to work in v0.29 and is a natural way to write the constructor when you want to
  assert the dimensionality. It would be nice to add a `DimArray{T,N}` undef method (or
  explicitly document that only `DimArray{T}` is supported).

  The N is redundant since it's already fixed by the number of dimensions passed, which is
  presumably why only {T} was kept — but the asymmetry with Base.Array is the reportable
  surprise.

tiemvanderdeure:
Closes #1202

### DIFF
--- a/src/array/array.jl
+++ b/src/array/array.jl
@@ -491,6 +491,8 @@ DimArray{T}(A::AbstractDimArray; kw...) where T = DimArray(convert.(T, A))
 DimArray{T}(A::AbstractDimArray{T}; kw...) where T = DimArray(A; kw...)
 DimArray{T}(x::UndefInitializer, dims::Dimension...; kw...) where T = DimArray{T}(x, dims; kw...)
 DimArray{T}(x::UndefInitializer, dims::MaybeDimTuple; kw...) where T = DimArray(Array{T}(undef, map(length, dims)), dims; kw...)
+DimArray{T,N}(x::UndefInitializer, dims::Dimension...; kw...) where {T,N} = DimArray{T,N}(x, dims; kw...)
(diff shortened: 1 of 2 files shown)


felixcremer:
Looks good to me. The test failure seems to be the Makie related fail.

felixcremer (commented):
Looks good to me.

tiemvanderdeure:
Anyone can review and merge @felixcremer @asinghvi17 @rafaqz ?

felixcremer:
The nightly failure is something with unbounded type parameters but seems to be unrelated.

felixcremer (approved):
LGTM

PR #1014, opened 2025-05-27
fix `cat` for rasterstacks

tiemvanderdeure:
closes https://github.com/rafaqz/DimensionalData.jl/issues/1013

PR #1208, opened 2026-07-17
Breaking: cat: preserve lookup values by demoting order to Unordered on misalignment

### LINKED ISSUE #1207: `cat` drops lookups, no avenues for sorting?
Hello,

I was trying to `cat` two stacks along the time dimensions.
both stacks are `ForwardOrdered` but not contiguous (example: 2018 jan 2018 jul 2024 jan 2024 jul)

but after `cat`ing there is the warning. 
```
`cat` cannot concatenate `Dimension`s, falling back to `parent` type:
`Ordered` lookups are misaligned at 2024-10-31T23:00:00 and 2018-05-01T00:00:00 on dimension DimensionalData.Dimensions.Ti. 
```

I am curious why would it not be possible to just relax the ordering to `Unordered` and keep the unordered lookups values so that in the future it could be possible to simply call `sortby` on a desired dimension.

can't the following just be the default behaviour in case of unordered dimensions? 
```
	newlookups = vcat(lookup(stack1, Ti), lookup(stack2, Ti))
	cat(stack1, stack2; dims=Ti(newlookups))
``` 
Am I approaching this from the wrong angle?

Thanks!

ghyatzo:
Hello,

Currently, `cat` (or `Base.cat`) of two `AbstractDimArray`s (or `AbstractDimStack`s) along a dimension whose lookups are `ForwardOrdered` but not strictly contiguous silently **drops the lookup values**, replacing them with `NoLookup()`. For example:

```julia
dd_1 = DimArray(1:3, Ti([Date(2011), Date(2013), Date(2015)]))
dd_2 = DimArray(1:3, Ti([Date(2010), Date(2012), Date(2014)]))
cat(dd_1, dd_2; dims=Ti)     # ─╯ warns "misaligned" and returns Ti(NoLookup(...))
```

Once the coordinates are gone, there is no way to sort/recover them `sortby` (or `Base.sort`) also drops lookups (differently, but just as fatally). This makes it impossible to concatenate disjoint time-series chunks and then sort the result.

Closes #1207

## Approach

Replace the `check_cat_lookups(...)::Bool` return with a **Symbol tri-state**:

| Symbol | Meaning | Action in `_cat` |
|---|---|---|
| `:ok` | contiguous ordered, or any `NoLookup` | proceed normally (`_vcat_dims`) |
| `:demote` | misaligned/mixed-order but **no exact value overlap** | values preserved via new `_vcat_dims_demote`, order → `Unordered()`, span → `Irregular` |
| `:error` | exact value overlap (`intersect(lookups...)` non-empty) | `throw(DimensionMismatch(...))` |
| `:drop` | span/step mismatch | `rebuild(catdim, NoLookup())` (unchanged) |

The new `_vcat_dims_demote` helper reuses the existing `_vcat_index` machinery (the same `maybeshiftlocus` + `reduce(vcat, …)` from the success path) to concatenate the raw lookup values, then rebuilds the dimension with `order=Unordered()` and an `Irregular` span. This keeps the coordinates intact so downstream operations like `sortperm` + indexing (or a future `sortby`) can recover a `ForwardOrdered` result.

For the fallbacks `hcat`/`vcat`/`vcat(::Dimension,…)` the Symbol return is handled but **behaviour is unchanged**: they still warn and fall back to `Base.hcat(map(parent,…))` / `Base.vcat(map(parent,…))`. 

## Example after the fix

```julia
julia> cat(dd_1, dd_2; dims=Ti)
┌ Warning: `Ordered` lookups are misaligned at 2014-01-01 and 2010-01-01
│ on dimension Ti.
└ @ DimensionalData .../methods.jl:557
┌ 6-element DimArray{Int64, 1} ┐
├──────────────────────────────┴───────────────────────────────────── dims ┐
  ↓ Ti Sampled{Date} [Date("2011-01-01"), …, Date("2014-01-01")]
      Unordered Irregular Points
└──────────────────────────────────────────────────────────────────────────┘
 2011-01-01  1
 2013-01-01  2
 2015-01-01  3
 2010-01-01  1
 2012-01-01  2
 2014-01-01  3
```

The values are preserved. A subsequent `sortby` (or `ds[Ti=sortperm(lookup(ds,Ti))]`)
recovers `ForwardOrdered`.

## What changed

**`src/array/methods.jl`**
- `check_cat_lookups` → returns `::Symbol` instead of `::Bool`
- `_check_cat_lookup_order` → returns `:ok`/`:demote`/`:error`; tracks an
  accumulated `Set` to detect exact value overlap
- `_check_cat_lookups` → two-level chain: order check passes through to span
  check only on `:ok`; `:demote` short-circuits (span validation is meaningless
  when demoted)
- `_check_cat_lookups(D, ::Regular, …)` / `::Explicit` / `::Irregular` → return
  `:ok`/`:drop` (converted from `true`/`false`)
- New `_vcat_dims_demote(d1, ds…)` — concatenates values and rebuilds with
  `order=Unordered()`, `span=Irregular(combined bounds)` for `Intervals` /
  `Irregular(nothing,nothing)` for `Points`
- `_cat` call site: branches on `:ok` / `:demote` / `:error` / `else` (`:drop`)
- `hcat`, `vcat`, `vcat(::Dimension)` call sites: only `:ok` succeeds; anything
  else falls back (unchanged behaviour)
- Fixed `Unorderd` typo → `Unordered` in `_cat_lookup_intersect_warn`

**`test/methods.jl`**
- Flipped `@test_warn "misaligned" cat(da, db; dims=2)` → `@test_throws`
  (identical Y lookups = exact overlap, now correctly errors)
- New `"misaligned ordered lookups are demoted to Unordered"` testset
  (reproduces #1207, asserts `Unordered` + values preserved)
- New `"overlap throws DimensionMismatch via cat"` testset

## Open questions

1. **`hcat`/`vcat` inconsistency**: should they also demote on misalignment
   rather than falling back to a plain Array? Currently out of scope — only
   `_cat` was changed. Maybe a separate PR?
2. **`sortby` companion**: a helper like
   `sortby(x, dim) = set(x[rebuild(dims(x,dim), sortperm(lookup(x,dim)))], dim => ForwardOrdered())`
   is a natural partner to this fix. Should it be added in a follow-up PR
   or included here?
3. **Overlap behaviour**: currently overlap (exact shared values) throws
   `DimensionMismatch` in `_cat`. For `hcat`/`vcat` it still warns and
   falls back. Should this be made consistent in a follow-up?

## Related:
- #1014 touches the same code but it's orthogonal to this one, both can be merged together mechanically.

## Disclosure
PR was an initial hand implementation of https://github.com/rafaqz/DimensionalData.jl/issues/1207#issuecomment-4980788614, then checked and refined with an agent.

### DIFF
--- a/src/array/methods.jl
+++ b/src/array/methods.jl
@@ -455,13 +455,23 @@ end
 
 function _check_cat_lookups(D, ::Explicit, lookups...)
     for l in lookups
-        span(l) isa Explicit || (_mixed_span_warn(D, Explicit, span(l)); return :drop)
+        if span(l) isa Explicit
+            continue
+        else
+            _mixed_span_warn(D, Explicit, span(l))
+            return :drop
+        end
     end
     return :ok
 end
 function _check_cat_lookups(D, ::Irregular, lookups...)
     for l in lookups
-        span(l) isa Irregular || (_mixed_span_warn(D, Irregular, span(l)); return :drop)
+        if span(l) isa Irregular
+            continue
+        else
+            _mixed_span_warn(D, Irregular, span(l))
+            return :drop
+        end
     end
     return :ok
 end
--- a/src/array/methods.jl
+++ b/src/array/methods.jl
@@ -301,10 +301,21 @@ function _cat(catdims::Tuple, A1::AbstractDimArray, As::AbstractDimArray...)
             else
                 # vcat the index for the catdim in each of Xin
                 joindims = map(A -> dims(A, catdim), Xin)
-                if !check_cat_lookups(joindims...) 
-                    return rebuild(catdim, NoLookup())
+                _status = check_cat_lookups(joindims...)
+                if _status === :ok
+                    _vcat_dims(joindims...)
+                elseif _status === :demote
+                    _vcat_dims_demote(joindims...)
+                elseif _status === :error
+                    D = basetypeof(first(joindims))
+                    intr = intersect(map(l -> Set(val(l)), lookup.(joindims))...)
+                    throw(DimensionMismatch(
+                        "Cannot concatenate along $D: lookups share values $intr. " *
+                        "Pass `dims=$D(newlookupvals)` with explicit lookup values."
+                    ))
+                else
+                    rebuild(catdim, NoLookup())
                 end
-                _vcat_dims(joindims...)
             end
         else
             # Concatenate new dims
@@ -351,7 +362,7 @@ function Base.hcat(As::Union{AbstractDimVector,AbstractDimMatrix}...)
         AnonDim(NoLookup())
     else
         joindims = map(last ∘ dims, As)
-        check_cat_lookups(joindims...) || return Base.hcat(map(parent, As)...)
+        check_cat_lookups(joindims...) === :ok || return Base.hcat(map(parent, As)...)
         _vcat_dims(joindims...)
     end
     noncatdim = dims(A1, 1)
@@ -369,7 +380,7 @@ end
 function Base.vcat(As::Union{AbstractDimVector,AbstractDimMatrix}...)
     A1 = first(As)
     firstdims = map(first ∘ dims, As)
-    check_cat_lookups(firstdims...) || return Base.vcat(map(parent, As)...)
+    check_cat_lookups(firstdims...) === :ok || return Base.vcat(map(parent, As)...)
     newdims = if A1 isa AbstractDimVector
         catdim = _vcat_dims(firstdims...)
         (catdim,)
@@ -391,38 +402,39 @@ end
 function Base.vcat(d1::Dimension, ds::Dimension...)
     dims = (d1, ds...)
     comparedims(dims...; length=false)
-    check_cat_lookups(dims...) || return Base.vcat(map(parent, dims)...)
+    check_cat_lookups(dims...) === :ok || return Base.vcat(map(parent, dims)...)
     return _vcat_dims(d1, ds...)
 end
 
 check_cat_lookups(dims::Dimension...) =
-    _check_cat_lookups(basetypeof(first(dims)), lookup(dims)...)
+    _check_cat_lookups(basetypeof(first(dims)), lookup(dims)...)::Symbol
 
 # Lookups may need adjustment for `cat`
 _check_cat_lookups(D, lookups::Lookup...) = _check_cat_lookup_order(D, lookups...)
-_check_cat_lookups(D, l1::NoLookup, lookups::NoLookup...) = true
+_check_cat_lookups(D, l1::NoLookup, lookups::NoLookup...) = :ok
 function _check_cat_lookups(D, l1::AbstractSampled, lookups::AbstractSampled...)
-    length(lookups) > 0 || return true
-    _check_cat_lookup_order(D, l1, lookups...) || return false
+    length(lookups) > 0 || return :ok
+    ostatus = _check_cat_lookup_order(D, l1, lookups...)
+    ostatus === :ok || return ostatus
     _check_cat_lookups(D, span(l1), l1, lookups...)
 end
 function _check_cat_lookups(D, ::Regular, lookups...)
-    length(lookups) > 1 || return true
+    length(lookups) > 1 || return :ok
     lastval = last(first(lookups))
     s = step(first(lookups))
-    map(Base.tail(lookups)) do l
+    for l in Base.tail(lookups)
         if !(span(l) isa Regular)
             _mixed_span_warn(D, Regular, span(l))
-            return false
+            return :drop
         end
         if !(step(span(l)) == s)
             @warn _cat_warn_string(D, "step sizes $(step(span(l))) and $s do not match")
-            return false
+            return :drop
         end
-        _check_cat_step_join(D, lastval, first(l), s) || return false
+        _check_cat_step_join(D, lastval, first(l), s) || return :drop
         lastval = last(l)
-        return true
-    end |> all
+    end
+    return :ok
 end
 
 function _check_cat_step_join(D, lastval::Number, firstval::Number, steplen)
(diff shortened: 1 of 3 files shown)


ghyatzo:
test failures are unrelated (Makie) and I have all greens on 1.12.6 locally

ghyatzo:
pushed a fix to recover a pretty bad regression in perfomance on the happy path. I was computing the intersection set too eagerly. Now it's comparable to master.

```
julia> dd_1_c = DimArray(1:1000, Ti(DateTime.(1000:1999)));
julia> dd_2_c = DimArray(1:1000, Ti(DateTime.(2000:2999)));
julia> @btime cat($dd_1_c, $dd_2_c, dims=Ti)
```
this PR (before): `35.000 μs (80 allocations: 87.31 KiB)`
this PR (after): `4.100 μs (66 allocations: 32.98 KiB)`
v0.30.1: `3.557 μs (75 allocations: 33.12 KiB)`

src/array/methods.jl
```diff
@@ -391,38 +402,39 @@ end
...
-    length(lookups) > 1 || return true
+    length(lookups) > 1 || return :ok
     lastval = last(first(lookups))
     s = step(first(lookups))
-    map(Base.tail(lookups)) do l
+    for l in Base.tail(lookups)
         if !(span(l) isa Regular)
             _mixed_span_warn(D, Regular, span(l))
-            return false
+            return :drop
```
rafaqz: `:drop` is overloading `Base.drop` which drops 1d dimensions. Can you think of words that better communicate the status?

rafaqz (commented):
This is a really nice PR. We never put the time in to properly resolve these Unordered lookups properly before but its correct.

My main comment is that the status symbols could more clearly communicate the status or action it triggers.

felixcremer:
I am wondering whether we should rather return the order types from check_cat_lookups instead of symbols?
This would enable to specify the order that results from the combination of the lookups of a cat without having to reinvent now nomenclature.

rafaqz:
I think the symbols specify more than just the order, like Nolookup is `:ok` too

src/array/methods.jl
```diff
@@ -442,58 +454,70 @@ function _check_cat_step_join(D, lastval, firstval, steplen)
...
+        span(l) isa Explicit || (_mixed_span_warn(D, Explicit, span(l)); return :drop)
+    end
+    return :ok
 end
 function _check_cat_lookups(D, ::Irregular, lookups...)
-    map(lookups) do l
-        span(l) isa Irregular || _mixed_span_warn(D, Irregular, span(l))
-    end |> all
+    for l in lookups
+        span(l) isa Irregular || (_mixed_span_warn(D, Irregular, span(l)); return :drop)
```
rafaqz: We should jist make an if block here these are gettimg hard to read

ghyatzo:
what about this for the status
- `:ok` -> `:simple` (but maybe `:ok` is fine as well?)
- `:demote` -> `:unordered`
- `:error` -> `:intersecting`
- `:drop` -> `:conflicting`

it separates more the "what" from the "what we should do about it" which now are indeed a bit intertwined.

EDIT: went with `:mismatch` instead of `:conflicting`

src/array/methods.jl
```diff
@@ -301,10 +301,21 @@ function _cat(catdims::Tuple, A1::AbstractDimArray, As::AbstractDimArray...)
...
+                if _status === :ok
+                    _vcat_dims(joindims...)
+                elseif _status === :demote
+                    _vcat_dims_demote(joindims...)
+                elseif _status === :error
+                    D = basetypeof(first(joindims))
+                    intr = intersect(map(l -> Set(val(l)), lookup.(joindims))...)
+                    throw(DimensionMismatch(
+                        "Cannot concatenate along $D: lookups share values $intr. " *
+                        "Pass `dims=$D(newlookupvals)` with explicit lookup values."
```
ghyatzo, line 315: This has the potential of becoming a bit spammy if two big arrays share a lot of lookup values.
shall I remove it? Or maybe add a check that only shows the full list if there are less than X intersecting values?

felixcremer: I think it is fine and I would hope, that the print machinery would reduce this to a reasonable amount. 
We can still change it if it is annoying in practice.

ghyatzo:
```
ERROR: DimensionMismatch: Cannot concatenate along Ti: lookups share values
[DateTime("1799-01-01T00:00:00"), DateTime("1262-01-01T00:00:00"), DateTime("1832-01-01T00:00:00"), DateTime("1526-01-01T00:00:00"), DateTime("1696-01-01T00:00:00"), DateTime("1461-01-01T

Pass `dims=Ti(newlookupvals)` with explicit lookup values.
Stacktrace:
```
not spammy but pretty ugly. it just truncates.
I'll see if there's a nicer way

tiemvanderdeure: For other DimensionMismatch we do limit, maybe use a pattern similar to here: https://github.com/rafaqz/DimensionalData.jl/blob/f5af202f815d8e1e081149df72ec45b3e2a798cf/src/Dimensions/primitives.jl#L562-L567 ?

tiemvanderdeure: you can just use the internal `_limit_sprint`

ghyatzo: added a parent method near `_limit_sprint` that calls `Base.print_array` instead.
```
ERROR: DimensionMismatch: Cannot concatenate along Ti: lookups share values:
Shared:
 1799-01-01T00:00:00
 1262-01-01T00:00:00
 1832-01-01T00:00:00
 1526-01-01T00:00:00
 1696-01-01T00:00:00
 1461-01-01T00:00:00
 1567-01-01T00:00:00
 1059-01-01T00:00:00
 1610-01-01T00:00:00
 1516-01-01T00:00:00
 ⋮
 1587-01-01T00:00:00
 1781-01-01T00:00:00
 1961-01-01T00:00:00
 1291-01-01T00:00:00
 1810-01-01T00:00:00
 1232-01-01T00:00:00
 1361-01-01T00:00:00
 1073-01-01T00:00:00
 1270-01-01T00:00:00

Pass `dims=Ti(newlookupvals)` with explicit lookup values.
```
also got rid of the warn for this case since we already show the error.

ghyatzo:
Should be ready for review

ghyatzo:
Gentle bump

rafaqz:
Thanks this is great. But its mildly breaking... so `hcat` and `vcat` should probably be lumped with it.

ghyatzo:
I've rebased on master and boundled in hcat and vcat.
cleaned up the logic as well a bit.

Two notes:
- in case of hcat/vcat shared values I went with falling back to simple array hcat/vcat (like before-ish). 
while cat throws currently. where do we want to go? 

- Noticed a possible issue in overlap detection: it only runs on some of the misaligned paths:
two mixed order dimensions (forward/reverse) just return unordered with a warn but don't check for overlaps and keep duplicates. Also it might not check for all overlaps. Will look into it more in the following days!

ghyatzo:
I went a bit back to the drawing board. I am not sure I know what should be the policy for dimensions without unique values.
That murkies a little bit what should happen on intersection.

for example, this currently works (and it's a bit confusing):
```julia
julia> a = DimArray([10,20,30], X([1,1,2]))
┌ 3-element DimArray{Int64, 1} ┐
├──────────────────────────────┴─────────────────────────── dims ┐
  ↓ X Sampled{Int64} [1, …, 2] ForwardOrdered Irregular Points
└────────────────────────────────────────────────────────────────┘
 1  10
 1  20
 2  30

julia> a[X=1]
10

julia> a[X=At(1)]
20

julia> a[X=1:2]
┌ 2-element DimArray{Int64, 1} ┐
├──────────────────────────────┴──────────────────────── dims ┐
  ↓ X Sampled{Int64} [1, 1] ForwardOrdered Irregular Points
└─────────────────────────────────────────────────────────────┘
 1  10
 1  20
```

I guess that unless we have the perfect "disjoint same order" case we can drop to unordered, without trying to be too fancy trying to check for not disjoint but still monotonic, or interleaved cases (like `[1,3,5] + [2,4,6]`).

But it doesn't clarify what should happen on intersection: This is the current status and is highly incosistent:
<details><summary>Details</summary>
<p>

```julia
using DimensionalData
using DimensionalData.Lookups, DimensionalData.Dimensions
 U(v) = X(Sampled(v, Unordered(), Irregular(nothing, nothing), Points(), NoMetadata()))

 function run1(n, f)
     print(rpad(n, 24))
     try
         r = f()
         println("DEMOTED  ", order(lookup(r, X)), "  vals=", collect(val(dims(r, X))))
     catch e
         println("THROWS   ", first(split(sprint(showerror, e), '\n'))[1:min(end, 44)])
     end
 end

 run1("F/F sharing", () -> cat(DimArray(1:2, X([1, 2])), DimArray(1:2, X([2, 3])); dims=X))
 run1("R/R seam sharing", () -> cat(DimArray(1:2, X([3, 2])), DimArray(1:2, X([2, 1])); dims=X))
 run1("F/R mixed sharing", () -> cat(DimArray(1:2, X([1, 2])), DimArray(1:2, X([2, 1])); dims=X))
 run1("R/F mixed sharing", () -> cat(DimArray(1:2, X([2, 1])), DimArray(1:2, X([1, 2])); dims=X))
 run1("U/U sharing", () -> cat(DimArray(1:3, U([3, 1, 2])), DimArray(1:2, U([2, 5])); dims=X))
 run1("U/U after misalignment", () -> cat(DimArray(1:3, U([3, 1, 2])), DimArray(1:2, U([2, 9])), DimArray(1:1, U([3])); dims=X))
 run1("F/F sharing in 3rd arg", () -> cat(DimArray(1:2, X([1, 10])), DimArray(1:2, X([5, 6])), DimArray(1:1, X([1])); dims=X))
 run1("dup inside an input", () -> cat(DimArray(1:3, X([1, 1, 2])), DimArray(1:1, X([5])); dims=X))
```

</p>
</details> 

   | Condition | Example | `cat` today | Decided at |
   |---|---|---|---|
   | F/F, sharing pair reached | `cat(X([1,2]), X([2,3]))` | throws | `methods.jl:490` |
   | R/R, sharing at the seam | `cat(X([3,2]), X([2,1]))` | throws | `methods.jl:507` |
   | F/R or R/F, mixed order | `cat(X([1,2]), X([2,1]))` | demotes → `Unordered`, `[1,2,2,1]` | no sharing test at all |
   | U/U pair | `cat(U([3,1,2]), U([2,5]))` | throws | `methods.jl:521` |
   | U/U/U, only a *subset* of the inputs shares | `cat(U([1,2]), U([2,3]), U([5]))` | demotes → `[1,2,2,3,5]` | `methods.jl:521` — n-way `intersect(lookups...)` |
   | sharing only in a later argument | `cat(X([1,10]), X([5,6]), X([1]))` | demotes → `[1,10,5,6,1]` | `methods.jl:490` — `1:i` list plus early return |
   
  Should we always throw on shared values? Always allow with a warning?
  
One way could be of instead of computing the intersection in each branch like we do now, we first
check for shared values acroll all lookups and then to the order comparison and then act upon
the result
  ```
  shared = <shared values across all lookups>
  monotone = <order walk outcome (what we do currently)>
  
  # policy decision:
  shared & monotone -> ?
      - keep F/R ? 
      - demote to IrregularPoints ?
      - throw ?
      - differentiate between Shared at seams or in the interior ?
  shared & !monotone -> ? / unordered
      - throw ?
      - demote to irregular points ?
      - (we don't need to differentiate here since the seam not seam difference is just to keep eventually a monotonic order or not)
 !shared & monotone -> ok
 !shared & !monotone -> demote / unordered
```

PR #1209, opened 2026-07-17
Makie v0.25 support

icweaver:
Hi all, prompted by @felixcremer, this is a first crack at getting DD compatible with the upcoming Makie v0.25 release. Happy to make any edits. Here's a summary below (assisted with Fable 5). Thanks for taking a look!

---

Updates the Makie extension for the upcoming breaking Makie v0.25 release (tested against [`MakieOrg/Makie.jl#ff/breaking-0.25`](https://github.com/MakieOrg/Makie.jl/tree/ff/breaking-0.25)), while keeping compatibility with all currently supported Makie versions (0.20–0.24).

### Extension changes (`ext/DimensionalDataMakieExt.jl`)

- **`Makie.attribute_names` was removed in v0.25** (nested-attributes rework, [Makie#5620](https://github.com/MakieOrg/Makie.jl/pull/5620)). `Makie_attribute_names` now falls back to `propertynames(P)`, which on v0.25 returns Block attribute names via `root_keys(documented_attributes(T))`. This is what filters axis keywords for `Axis`/`Axis3`/`LScene`/`PolarAxis`.
- **Fixed default axis type selection** ([Makie#5375](https://github.com/MakieOrg/Makie.jl/pull/5375)). Makie v0.24.11 replaced the `args_preferred_axis(::Type, args...)` methods with instance-based `preferred_axis_type` dispatch, but we pick the axis type *before* a plot instance exists, so `volume`/`volumeslices` silently fell back to `Axis` instead of `LScene` and the `surface` flat-data check was lost. `default_axis_type` now uses a `_preferred_axis_type` helper that mirrors Makie's fallback hierarchy (plot type + converted args --> plot type --> per-argument), with explicit methods for plot types whose Makie methods are instance-only (`Volume` --> `LScene`, `Surface`/`Wireframe`/`Contour3d` --> flat check). This also fixes a latent regression on released Makie v0.24.11/v0.24.12.
- **Fixed LScene axis labels**: multi-key nested attribute access (`plot[:names, :axisnames]`) errors on v0.25's compute-graph plots. `add_labels_to_lscene` now branches on what `to_value(plot[:names])` returns. `Attributes` (≤ v0.24) keeps the old semantics, otherwise it uses the v0.25 nested-node chain `plot[:names][:axisnames]`.
- **Supported colorbar plots**. Added `contour`, `surface`, and `spy` to the list (was just `heatmap` and `contourf` before)

### Test changes (`test/makie.jl`)

- Unitful `rainclouds`/`violin`/`boxplot` now work on v0.25 thanks to the expanded dim converts ([Makie#5323](https://github.com/MakieOrg/Makie.jl/pull/5323)), so the `broken = true` marks are now version-conditional: still broken below v0.25, asserted to return a `FigureAxisPlot` on v0.25+.
- `ax.finallimits.val` --> `ax.finallimits[]`: the `Axis` compute-graph refactor ([Makie#5546](https://github.com/MakieOrg/Makie.jl/pull/5546)) removed the Observable `.val` field. `[]` works on every version.
- `rainclouds` on data containing `missing` now throws `MethodError` from the dim-converts machinery instead of `ErrorException`. The `@test_throws` is version-conditional.
- The AlgebraOfGraphics testset (and its `using AlgebraOfGraphics`) is gated on `pkgversion(Makie) < v"0.25"`, since AlgebraOfGraphics does not support Makie v0.25 yet. This lets `test/makie.jl` run in a Makie v0.25 environment where AlgebraOfGraphics cannot be installed. The `DimPoints` testset moved out of the AlgebraOfGraphics block since it only uses Makie, so it always runs. Note that AlgebraOfGraphics is still in the test target, so `Pkg.test()`/CI will keep resolving Makie v0.24.x until AlgebraOfGraphics gains v0.25 compat.

## TODO

- [ ] Felix has a PR for fixing the `_preferred_axis_type` workaround upstream (https://github.com/MakieOrg/Makie.jl/issues/5704)


JamesWrigley:
Is there any chance of getting this in before the upstream Makie issue is fixed, maybe even with some `@test_broken`'s? :see_no_evil: Asking because it'd be cool to unblock CI for some other PRs I have.

rafaqz:
Will be faster to merge if you could review it

JamesWrigley:
LMAO I see you nerd-sniping me 😂
(but sure, will try to look today)

test/makie.jl
```diff
@@ -1,12 +1,17 @@
...
-using AlgebraOfGraphics
 using CairoMakie
 using CairoMakie: ComputePipeline
 using ColorTypes
 using Unitful, Unitful.DefaultSymbols
 import Distributions
 import DimensionalData as DD
 
+# AlgebraOfGraphics does not support Makie v0.25 yet, so only test it on older versions
+const TEST_ALGEBRAOFGRAPHICS = pkgversion(Makie) < v"0.25"
```
tiemvanderdeure:
```suggestion
const TEST_ALGEBRAOFGRAPHICS = "AlgebraOfGraphics" in keys(Pkg.project().dependencies)
```

This doesn't really work for actually testing Makie 0.25 through CI, since AoG is in the test environment and so Makie resolves to 0.24. But for running locally in an environment without AoG this should work and is much more future-proof than testing for the version of Makie (it will work in the future when a version of AoG with Makie 0.25 is released)

icweaver: For sure, thanks!

test/makie.jl
```diff
@@ -1,11 +1,17 @@
...
 using ColorTypes
 using Unitful, Unitful.DefaultSymbols
 import Distributions
 import DimensionalData as DD
+import Pkg
+
+# AlgebraOfGraphics does not support Makie v0.25 yet, so only test it on older versions
+const TEST_ALGEBRAOFGRAPHICS = "AlgebraOfGraphics" in keys(Pkg.project().dependencies)
+if TEST_ALGEBRAOFGRAPHICS
+    @eval using AlgebraOfGraphics
```
JamesWrigley: Nitpicking, I don't think the `@eval` is really necessary since it's a top-level expression?

JamesWrigley (changes requested):

test/makie.jl
icweaver: oop, fixed https://github.com/rafaqz/DimensionalData.jl/pull/1209/commits/f21d794885609811447a7b69d06fe94e077e4b1c

JamesWrigley (approved):
IANAE, but LGTM!

tiemvanderdeure:
Have you tested locally in an environment without AoG to make sure all tests pass with Makie 0.25?

EDIT: okay I've been a bit out of the loop and only now realized Makie 0.25 and CairoMakie 0.16 aren't even out yet. I'm not sure if there is any general consensus about how to handle this but it seems a bit odd to add compats for unreleased versions? Maybe we can merge this just without those compats now, and then widen the compat bounds as soon as Makie 0.25 and CairoMakie 0.16 are released? Would be nice to get this merged soon as it will unblock all of those other PRs

ext/DimensionalDataMakieExt.jl
```diff
@@ -368,13 +377,47 @@ end
...
-    default_type = Makie.args_preferred_axis(P, output...) 
+    default_type = _preferred_axis_type(P, output...)
     isnothing(default_type) ? Axis : default_type
 end
 
+if isdefined(Makie, :preferred_axis_type)
+    # Makie >= 0.24.11 moved axis type selection from `args_preferred_axis` to
+    # `preferred_axis_type`, which mostly dispatches on plot instances. We only have
+    # a plot type and converted arguments here, so mirror the parts of
+    # `Makie._preferred_axis_type` that work with those.
```
tiemvanderdeure, line 388:
```suggestion
    # `Makie._preferred_axis_type` that work with those.
    # see: https://github.com/MakieOrg/Makie.jl/issues/5704
```

Let's link the issue here for our future selves?

icweaver: thx!

tiemvanderdeure:
@icweaver @rafaqz What do you think about my proposal - to merge this now without the changes to compat bounds and then just update the compat as soon as Makie 0.25 is released?

Then we can merge all of those PRs where tests now fail because of Makie

Project.toml
```diff
@@ -81,11 +81,12 @@ InvertedIndices = "1"
 IteratorInterfaceExtensions = "1"
 JLArrays = "0.1, 0.3"
 LinearAlgebra = "1"
-Makie = "0.20, 0.21, 0.22, 0.23, 0.24"
+Makie = "0.20, 0.21, 0.22, 0.23, 0.24, 0.25"
```
tiemvanderdeure:
```suggestion
Makie = "0.20, 0.21, 0.22, 0.23, 0.24"
```

Project.toml
```diff
@@ -57,7 +57,7 @@ AlgebraOfGraphics = "0.8, 0.9, 0.10, 0.11, 0.12"
 Aqua = "0.8"
 ArrayInterface = "7"
 BenchmarkTools = "1"
-CairoMakie = "0.10, 0.11, 0.12, 0.13, 0.14, 0.15"
+CairoMakie = "0.10, 0.11, 0.12, 0.13, 0.14, 0.15, 0.16"
```
tiemvanderdeure:
```suggestion
CairoMakie = "0.10, 0.11, 0.12, 0.13, 0.14, 0.15"
```

tiemvanderdeure (commented):
Let's let CompatHelper take care of the compat bounds and merge as is otherwise

rafaqz:
Yeah Ive never merged before the version comes out! The one risk is this is code against promises that have not been made yet. Makie can still change anything until the version actually gets registered. Can we just wait until that happens? Then we can run tests against it

ext/DimensionalDataMakieExt.jl
```diff
@@ -368,13 +377,48 @@ end
 
 function default_axis_type(::Type{P}, dd; kwargs...) where P
     output = Makie.convert_arguments(P, to_value(dd); kwargs...)
-    default_type = Makie.args_preferred_axis(P, output...) 
+    default_type = _preferred_axis_type(P, output...)
     isnothing(default_type) ? Axis : default_type
 end
 
+if isdefined(Makie, :preferred_axis_type)
```
felixcremer, line 384: I am wondering whether this is all still necessary after https://github.com/MakieOrg/Makie.jl/pull/5722

felixcremer (approved):
Looks good to me, thank you. 
I think we can sort out the changes by https://github.com/MakieOrg/Makie.jl/pull/5722 in a later PR because it is not yet released in a proper version. 
For now we should merge this PR without the bounds changes so that we unblock the CI for Makie on the other PRs.

ext/DimensionalDataMakieExt.jl
icweaver: Sorry, I'm not too sure anymore. I've gotten pretty lost in the `args_preferred_axis`/`_preferred_axis_type`/`preferred_axis_type` story, tbh, and will leave it to someone more experienced with Makie's internals to take over from here

felixcremer: Yes, this should not block this PR.

tiemvanderdeure (approved):
I think we can merge this now. @rafaqz ?

felixcremer:
The nightly failure is an unrelated failure from Statistics and the MacOS failure is a failure of installing conda. So I'd say good to ignore.

PR #1212, opened 2026-07-26
Use optimization level 0 in the tests

JamesWrigley:
Should make running them a bit faster. Also updated some other action versions.

Tried it locally on the FFT tests and the running time went from 25s -> 16s.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
           - macOS-latest
           - windows-latest
         arch:
-          - x64
+          - default
         allow_failure: [false]
         include:
           - version: 'nightly'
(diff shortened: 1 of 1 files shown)


JamesWrigley:
Makie tests are still failing so we can only compare 1.10 on linux, but still:
- Precompilation time: 471s -> 347s
- FFT: 61s -> 34s
- Selector: 45s -> 23s
- Broadcast: 71s -> 34s
- Tables: 69s -> 38s
- Xarray 50s -> 42s

Selection of old timings taken from this CI job: https://github.com/rafaqz/DimensionalData.jl/actions/runs/29649062601/job/88092149639?pr=1210
And importantly it seems that this doesn't affect any of the allocations/inference tests.

JamesWrigley:
Some other data points for 1.12 on ubuntu compared to [main](https://github.com/rafaqz/DimensionalData.jl/actions/runs/31100060235/job/92611179557):
- Wall time: 35m -> 25m
- Precompilation time: 860s -> 703s
- Makie: 318s -> 178s

tiemvanderdeure (approved):
Crazy this is such a speedup! I see no reason not to merge right away

JamesWrigley:
I haven't figured out a nice way to do it yet, but I think we might want to go even further and just disable precompilation for most test dependencies like Makie. On the 1.12 windows runs precompilation takes almost 15 minutes, which is half the total time :grimacing:

tiemvanderdeure:
And we don't run makie tests on windows, so all that precompiled code is not even used...

asinghvi17 (approved):

asinghvi17:
Disabling precompilation is a bit harder, ideally with the cache it _should_ just hit it.

JamesWrigley:
(bump)

PR #1213, opened 2026-07-27
Add support for converting DimArray/DimStack to DataArray/Dataset

JamesWrigley:
Largely based on https://github.com/arviz-devs/ArviZPythonPlots.jl/blob/main/src/xarray.jl, which was pointed out by @sethaxen in https://github.com/rafaqz/DimensionalData.jl/pull/882#issuecomment-2571612610.

I figured that it could be annoying in some cases to always explicitly convert to an xarray/generic type, so I added support for a `xarray_convert` preference that lets the user specify the default behaviour. Shouldn't have any effect on load times since Preferences.jl is already a dependency of PrecompileTools.jl and Tables.jl.


JamesWrigley:
(bump)

PR #1214, opened 2026-08-04
docs: rewrite "Extending DimensionalData" around wrap-vs-subtype

bjarthur:
i had claude extend DD for me to create [SparseDimArrays.jl](https://github.com/JaneliaSciComp/SparseDimArrays.jl).  it was taking me awhile to figure out exactly how what it did works, so i then had it add to the existing docs on extending DD enough detail that a dummy like me could understand.  let me know what you think!

### DIFF
--- a/docs/src/extending_dd.md
+++ b/docs/src/extending_dd.md
@@ -1,106 +1,422 @@
 ```@meta
-Description = "Extend DimensionalData.jl - create custom array types, stacks, and lookups by implementing key interface methods in Julia"
+Description = "Extend DimensionalData.jl - create custom array types, stacks,
+lookups and dimensions by implementing a small set of interface methods in
+Julia"
 ```
 
 # Extending DimensionalData
 
-Nearly everything in DimensionalData.jl is designed to be extensible.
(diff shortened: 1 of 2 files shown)


tiemvanderdeure:
Thanks for this - I have run into similar things before where I was unsure which methods I needed to define and which had AbstractDimArray fallbacks.

I'm not so sure if the framing here is the most logical. It feels very julian to be able to wrap any AbstractArray in a DimArray, so I had never thought of that as "extending" DimensionalData. In almost any case it "just works" and you wouldn't have to implement any additional methods, just by using the AbstractArray interface. Or put differently: wrapping a custom array in a DimArray extends AbstractArray, not AbstractDimArray. In constrast when making a AbstractDimArray it's possible to change how DD behaves, and users would need to know a bunch of DD methods.

But maybe it's good to have an example of wrapping some different array type in a DimArray. Then the extension of DD on CategoricalArrays might be a more intuitive example, since it's an array type that's actually out there and used (and it still needed a bit of work of forwarding CategoricalArray methods to the parent)?

Issue #1215, opened 2026-08-11
Wrong bins and categorical axis values in boxplot/violin/raincloud plots


jamesalster:
When using the automatic plotting functionality for boxplot, violin and raincloud plots, there appear to be two bugs:

* The binning along the x/categorical axis is incorrect, and values are mixed between the x-axis (categorical) dimensions
* The x-axis positions appear to be weird (do they come from conversion to character the numeric values)?

This bug may apply to other plot types as well. I'm happy to contribute a PR to fix it, if that would be helpful.

```julia
using DimensionalData, GLMakie

# category `aa` is all 0.0, category `bb` is all 10.0
A = DimArray(repeat([0.0, 10.0], 1, 100), (X([:aa, :xx]), Y(1:100)))

xs, ys = Makie.convert_arguments(BoxPlot, A)
@show unique(xs)                    # bug: why are positions 194, 240?
@show unique(ys[xs .== first(xs)])  # bug: should be [0.0] only since x is all 0s
boxplot(A) # incorrect
```

tiemvanderdeure:
Can confirm, looks like a bug. Probably the login in convert_arguments is just wrong? Would be great if you can make a PR

jamesalster:
Thanks - I've opened a PR #1216 proposing a solution.

PR #1216, opened 2026-08-12
Fix bins and categorical axes in violin/raincloud/boxplots

jamesalster:
Here's a PR for #1215 . There were in fact two bugs:

1. Values were mixed between categorical bins by `Makie.convert_arguments()` by using `outer` rather than `inner` in repeat.
2. X (categorical) axis positions were assigned by summing char values by `get_number_version()`, which led to weird positions for strings of length > 1.

When fixing (2.) I decided that the best behaviour would be to always preserve the order of the original `Lookup` and so modified the `get_number_version()` call by moving `parent()` inside the function so that it knows when it's passed a `Lookup`. 

However, I have a question: the code specifically converts `Char` lookups to numeric codes, so an axis of e.g. `['a', 'z']` would become `[97, 122]`. I've maintained that behaviour for fear of breaking another use case, only fixing bug (2.) for other categorical types like `String`. But it seems strange, and I wondered whether we wouldn't want to make `Char` consistent with `String`? 

I also fixed two separate bugs: z-axis labels were broken because of a misplaced bracket, and `DimVector`s were wrongly passed to base Makie by a `DimMatrix` type signature.

I used Claude to brainstorm, suggest tests and check for mistakes, but I reviewed all code myself.
It's my first time contributing to DimensionalData, so let me know if anything is out of place.

### DIFF
--- a/ext/DimensionalDataMakieExt.jl
+++ b/ext/DimensionalDataMakieExt.jl
@@ -169,7 +169,7 @@ for (p1) in PlotTypes_Cat_1D
     f1! = Symbol(f1, '!')
     
     @eval begin
-        function Makie.$f1(fig::MakieGrids, A::MayObs{AbstractDimArray}; categoricaldim = nothing, axis = (;), plot_user_attributes...)
+        function Makie.$f1(fig::MakieGrids, A::MayObs{AbstractDimMatrix}; categoricaldim = nothing, axis = (;), plot_user_attributes...)
             error_if_has_content(fig)
 
             ax_type = haskey(axis, :type) ? axis[:type] : default_axis_type($p1, A)
@@ -226,7 +226,7 @@ function axis_attributes(::Type{Series}, A::MayObs{DD.AbstractDimMatrix}; labeld
     )
 end
 
-function axis_attributes(::Type{<:Union{RainClouds, BoxPlot, Violin}}, A::MayObs{DD.AbstractDimArray}; categoricaldim)
+function axis_attributes(::Type{<:Union{RainClouds, BoxPlot, Violin}}, A::MayObs{DD.AbstractDimMatrix}; categoricaldim)
     categoricaldim = _categorical_or_dependent(to_value(A), categoricaldim)
     isnothing(categoricaldim) && throw(ArgumentError("No dimensions have Categorical lookups"))
 
@@ -664,10 +664,8 @@ function _lookup_to_vector(l)
         bs = intervalbounds(l)
         x = first.(bs)
         push!(x, last(last(bs)))
-    elseif iscategorical(l)
+    else # points or categorical
         get_number_version(l)
-    else # ispoints(l) but not categorical
-        collect(parent(l))
     end
 end
 
--- a/ext/DimensionalDataMakieExt.jl
+++ b/ext/DimensionalDataMakieExt.jl
@@ -169,7 +169,7 @@ for (p1) in PlotTypes_Cat_1D
     f1! = Symbol(f1, '!')
     
     @eval begin
-        function Makie.$f1(fig::MakieGrids, A::MayObs{AbstractDimMatrix}; categoricaldim = nothing, axis = (;), plot_user_attributes...)
+        function Makie.$f1(fig::MakieGrids, A::MayObs{DD.AbstractDimVecOrMat}; categoricaldim = nothing, axis = (;), plot_user_attributes...)
             error_if_has_content(fig)
 
             ax_type = haskey(axis, :type) ? axis[:type] : default_axis_type($p1, A)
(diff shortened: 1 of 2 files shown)


test/makie.jl
```diff
@@ -124,22 +124,26 @@ using DimensionalData: Metadata, NoMetadata, ForwardOrdered, ReverseOrdered, Uno
...
 
+    # multi-character labels are equally spaced, in lookup order
+    dd_cat_multi = DimArray(randn(4, 20), (X([:α, :Cyl, :Disp, :σ]), Y(1:20)), name = :test)
+    @test boxplot(dd_cat_multi).axis.xticks[] == ([1, 2, 3, 4], ["α", "Cyl", "Disp", "σ"])
 
     dd_cat = DimArray((1:6).^2, X(cat(fill('A', 3), fill('B', 3), dims = 1)), name = :test)
     for plot_i in (rainclouds, violin, boxplot)
         fig, ax, plt = plot_i(dd_cat)
         @test all(plt[1][] .== Int.(lookup(dd_cat, X)))
+        @test length.(plot_i(dd_cat).axis.xticks[]) == (2, 2)
```
tiemvanderdeure:
```suggestion
        @test plot_i(dd_cat).axis.xticks[] == ([Int('A'),Int('B')], ["A", "B"])
```

Might as well test this explicitly and also makes it easier to read

test/makie.jl
```diff
@@ -124,22 +124,26 @@ using DimensionalData: Metadata, NoMetadata, ForwardOrdered, ReverseOrdered, Uno
             x = parent(lookup(to_value(dd_cat), Y))
             y = collect(parent(to_value(dd_cat)))
             fig, ax, plt = plot_i(obs(dd_cat))
-            @test all(plt[1][] .== repeat(97:98, outer = 6)) 
+            @test all(plt[1][] .== repeat(97:98, inner = 6))
             @test all(plt[2][] .== vec(y'))
```
tiemvanderdeure:
```suggestion
            @test plt[1][] == repeat(97:98, inner = 6)
```

tiemvanderdeure:
```suggestion
            @test plt[2][] == vec(y')
```

ext/DimensionalDataMakieExt.jl
```diff
@@ -169,7 +169,7 @@ for (p1) in PlotTypes_Cat_1D
     f1! = Symbol(f1, '!')
     
     @eval begin
-        function Makie.$f1(fig::MakieGrids, A::MayObs{AbstractDimMatrix}; categoricaldim = nothing, axis = (;), plot_user_attributes...)
+        function Makie.$f1(fig::MakieGrids, A::MayObs{AbstractDimArray}; categoricaldim = nothing, axis = (;), plot_user_attributes...)
```
tiemvanderdeure: I don't think we should allow any DimArray for these plots? At least I just tried with a simple one and got a complicated error. We also don't test for anything else than DimMatrices.

ext/DimensionalDataMakieExt.jl
```diff
@@ -226,7 +226,7 @@ function axis_attributes(::Type{Series}, A::MayObs{DD.AbstractDimMatrix}; labeld
     )
 end
 
-function axis_attributes(::Type{<:Union{RainClouds, BoxPlot, Violin}}, A::MayObs{DD.AbstractDimMatrix}; categoricaldim)
+function axis_attributes(::Type{<:Union{RainClouds, BoxPlot, Violin}}, A::MayObs{DD.AbstractDimArray}; categoricaldim)
```
tiemvanderdeure: same comment here, restricting dispatch to DimMatrix is correct I think?

ext/DimensionalDataMakieExt.jl
```diff
@@ -663,7 +664,9 @@ function _lookup_to_vector(l)
         bs = intervalbounds(l)
         x = first.(bs)
         push!(x, last(last(bs)))
-    else # ispoints(l)
+    elseif iscategorical(l)
+        get_number_version(l)
+    else # ispoints(l) but not categorical
         collect(parent(l))
```
tiemvanderdeure:
```suggestion
    else # points or categorical
        get_number_version(l)
```

claude points out that this breaks volumeslices plots if `l` has `Char` eltype but is not categorical. E.g. `DimArray(rand(X(Sampled(['a', 'b', 'c'])), Y(1:4), Z(1:5))) |> volumeslices` now triggers a weird error.

Maybe we should just always forward to `get_number_version` for that reason? If `l` is already numbers it would just be a no-op anyway.

jamesalster:
Thank you for the review @tiemvanderdeure . I've incorporated all your changes, the only extra thing was that setting the type to `AbstractDimMatrix` meant that 1D vectors did not have their string labels passed to Makie, which was caught by the test at `makie.jl` line 146. 

I adjusted to `AbstractDimVecOrMat` and added a function to informatively catch arrays with 3 or more dimensions (which otherwise were passed straight to Makie and broke uninformatively as you say).

tiemvanderdeure (approved):
I think this all looks fine. But maybe someone who knows more about Makie internals than me could take another look?

PR #1221, opened 2026-08-21
Disable precompilation workloads in the tests

JamesWrigley:
I got rather annoyed with [Cairo]Makie constantly getting precompiled when activating the test environment, so this PR disables precompilation entirely in the tests. It doesn't disable precompilation outside of the tests, which is handy for benchmarking. Separating the test dependencies into a `test/Project.toml` seemed like the simplest way to do it since otherwise a `test/LocalPreferences.jl` file will not be picked up.

Also switched to using dependabot over CompatHelper, which is what the workspace is for: https://discourse.julialang.org/t/psa-github-dependabot-now-supports-julia/134997

It *should* also make CI a bit faster.


JamesWrigley:
Some numbers compared to [main](https://github.com/rafaqz/DimensionalData.jl/actions/runs/32456977978/job/96696043215):
- On 1.12 on windows the tests execute with the current settings ~4 minutes slower than a job that didn't require any precompilation at all (i.e. precompilation takes an extra 4 minutes on top of the tests compared to 15 before).
- Running time on 1.12 on ubuntu went from 25 minutes -> 15 minutes.

JamesWrigley:
(bump)

tiemvanderdeure (approved):
seems like a good change to me

rafaqz (approved):

